PR #7, opened 2026-07-18
fix(dashboard): resolve API origin correctly in docker-compose

shiv19:
## Summary
- Dashboard SSR fetches (e.g. `GET /health/env`) always resolved the API origin from a static runtime-target table, which defaults to `http://localhost:4000`. Inside the dashboard container that never reaches the `api` container, so every `docker compose up` fails with `Cannot resolve deployment info: GET /health/env is unreachable ... ECONNREFUSED`.
- `INTERNAL_API_URL` already existed as an override in `apps/dashboard/src/lib/server/api.ts`, but was only honored when `NEXT_PUBLIC_API_PROXY=true` — a `NEXT_PUBLIC_*` build-time flag the dashboard `Dockerfile` never sets, so the override was effectively dead code in this deployment path.
- Loosened `getServerApiBaseUrl` to honor `INTERNAL_API_URL` unconditionally for server-side requests (browser-side resolution is untouched), and set it in `docker-compose.yml` to the in-cluster DNS name (`http://api:4000`) for the `dashboard` service — matching the same pattern already used for `api`'s own `DATABASE_URL`/`REDIS_URL` overrides.

## Test plan
- [x] `docker compose up -d --build dashboard` — dashboard boots clean, no `ECONNREFUSED` / `health/env` error in logs
- [x] `curl http://localhost:3001` returns a normal redirect instead of the fatal error page
- [ ] Maintainers verify no regression for non-docker (bare-metal/desktop) `OPENSHIP_LOCAL_API_URL` and proxy-mode (`NEXT_PUBLIC_API_PROXY=true`) paths, which are unaffected in code review but worth a sanity check

### DIFF
--- a/apps/dashboard/src/lib/server/api.ts
+++ b/apps/dashboard/src/lib/server/api.ts
@@ -120,15 +120,12 @@ function getServerApiBaseUrl(requestHeaders: Headers): string {
   // Desktop: the API runs on a dynamic port Electron injects here. Wins over
   // the header→table fallback (which can't know a dynamic port).
   const localOverride = process.env.OPENSHIP_LOCAL_API_URL?.replace(/\/+$/, "");
+  const internal = process.env.INTERNAL_API_URL?.replace(/\/+$/, "");
   if (localOverride) {
     origin = localOverride;
-  } else if (process.env.NEXT_PUBLIC_API_PROXY === "true") {
-    const internal = process.env.INTERNAL_API_URL;
-    if (internal) {
-      origin = internal.replace(/\/+$/, "");
-    } else {
-      origin = getApiOrigin(getRequestOriginFromHeaders(requestHeaders));
-    }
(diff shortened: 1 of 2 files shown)


shiv19:
The dashboard wasn't really loading for me after docker compose up without this fix.

Hydralerne:
Reviewed it - fix looks right, http://api:4000 matches the compose service and the SSR-direct routing makes sense. Two things first:

1. The JSDoc on getServerApiBaseUrl() still says the INTERNAL_API_URL short-circuit only kicks in when NEXT_PUBLIC_API_PROXY=true — not true anymore after this, so update it.
2. This makes INTERNAL_API_URL work regardless of the proxy flag. Can you check the SaaS build doesn't set it somewhere that relies on per-request origin? Would bypass it for SSR. Self-host path is fine.

PR #15, opened 2026-07-19
Support reverse proxy public URLs

DiogoDuart3:
## Summary

- add `openship up --public-url <origin>` for installations behind a reverse proxy
- persist the public URL when `openship up` installs an OS service
- use the configured public origin for dashboard auth redirects instead of falling back to `localhost:4000`
- preserve existing trusted origins and validate that the supplied URL is an HTTP(S) origin

## Why

The dashboard login flow passed `window.location.origin` into the static runtime-target resolver. A custom hostname, such as a private Tailscale DNS name or a normal reverse-proxy domain, does not match a compiled target and therefore fell back to the local target at `http://localhost:4000`.

The CLI also unconditionally replaced `OPENSHIP_LOCAL_API_URL` with its loopback API URL when launching the dashboard, which prevented operators from supplying the browser-facing origin.

With this change, an operator can put the dashboard and `/api` behind one proxy and run:

```bash
openship up --public-url https://openship.example.com
```

Authentication callbacks, cookies, CORS trusted origins, and the dashboard API client then use that origin. Existing local-only behavior remains unchanged when the option is omitted.

## Validation

- `bun run --cwd apps/cli test` — 6 tests passed
- `bun run --cwd apps/cli lint`
- `bunx tsc --noEmit -p apps/dashboard/tsconfig.json`
- `bun run --cwd apps/cli build`
- verified the generated service command retains `--public-url`
- verified URLs containing paths are rejected

Note: the existing dashboard `next lint` script is incompatible with the installed Next 16 CLI because `next lint` is interpreted as a project directory, so dashboard validation used TypeScript directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,9 @@ openship up           # installs Openship as a background service (starts on boo
 
 `openship open` opens the dashboard; `openship stop` stops the service. Want a one-off attached run instead? `openship up --foreground`. To deploy a project:
 
+When a reverse proxy exposes the dashboard and `/api` on one origin, configure
+the browser-facing URL with `openship up --public-url https://openship.example.com`.
+
 ```bash
 cd your-project
 openship init         # link this directory to a project
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -26,6 +26,7 @@
   "scripts": {
     "build": "tsup && bun run build/stage-server.ts",
     "lint": "tsc --noEmit",
+    "test": "vitest run",
     "clean": "rm -rf dist .turbo node_modules"
   },
   "dependencies": {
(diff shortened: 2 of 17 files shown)


DiogoDuart3:
Follow-up fix included in 0b7b192:

- expose **This Machine** as an actual local deploy target on self-hosted/desktop installations
- allow the wizard to continue when `deployTarget` is `local`
- count the local machine when determining whether multiple targets are available
- let the settings API default take precedence over the browser last-pick instead of skipping the target resolver
- add localized labels and descriptions for the local deploy target

Validation: dashboard TypeScript check passed and the production dashboard bundle was syntax-checked after applying the equivalent live patch.

Hydralerne:
Where's the linking to openresty / routing pipeline we have ? 
shouldn't you issue ssl / complete the route for it ? would be nice feature also to handle the app as deployed app by itself, 
so openship app since its public, it can be added inside projects to monitor / control

PR #45, opened 2026-07-20
fix(dashboard): Update docs urls to the correct domain

roybarber:
## Simple patch

When navigating the dashboard documentation urls are pointing to an inccorect subdomain

Updated any reference of `docs.openship.io` to `openship.io/docs`


roybarber:
Hi @Hydralerne , thanks for getting these doc links fixed! 🙂

I noticed this PR was closed rather than merged, but the same fixes landed shortly after in 445b8353 and 362e6be9 — which means the change shipped without contributor attribution.

Totally understand if you prefer working in batch commits rather than merging PRs directly — no problem at all. Would you be open to adding a Co-authored-by trailer in a future commit so the contribution is reflected?

Co-authored-by: Roy Barber <hi@roybarber.com>
Happy to keep contributing either way, contributions is why we do this kind of thing — thanks for building OpenShip in the open! 🚀

Hydralerne:
hello, sorry i closed it due to dupplication fix, it waas fexed on another i guess? or what

PR #55, opened 2026-07-20
fix(dashboard): compute real deployment stats for Activity widget

AbdullahM07:
## Summary

The dashboard's Activity card always showed `Projects: 0`,
`Deployments: 0`, `Success rate: 0%` regardless of how many projects
or deployments actually existed.

## Root cause

Two independent bugs in `GET /projects/home`:

| Field | Before | After |
|---|---|---|
| Projects count | Backend returned `total_projects`; frontend reads `total_active_projects` — key never matched, always fell back to `0` | Backend now returns `total_active_projects` |
| Deployments count | Hardcoded to `0` in `project.controller.ts` | Computed from the `deployment` table via a real query |
| Success rate | Derived client-side from the two hardcoded-zero fields above → always `0%` | Derived from real success/total counts |

## Changes

- `packages/db/src/repos/deployment.repo.ts` — added `countStatsByProjects()`, a single aggregate query (`count(*)` + `count(*) filter (where status in ('ready', 'partial_failure'))`).
- `apps/api/src/modules/projects/project.controller.ts` — wired the new query into `getHome`, fixed the response key, and merged cloud deployment/success counts the same way cloud project counts were already merged.

## Test plan

- [x] `tsc --noEmit` on `packages/db` and `apps/api` — no type errors
- [x] Seeded an isolated dev instance with 1 project and 3 deployments (2 `ready`, 1 `failed`)
- [x] Verified `GET /projects/home` returns `{ total_active_projects: 1, total_deployments: 3, total_success_deployments: 2 }`
- [x] Verified the dashboard's Activity card renders `Projects: 1`, `Deployments: 3`, `Success rate: 67%`

### DIFF
--- a/apps/api/src/modules/projects/project.controller.ts
+++ b/apps/api/src/modules/projects/project.controller.ts
@@ -164,7 +164,7 @@ export async function getHome(c: Context) {
     return c.json({
       success: true,
       projects: [],
-      numbers: { total_projects: 0, total_deployments: 0, total_success_deployments: 0 },
+      numbers: { total_active_projects: 0, total_deployments: 0, total_success_deployments: 0 },
       otherOrgs: [],
     });
   }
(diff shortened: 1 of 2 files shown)


Hydralerne:
done, sorry for that but noticed from the team and done this morning

PR #59, opened 2026-07-20
feat(tunneling): implement the ngrok tunnel provider

ishanj12:
## Summary

- Implements the ngrok tunnel provider (was a stub throwing `ProviderNotImplementedError`), matching the shape of the existing Oblien provider.
- Uses `@ngrok/ngrok` (agent SDK) to open/manage the tunnel, and `@ngrok/ngrok-api` (optional, needs `NGROK_API_KEY`) to reserve a custom domain when `input.slug` is given.
- Free-tier accounts (only `NGROK_AUTHTOKEN` set) get their auto-assigned dev-domain — `create()` probe-connects once to learn the hostname, then `connect()` opens the real long-lived listener against it.
- `TunnelAgent` adapter wraps `@ngrok/ngrok`'s `Listener` (which has no native disconnect/error events, only `listener.join()`) into the `disconnect`/`close` events `manager.ts`'s reconnect-with-backoff expects.

## Test plan

- [x] `tsc --noEmit` clean across `apps/api`
- [x] New unit test suite (`apps/api/test/modules/tunneling/ngrok.provider.test.ts`, 22 tests, mocked SDKs) covering preflight, both create() paths (with/without API key, no-slug probe path), conflict/ownership disambiguation, delete()'s resource-id gating, and the TunnelAgent adapter's close-vs-drop event distinction
- [x] Live-tested against a real ngrok pay-as-you-go account: auto-assigned dev-domain tunnel created, connected, and confirmed reachable end-to-end (real HTTP response through the tunnel); reserved custom-domain path created/connected/deleted and verified via the API that only the test reservation was removed, account's existing dev-domain untouched
- [x] Full existing test suite passes with no new failures (`bun run test`)

### DIFF
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -19,6 +19,8 @@
     "@hono/node-server": "^1.19.11",
     "@hono/node-ws": "^1.3.1",
     "@hono/typebox-validator": "^0.3.0",
+    "@ngrok/ngrok": "^1.7.0",
+    "@ngrok/ngrok-api": "^0.19.0",
     "@repo/adapters": "workspace:*",
     "@repo/core": "workspace:*",
     "@repo/db": "workspace:*",
--- a/apps/api/src/config/env.ts
+++ b/apps/api/src/config/env.ts
@@ -177,6 +177,20 @@ const envSchema = z.object({
    */
   OBLIEN_WEBHOOK_SECRET: z.string().optional(),
 
+  /* ---------- ngrok tunnel provider ---------- */
+  /**
+   * Agent authtoken (from the ngrok dashboard → "Your Authtoken").
+   * Required to open any tunnel, free or paid. A per-call override
+   * can be passed via TunnelProvisionInput.context.authtoken.
+   */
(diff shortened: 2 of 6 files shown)


Hydralerne:
which feature needs this ?

ishanj12:
Hi @Hydralerne ,

Today nothing calls this provider at runtime yet. Team-mode's only wired consumer (attachTunnelingLifecycle in index.ts) hardcodes provider: "oblien". ngrok/cloudflare are registered in registry.ts purely as stub entries so describeProviders() can list them as "known but not available" for UI dropdowns.

This PR turns the ngrok stub into a real, tested implementation so it's ready to be:

A selectable alternative to Oblien for team-mode tunneling (not locking self-hosted operators into the project's own tunnel service), and the base for the "user-defined tunnels" use case the module's own header already calls out as planned (webhook receivers, preview URLs, per-service "expose via tunnel" toggles).

Happy to also wire it into the team-mode provider selector in this PR or a follow-up if that's the bar for merging — let me know what you'd prefer.

PR #65, opened 2026-07-21
fix(api): paginate GitHub App installation repositories

woosal1337:
## Summary

`GitHubAppSource.listInstallationRepos` made a single `GET /installation/repositories?per_page=100` request, so an App installation with **more than 100 accessible repositories** was silently truncated to the first 100. Everything downstream — `getHome`, `listReposForOwner`, and the dashboard's repo picker — could only ever see those first 100, so a user's newer repos were unselectable.

## Fix

Walk every page (`per_page=100`, incrementing `page`) until the response's `total_count` is reached, hard-capped at 100 pages (10k repos) as a safety bound. `ghFetch` already serializes `params` into the GET query string, so no other change is needed.

## Testing

- `bun run --cwd apps/api lint` passes (`tsc --noEmit`).
- On an account with >100 repos accessible to the installation, the repo list now returns all of them instead of the first 100.

### DIFF
--- a/apps/api/src/modules/github/sources/app-source.ts
+++ b/apps/api/src/modules/github/sources/app-source.ts
@@ -93,11 +93,24 @@ export class GitHubAppSource implements GitHubSource {
       () => null,
     );
     if (!token) return [];
-    const data = await ghFetch<{ repositories: GitHubRepository[] }>(token, {
-      url: "https://api.github.com/installation/repositories",
-      params: { per_page: 100 },
-    });
-    return mapRepositories(data.repositories ?? []).map((r) => ({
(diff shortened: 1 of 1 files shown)


valgardg:
Would it be better to preserve pagination through the API and UI rather than eagerly fetching every repository? This fixes the current 100-repository truncation, but it instead moves the ceiling to 10k repos which can end in a really large amount of returned repos. Is there a reason the repo picker needs the entire list at once rather than supporting paginated or incremental loading?

Hydralerne:
please add the ui part at the end of /library page so we can make use of this

PR #71, opened 2026-07-21
fix(dashboard): move locale cookie outside client boundary

### LINKED ISSUE #70: dashboard: LOCALE_COOKIE resolves to RSC server-function stub in inline <script>, breaking JS parsing on every page
## Summary

On a fresh self-hosted deploy (`docker compose up -d --build` from `main`, commit `14ac98f` / v0.2.1, reproduced again after fast-forwarding to `f05cbf2`), the dashboard throws a `SyntaxError: missing ) after argument list` on **every page** at load time, at `<page-url>:21:396`. The error doesn't crash React/hydration, but it silently breaks other inline-script-dependent behavior — in my case, clicking "Connect Openship Cloud" on `/settings?tab=cloud` did nothing (no popup, no network request, no console error at click time) because a JS parse error earlier in `<head>` appears to interfere with event wiring.

## Root cause

`apps/dashboard/src/app/layout.tsx` imports `LOCALE_COOKIE` from `apps/dashboard/src/components/i18n-provider.tsx`, which is a `"use client"` module:

```ts
// layout.tsx (Server Component)
import { I18nProvider, LOCALE_COOKIE } from "@/components/i18n-provider";
...
<script
  dangerouslySetInnerHTML={{
    __html: `(function(){try{var m=document.cookie.match(/(?:^|;\s*)${LOCALE_COOKIE}=([^;]+)/);var l=m?decodeURIComponent(m[1]):(localStorage.getItem('${LOCALE_COOKIE}')||'');...})();`,
  }}
/>
```

In this build, the `${LOCALE_COOKIE}` template interpolation does **not** resolve to the string `"openship-locale"`. Instead it resolves to the stringified Next.js RSC boundary stub that gets generated when a client-module export is referenced somewhere it can't be safely serialized:

```
function(){throw Error("Attempted to call LOCALE_COOKIE() from the server but LOCALE_COOKIE is on the client. It's not possible to invoke a client function from the server, it can only be rendered as a Component or passed to props of a Client Component.")}
```

That string contains unescaped `"` and `'` characters, which breaks the surrounding template literal / inline script syntax, producing the `SyntaxError` seen in the console on every route.

I confirmed this by iterating every inline `<script>` in the rendered `<head>` through `new Function(script.textContent)` — the script at the locale-cookie call site is the only one that throws, and bisecting the throw position lands exactly inside the interpolated stub text.

## Reproduction

1. Clone `main`, `docker compose up -d --build` (self-hosted, `CLOUD_MODE=false`)
2. Open any dashboard page (`/`, `/settings`, `/library`, etc.)
3. Check browser console → `SyntaxError: missing ) after argument list` at `<url>:21:396`, reproducible on every load/route

Note: `LOCALE_COOKIE` is *also* used correctly on the server side a few lines earlier in the same file (`cookieStore.get(LOCALE_COOKIE)?.value`) — only the inline-`<script>` interpolation site is affected, suggesting this is specifically about referencing a client-module const inside a `dangerouslySetInnerHTML` template string versus using it as a plain server-side value.

## Fix I applied locally

Replaced the two `${LOCALE_COOKIE}` interpolations inside the inline script with the literal string `openship-locale` (matching the actual constant value), leaving the server-side `cookieStore.get(LOCALE_COOKIE)` usage untouched:

```diff
- __html: `(function(){try{var m=document.cookie.match(/(?:^|;\s*)${LOCALE_COOKIE}=([^;]+)/);var l=m?decodeURIComponent(m[1]):(localStorage.getItem('${LOCALE_COOKIE}')||'');...
+ __html: `(function(){try{var m=document.cookie.match(/(?:^|;\s*)openship-locale=([^;]+)/);var l=m?decodeURIComponent(m[1]):(localStorage.getItem('openship-locale')||'');...
```

After rebuilding, the SyntaxError is gone on every route and the Cloud-connect popup flow starts firing correctly on click.

A more robust upstream fix would probably avoid interpolating any client-module export directly into a server-rendered inline script string — e.g. define the cookie name as a plain constant in a non-`"use client"` module (or a `.ts` file with no directive) that both the layout and the client provider import, so there's no cross-boundary reference at all.

## Environment

- Self-hosted, Docker Compose, built from source
- Accessed over plain HTTP via a Tailscale IP (not HTTPS/localhost) — mentioning this in case it's somehow relevant to how Next resolves the boundary, though I don't see why it would be
- `NODE_ENV=production`, `CLOUD_MODE=false`, `DEPLOY_MODE=docker`
- Next.js 16.1.6, dashboard built with webpack (not Turbopack)

Rajioba1:
## Summary

- move `LOCALE_COOKIE` from the client-only i18n provider to the shared i18n module
- prevent Next.js from replacing the cookie value with an RSC client-reference stub in the server-rendered bootstrap script
- preserve the existing SSR, cookie, and local-storage locale behavior

## Verification

- existing dashboard tests still pass: `bun --cwd apps/dashboard test`
- TypeScript check passes: `bunx tsc --noEmit -p apps/dashboard/tsconfig.json`
- production webpack build passes: `bun --cwd apps/dashboard build`
- verified the emitted production bundle contains the locale cookie literal and no `LOCALE_COOKIE()` RSC server-function stub

No focused unit test was added because Vitest does not apply Next.js's production RSC transform; the reported failure was verified against the emitted production bundle instead.

## Scope

The touched files contain pre-existing Prettier differences on upstream; this PR intentionally avoids unrelated formatting churn. The existing `openship-locale` literal in `proxy.ts` is unchanged because it runs outside the affected RSC import boundary.

Fixes #70

### DIFF
--- a/apps/dashboard/src/app/layout.tsx
+++ b/apps/dashboard/src/app/layout.tsx
@@ -3,12 +3,19 @@ import { cookies, headers } from "next/headers";
 import "./globals.css";
 import { ThemeProvider, ThemeScript } from "@/components/theme-provider";
 import { ToastProvider } from "@/components/toast";
-import { I18nProvider, LOCALE_COOKIE } from "@/components/i18n-provider";
+import { I18nProvider } from "@/components/i18n-provider";
 import { AuthProvider } from "@/context/AuthContext";
 import { NetworkErrorHandler } from "@/components/network-error-handler";
 import { ModalProvider } from "@/context/ModalContext";
 import { DesktopChrome } from "@/components/desktop-chrome";
-import { defaultLocale, isRtl, loadDictionary, locales, type Locale } from "@/i18n";
+import {
+  defaultLocale,
+  isRtl,
+  loadDictionary,
+  LOCALE_COOKIE,
+  locales,
+  type Locale,
+} from "@/i18n";
 
 /** Resolve the request locale server-side: explicit cookie first, then the
  *  browser's Accept-Language, else the default. Keeps SSR and first paint in
--- a/apps/dashboard/src/components/i18n-provider.tsx
+++ b/apps/dashboard/src/components/i18n-provider.tsx
@@ -13,6 +13,7 @@ import {
   defaultLocale,
   isRtl,
   loadDictionary,
+  LOCALE_COOKIE,
   locales,
   type Dictionary,
   type Locale,
@@ -34,8 +35,6 @@ const I18nContext = createContext<I18nContextValue | null>(null);
 /* ------------------------------------------------------------------ */
 /*  Provider                                                           */
 /* ------------------------------------------------------------------ */
-
-export const LOCALE_COOKIE = "openship-locale";
 
 /** Read the chosen locale on the client: cookie first (what SSR also reads),
  *  then the localStorage mirror. Returns null when nothing valid is stored. */
--- a/apps/dashboard/src/i18n/index.ts
+++ b/apps/dashboard/src/i18n/index.ts
@@ -43,6 +43,7 @@ export type Dictionary = typeof baseDictionary;
 export const locales = ["en", "ar", "es", "fr", "de", "pt", "ja", "zh"] as const;
 export type Locale = (typeof locales)[number];
 export const defaultLocale: Locale = "en";
+export const LOCALE_COOKIE = "openship-locale";
 
 /** RTL languages. */
 const rtlLocales = new Set<Locale>(["ar"]);


ValenFelizia:
Tested successfully on Windows 11 with Bun 1.3.10 and Node 24.15.0.

On main, I reproduced the issue with a production webpack build:

- the emitted server bundle contained the Attempted to call LOCALE_COOKIE RSC stub;
- loading /login produced Uncaught SyntaxError: missing ) after argument list.

On this PR, the production build completed successfully, the RSC stub was no longer present in the emitted server bundle, and the browser syntax error no longer occurred. The openship-locale literal remains present as expected.

I did not validate API-dependent dashboard functionality.

Hydralerne (approved):

PR #90, opened 2026-07-22
fix: ignore outbound sockets in edge port preflight

### LINKED ISSUE #85: edge preflight: lsof fallback matches OUTBOUND connections, falsely reports port 443 as occupied
## Bug

`openship server install -c openresty` aborts with:

```
openresty failed: Ports 443 are in use by another service (unknown). Accept a migrate or takeover to continue.
```

even though **nothing listens on 443** on the server.

## Root cause

In `probeListeningPort` the port probe is:

```sh
ss -tlnp sport = :443 2>/dev/null | grep LISTEN || lsof -ti tcp:443 2>/dev/null || true
```

When `ss` finds no listener (grep exits 1), the **lsof fallback runs without a LISTEN filter**, so it matches any socket that *touches* port 443 — including **outbound** connections `server -> remote:443`. The PID of a process that merely makes HTTPS requests is then treated as the port-443 "occupant", classified as `unknown`, and the edge preflight raises `EdgeConflictError`.

Reproduced on a server running a gateway daemon with persistent outbound connections to 149.154.166.110:443 (Telegram): `lsof -ti tcp:443` returns that daemon's PID while `ss -tlnp sport = :443` is empty. Servers without such outbound connections install fine — which makes the failure look flaky/machine-dependent.

## Fix

Add the LISTEN state filter to the fallback:

```sh
lsof -ti tcp:${port} -sTCP:LISTEN
```

Verified locally: patching this one line in `dist/server/index.js` makes `server install -c openresty` succeed on the affected machines, and the resulting openresty passes `server check` as healthy.

## Environment

- CLI v0.2.1 (npm), desktop app v0.2.1, macOS arm64
- Target: Ubuntu 24.04

santhiprakash:
## Summary

Fixes #85. The edge port probe used `lsof -ti tcp:${port}` as a fallback when `ss` found no LISTEN socket. Because that fallback matched any TCP socket touching the port, an unrelated outbound HTTPS connection could make port 443 appear occupied and abort OpenResty installation.

The fallback now uses `lsof -ti tcp:${port} -sTCP:LISTEN`, so only listening sockets are treated as port occupants.

## Testing

- `bun run --cwd packages/adapters lint`
- `bunx vitest run src/runtime/port-conflict.test.ts` from `packages/adapters`
- Prettier check and `git diff --check`

The focused test simulates an empty `ss` result and an outbound-only lsof result, and verifies that no occupant is reported. The full adapter suite also ran: 100 tests passed and 11 unrelated tests failed because this local sandbox denies binding `127.0.0.1` with `EPERM` in the existing reverse-tunnel/reachability tests.

No existing test assertions were changed.

### DIFF
--- a/packages/adapters/src/runtime/port-conflict.test.ts
+++ b/packages/adapters/src/runtime/port-conflict.test.ts
@@ -1,0 +1,21 @@
+import { describe, expect, test } from "vitest";
+import type { CommandExecutor } from "../types";
+import { probeListeningPort } from "./port-conflict";
+
+describe("probeListeningPort", () => {
+  test("does not treat an outbound socket as a listener in the lsof fallback", async () => {
+    const commands: string[] = [];
+    const executor = {
+      exec: async (command: string) => {
+        commands.push(command);
+        if (command.startsWith("ss ")) return "";
+        if (command.includes("-sTCP:LISTEN")) return "";
+        if (command.startsWith("lsof ")) return "4242";
+        throw new Error(`unexpected command: ${command}`);
+      },
+    } as unknown as CommandExecutor;
+
+    expect(await probeListeningPort(executor, 443)).toBeNull();
+    expect(commands.some((command) => command.includes("-sTCP:LISTEN"))).toBe(true);
+  });
+});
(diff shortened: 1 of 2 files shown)


Hydralerne:
was fixed before sorry

PR #94, opened 2026-07-22
fix(desktop): package OpenResty Lua assets

### LINKED ISSUE #82: [desktop] openresty install fails: ENOENT /$bunfs/root/lua/site_logger.lua (lua files not embedded in desktop app binary)
## Bug

On the **macOS desktop app (v0.2.1, arm64)**, installing OpenResty on a server (`server install -c openresty`, or from the dashboard) fails at the analytics-script deploy step:

```
openresty failed: ENOENT: no such file or directory, open '/$bunfs/root/lua/site_logger.lua'
```

## Root cause

Same class of packaging bug as #42 (fixed for the npm package in #47), but for the **desktop app distribution**: the Bun-compiled `openship-api` binary inside `Openship.app/Contents/Resources/bin/` does not have the `lua/` files embedded in its $bunfs virtual filesystem, so `readLua('site_logger.lua')` fails.

The npm CLI on the same machine (`/opt/homebrew/lib/node_modules/openship/dist/server/lua/`) **does** contain all 5 lua files — so the fix from #47 never reached the desktop build.

## Repro

1. Install Openship desktop app v0.2.1 on macOS (arm64)
2. Add an SSH server (root user)
3. `openship server install -c openresty <id>`
4. Fails with the ENOENT above

## Workaround (verified)

Quit the desktop app, then run the npm CLI's own server against the app's data dir and install through it:

```
openship up --foreground --port 4000 --no-ui --data-dir "$HOME/Library/Application Support/Openship/data"
openship ctx add tmp --api-url http://localhost:4000 --token <same PAT> --use
openship server install -c openresty <id>   # succeeds
```

The npm-bundled server has the lua files, so the install completes and `server check` then reports openresty healthy.

Rish-it:
Fixes #82

## Bug

Installing OpenResty from the packaged macOS arm64 desktop app fails while reading `site_logger.lua`:

```text
ENOENT: no such file or directory, open '/$bunfs/root/lua/site_logger.lua'
```

## Root cause

The compiled API resolves the Lua scripts inside the Bun virtual filesystem, but `bun build --compile` does not embed files read at runtime. The CLI packaging path already stages these assets for #47; the desktop packaging path did not.

## Fix

- allow the adapter to read Lua scripts from `OPENSHIP_LUA_DIR`, preserving the existing source-relative fallback
- stage the existing OpenResty Lua directory as a desktop resource
- pass the packaged resource path to the compiled API process
- add a regression test that compiles and executes the public Lua deployment path

## Verification

- compiled-executable regression test passed in two independent pre-commit runs and one post-commit run
- adapters suite passed: 33 tests
- adapter and desktop TypeScript checks passed
- macOS arm64 Forge package completed; packaged Lua files exactly match the source directory
- `git diff --check` passed

### DIFF
--- a/apps/desktop/build/stage.ts
+++ b/apps/desktop/build/stage.ts
@@ -7,6 +7,7 @@
  *   resources/dashboard/              the dashboard's own Next standalone output
  *   resources/migrations/             drizzle .sql  → OPENSHIP_MIGRATIONS_DIR
  *   resources/pglite/                 pglite.wasm + pglite.data → OPENSHIP_PGLITE_ASSETS_DIR
+ *   resources/lua/                    OpenResty scripts → OPENSHIP_LUA_DIR
  *
  * Invoked by electron-forge's `generateAssets` hook (forge.config.js) and also
  * runnable standalone with `bun run build/stage.ts`. Must run under bun — it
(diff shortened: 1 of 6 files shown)


Hydralerne:
fixed already, thank you for your contribution

PR #175, opened 2026-07-23
fix(docker): create the first admin declaratively on compose deploys

### LINKED ISSUE #138: Bug: Fresh Docker deployment cannot create first admin (SIGNUP_DISABLED) because bootstrap request comes from Docker bridge instead of loopback
# Bug: Fresh Docker deployment cannot create first admin (`SIGNUP_DISABLED`) because bootstrap request is not detected as loopback

## Description

I believe I've found a bug in the Docker deployment bootstrap flow that prevents creating the first administrator account on a fresh self-hosted installation.

## Environment

- OpenShip commit: `ba6abc6`
- Fresh Docker Compose deployment
- Ubuntu VPS
- Docker bridge networking
- Reverse proxy (OpenResty/Nginx)
- Public URL configured

```env
OPENSHIP_PUBLIC_URL=https://openship.yashmishra.in
OPENSHIP_EXTRA_TRUSTED_ORIGINS=https://openship.yashmishra.in
```

---

## Initial issue

Initially signup failed with:

```json
{
  "error": "Origin not allowed",
  "code": "ORIGIN_REJECTED"
}
```

This was resolved by configuring:

```env
OPENSHIP_PUBLIC_URL=https://openship.yashmishra.in
OPENSHIP_EXTRA_TRUSTED_ORIGINS=https://openship.yashmishra.in
```

After that, the error changed to:

```json
{
  "error": "Public sign-up is disabled on this instance. Use your invitation link to join.",
  "code": "SIGNUP_DISABLED"
}
```

---

## Database state

Fresh installation.

```sql
SELECT COUNT(*) FROM "user";
```

returns:

```text
 count
-------
 0
```

No users exist.

---

## Relevant code

`apps/api/src/modules/auth/auth.routes.ts`

```ts
// Invite-only sign-up guard
authRoutes.on("POST", "/sign-up/*", async (c, next) => {
  if (isSaasDeployment) return next();

  const [anyUser] = await db
    .select({ id: schema.user.id })
    .from(schema.user)
    .limit(1);

  if (!anyUser && isLoopbackRequest(c)) return next();

  return c.json(
    {
      error: "Public sign-up is disabled on this instance. Use your invitation link to join.",
      code: "SIGNUP_DISABLED",
    },
    403,
  );
});
```

---

## Debugging performed

To understand why `isLoopbackRequest()` was failing, I added logging to `loopback-peer.ts`.

Health checks:

```text
GET /api/health
Peer address: ::ffff:127.0.0.1
```

Signup request:

```text
POST /api/auth/sign-up/email
Peer address: ::ffff:172.18.0.6
403 SIGNUP_DISABLED
```

`isLoopbackRequest()` is implemented as:

```ts
export function isLoopbackRequest(c: Context): boolean {
  return isLoopbackPeer(peerAddress(c));
}
```

where

```ts
peerAddress(c)
```

uses

```ts
getConnInfo(c).remote.address
```

which returns the kernel TCP peer.

---

## Why I think this is a bug

On Docker deployments, the signup request reaches the API through another container (dashboard/OpenResty), so the API sees the Docker bridge IP:

```text
172.18.x.x
```

instead of

```text
127.0.0.1
```

As a result,

```ts
isLoopbackRequest(c)
```

returns `false`.

Since there are no users yet, the bootstrap path is never taken and the application always returns:

```json
{
  "code": "SIGNUP_DISABLED"
}
```

This appears to make it impossible to create the first administrator account through the documented web flow on a fresh Docker deployment.

---

## Temporary workaround

Changing:

```ts
if (!anyUser && isLoopbackRequest(c))
```

to:

```ts
if (!anyUser)
```

allows the first administrator account to be created successfully.

After creating the first user, restoring the original code works because the bootstrap path is no longer needed.

---

## Expected behavior

A fresh Docker deployment should provide a supported way to create the first administrator account.

Possible solutions:

- Allow trusted internal proxy/container requests for the initial bootstrap.
- Provide a dedicated bootstrap endpoint or CLI command.
- Document the intended first-admin creation workflow for Docker deployments if there is already a supported method.

---

## Additional information

I may be missing the intended bootstrap process, so if this behavior is expected I'd appreciate guidance on how the first administrator account is supposed to be created on a fresh Docker deployment.

shani-singh1:
## Problem

On a fresh Docker Compose deployment there's no supported way to create the first admin through the web flow. Public sign-up is disabled on self-hosted instances, and the loopback bootstrap shortcut in the sign-up guard never fires because the request reaches the API from a Docker-bridge IP (e.g. `172.18.x.x`) rather than `127.0.0.1`, so it always returns `SIGNUP_DISABLED`. `openship up`'s interactive wizard has no Compose equivalent.

Fixes #138.

## Approach

This follows the direction discussed in #138 (declarative credentials at start, endorsed by @Hydralerne) and reuses the exact mechanism the CLI already relies on — the internal-token-gated, one-shot `POST /api/system/bootstrap-admin` endpoint. **Public sign-up is not touched.**

An optional one-shot `init-admin` Compose service runs once the API is healthy and creates the first admin from credentials the operator declares in `.env`:

```env
OPENSHIP_ADMIN_NAME=Admin
OPENSHIP_ADMIN_EMAIL=admin@example.com
OPENSHIP_ADMIN_PASSWORD=a-strong-password   # 8-128 chars
```

Properties:

- **Opt-in / no-op by default** — if the admin vars are unset the service prints a hint and exits `0`, so nothing changes for operators who create the admin another way. `depends_on` is one-directional; no other service waits on it.
- **No new public surface** — it POSTs to the same invite-only bootstrap endpoint the CLI wizard uses, authenticated with the existing `INTERNAL_TOKEN`. It does not enable public registration.
- **Idempotent** — `bootstrap-admin` returns `409` once an admin exists, which the script treats as success, so re-running `docker compose up` never mutates an existing admin.

## Changes

- `scripts/docker-init-admin.sh` — the bootstrap script (POSIX `sh`; JSON body is escaped for embedded quotes/backslashes).
- `docker-compose.yml` — the `init-admin` service (pinned `curlimages/curl`, mounts the script read-only, waits for `api` health, then exits).
- `.env.example` — documents the optional `OPENSHIP_ADMIN_*` vars.
- `docs/installation.md` — a "Create the first admin" note in the Docker section.

## Testing

- `docker compose config` renders cleanly; `init-admin` receives `INTERNAL_TOKEN` and `OPENSHIP_API_URL`.
- `sh -n` / `bash -n` syntax-check pass.
- Ran the script against a mock of the `bootstrap-admin` contract covering every branch: vars-unset skip, successful create (incl. a password with embedded `"` and `\` to confirm JSON escaping), idempotent `409` re-run, `401` wrong-token, missing-`INTERNAL_TOKEN` abort, and unreachable-API. Success/skip paths exit `0`; error paths exit non-zero without blocking the rest of the stack.

## Notes for reviewers

- The service is intentionally non-blocking for the rest of the stack (`restart: "no"`, nothing depends on it), so an error state surfaces as an `Exited (1)` container without affecting the API/dashboard.
- No application/TypeScript code changes — this only wires up an existing endpoint for the Compose path.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -59,6 +59,17 @@ BETTER_AUTH_URL=http://localhost:4000
 #   node -e "console.log(require('crypto').randomBytes(32).toString('hex'))"
 INTERNAL_TOKEN=change-me-32-byte-random-hex
 
+# ─── First admin (Docker Compose only, optional) ──────────
+# On the CLI (`openship`) the setup wizard creates the first admin. Docker has no
+# wizard, so set these to have the one-shot `init-admin` service create the first
+# admin on `docker compose up` (issue #138). Leave unset to create it another way
+# (CLI, or a first sign-in you enable manually). Public sign-up stays disabled —
+# this calls the same internal-token-gated bootstrap endpoint the CLI uses, and
+# is idempotent (it won't touch an admin that already exists). Password: 8-128 chars.
+# OPENSHIP_ADMIN_NAME=Admin
+# OPENSHIP_ADMIN_EMAIL=admin@example.com
+# OPENSHIP_ADMIN_PASSWORD=
+
 # ─── OAuth login (optional) ───────────────────────────────
 # GITHUB_CLIENT_ID=
 # GITHUB_CLIENT_SECRET=
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -107,6 +107,26 @@ services:
       api:
         condition: service_healthy
 
+  # ─── First-admin bootstrap (one-shot, optional) ───────
+  # Creates the first administrator declaratively so a fresh `docker compose up`
+  # is usable without the CLI wizard (issue #138). No-op unless OPENSHIP_ADMIN_*
+  # is set in .env; idempotent (bootstrap-admin is one-shot — 409 once an admin
+  # exists). Does NOT enable public sign-up — it calls the same internal-token-
+  # gated endpoint the CLI uses. Runs after the API is healthy, then exits.
+  init-admin:
+    image: curlimages/curl:8.11.1
+    restart: "no"
+    env_file:
+      - .env
+    environment:
+      OPENSHIP_API_URL: http://api:4000
+    volumes:
+      - ./scripts/docker-init-admin.sh:/docker-init-admin.sh:ro
+    entrypoint: ["sh", "/docker-init-admin.sh"]
+    depends_on:
+      api:
+        condition: service_healthy
+
   # ─── Web / Landing ────────────────────────────────────
   web:
     build:
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -64,11 +64,24 @@ cp .env.example .env
 docker compose up -d
 ```
 
+**Create the first admin.** Docker has no interactive setup wizard, so declare the
+first administrator in `.env` before starting — the one-shot `init-admin` service
+creates it once the API is healthy (public sign-up stays disabled):
+
+```bash
+OPENSHIP_ADMIN_NAME=Admin
+OPENSHIP_ADMIN_EMAIL=admin@example.com
+OPENSHIP_ADMIN_PASSWORD=a-strong-password   # 8-128 chars
+```
+
+It's idempotent — re-running `docker compose up` never changes an existing admin.
+Leaving these unset skips the step; you can still create the admin another way.
+
 ---
 
 ## Users & access (self-hosted)
 
-- **Public signup is disabled.** The first admin is created by `openship up`'s setup wizard.
+- **Public signup is disabled.** On the CLI the first admin is created by `openship up`'s setup wizard; on Docker, set `OPENSHIP_ADMIN_*` in `.env` (see [Docker](#docker) above).
 - **Invite teammates** from **Settings → Team**. They get an accept link at your instance's URL (so the instance needs to be reachable — a public URL or your LAN).
 - **Lost the admin password?** Run `openship reset-admin-password` on the box (no sign-in needed; uses the local internal token).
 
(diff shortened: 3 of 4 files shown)


shani-singh1:
Rebasing this on the latest `main` and wanted to check direction before I push an update, since a lot changed here since I opened it.

Two things I noticed:

1. **`openship up --compose --non-interactive` (with `--admin-password` / `OPENSHIP_ADMIN_PASSWORD`) now bootstraps the first admin via the internal token** — which is the same declarative-credentials idea this PR implements, but done in the CLI. If the CLI is the intended path for first-admin on compose, this PR's separate `init-admin` service is probably redundant and I'm happy to close it.

2. **The compose stacks were split** since I opened this (root `docker-compose.yml` → SaaS / from-source control plane; self-hosting → `docker/docker-compose.yml`). My `init-admin` service auto-merged into the root file, which is the wrong target now — public signup is on there, so it doesn't belong.

Where I'd like your read: is there still a gap for **raw `docker compose -f docker/docker-compose.yml up`** used *without* the CLI (CI / GitOps / anyone not running `openship up`)? For that path #138 still seems to bite — the interactive message and the install docs tell users to "register the first account in the dashboard," but on the split-container stack the dashboard proxies the signup to `api:4000`, so the API sees the dashboard's bridge IP rather than loopback and returns `SIGNUP_DISABLED` (matching the original #138 report).

If you'd like, I can either:
- **(a)** re-target this to add the declarative `init-admin` to `docker/docker-compose.yml` for the raw-compose path, or
- **(b)** drop the service and instead send a small docs fix pointing raw-compose users to `openship up --compose --non-interactive`, or
- **(c)** just close this if `openship up --compose` is meant to be the one and only way.

Happy to go whichever way you prefer — just don't want to push a duplicate of what the CLI now does.

PR #178, opened 2026-07-23
docs: correct published rate limits, route groups, MCP tools, and logs usage

ahmedhesham6:
Four places where the docs no longer match the shipped code. Each was checked against source rather than inferred.

### Rate limits were wrong by 3–5×

`docs/api/index.mdx` published `default-anon` at 100/min and `default-authed` at 600/min. `apps/api/src/lib/rate-limit/policies.ts:39,50` sets them to **300** and **3000**. Anyone sizing a script or CI job against the docs was throttling themselves far harder than necessary.

Also in that section:
- `auth-tight` covers the self-host invite-signup route (`apps/api/src/modules/system/system.routes.ts:47`), not just `POST /api/auth/*`.
- Documented the loopback exemption in `apps/api/src/middleware/rate-limiter.ts:75` — with neither `TRUST_PROXY` nor `OPENSHIP_PUBLIC_URL` set, loopback requests skip the ordinary limits while the login gate still enforces. This materially changes what a self-hoster should expect.

I deliberately left `auth-loose`, `read-authed`, and `write-authed` out of the table. They exist in `policies.ts` but no route opts into them — only `billing-portal`, `mcp`, and `webhook-ingress` do. Listing them would imply enforcement that isn't there.

### Route-groups table was missing four mounted modules

Added `/api/apps`, `/api/updates`, `/api/notices` (both modes) and `/api/jobs` (self-hosted only). Worth noting `app.ts:141` mounts jobs unconditionally — the gate is `localOnly` applied inside `job.routes.ts`, which is easy to misread.

### MCP tool table listed 7 of 14 modules

`docs/mcp.mdx` showed 7 route families; the API exposes MCP-annotated routes across 14 modules. Added rows for settings, backups, cloud, jobs, apps, and updates. Tool names are derived from `toolName()` in `apps/api/src/modules/mcp/mcp-tools.ts:79`, so they match what `tools/list` actually returns. `apps` has no REST reference page to link, so that cell says so plainly rather than pointing somewhere wrong.

### Two smaller corrections

- `docs/dashboard/settings.mdx` omitted **Discord** from the delivery channels. It shipped as a first-class channel; the API page was updated at the time, this one was not.
- `docs/cli/deploy.mdx` required a deployment ID for `openship logs`. `apps/cli/src/commands/logs.ts:40` made it optional, falling back to the linked project's latest deployment. Every example still passed one.

### Verification

- Internal `/docs` link check: 0 broken links, 0 orphan pages.
- Markdown table column counts consistent across every edited table.
- No `bun format` run: `package.json:36` scopes prettier to `{ts,tsx,js,jsx,json,md}`, and all four changed files are `.mdx`, so a repo-wide run would only touch unrelated files.

### Not addressed here

Larger gaps found while checking these, left for separate PRs: missing changelog entries for v0.2.3 and v0.3.0, no API reference pages for `jobs`/`apps`/`updates`/`notices`, and no reference for Openship's own environment variables (the code reads ~90; `.env.example` has 15).

### DIFF
--- a/apps/web/content/docs/api/index.mdx
+++ b/apps/web/content/docs/api/index.mdx
@@ -62,15 +62,17 @@ are per rolling minute.
 
 | Policy | Limit | Keyed by | Applies to |
 |---|---|---|---|
-| `default-anon` | 100 / min | IP | Any unauthenticated route |
-| `default-authed` | 600 / min | user | Any authenticated route |
-| `auth-tight` | 10 / min | IP | `POST /api/auth/*` (login, signup, reset) |
+| `default-anon` | 300 / min | IP | Any unauthenticated route |
+| `default-authed` | 3000 / min | user | Any authenticated route |
+| `auth-tight` | 10 / min | IP | `POST /api/auth/*` (login, signup, reset) and self-host invite signup |
 | `mcp` | 300 / min | IP | `/api/mcp` (tool-call bursts) |
 | `webhook-ingress` | 120 / min | source IP | Inbound webhook deliveries |
 | `billing-portal` | 20 / min | org | Stripe portal / checkout creation |
 
 Every response carries `X-RateLimit-Limit`, `X-RateLimit-Remaining`, and `X-RateLimit-Reset`. `/api/health` is
-never limited (load balancers and SSR poll it).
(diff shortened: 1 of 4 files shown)


Hydralerne:
thank you

Issue #181, opened 2026-07-24
The CLI package (apps/cli) ships without any test coverage


Rish-it:
## Problem

Every other workspace that holds logic has a `test` script backed by vitest — `packages/core`, `packages/adapters`, `packages/db`, `apps/api`, `apps/dashboard`. `apps/cli` is the exception: its only check is `lint` (`tsc --noEmit`). So the whole terminal surface — request building in `api-client`, the SSE wire parser, the self-host capability gate, and every command's method/path/body/output contract — has no regression net. A refactor that silently repoints an endpoint, drops the bearer header, or breaks JSON mode ships green.

## What's needed

A vitest suite for `apps/cli`, wired into the existing `turbo run test`, covering:

- the shared library helpers (`api-client`, `sse`, `caps`, `output`, `github-releases`) as pure units;
- the command actions driven through a network seam (fetch stubbed) so each asserts its exact request + rendered output + guard behavior, without a live API;
- a black-box smoke test that spawns the assembled binary for help/version/error-exit paths.

The suite must run offline (no live server, no SSH/DNS), stay fast, and fail loudly on real contract drift rather than pad a coverage number.

## Not in scope

Commands that require live infrastructure or interactive prompts (`up`, `install`, `login`, `doctor`, live `mail` deploy) and the local-config commands that need a $HOME seam (`context`, `config`) — a follow-up, once the harness lands.

Hydralerne:
merged

Hydralerne:
thank you

PR #186, opened 2026-07-24
test(cli): cover the domain command (domains, DNS, SSL)

Rish-it:
## Problem

The CLI test suite (#182) covered `server`, `token`, `deployment`, `project`, and the raw `api` passthrough, but `domain` — the largest remaining API-shaped command, 9 subcommands spanning custom domains, DNS verification, and SSL — still has no coverage beyond `tsc --noEmit`. Its contracts are exactly the kind that drift silently: the per-subcommand method/path/body, the `verify` 200-vs-422 handling, and the non-zero exits that gate scripting (`verify`/`verify-ssl` exit 1 when a domain isn't verified). A refactor that repoints a route, reshapes a body, or drops an unverified-exit ships green today.

Follows #181; builds on the harness merged in #182.

## Summary

`test/e2e/domain.test.ts` (27 tests) drives `domainCommand` over the same stubbed-socket harness the rest of the suite uses — the command action and the real `api-client` (URL building, `/api` prefix, auth header, `ApiError` mapping) run for real; only `fetch` and the config/caps seams are stubbed.

- **list** — `GET /domains?projectId=…` (project id URL-encoded), table vs JSON rows, empty result.
- **add** — `POST /domains { projectId, hostname, isPrimary }`, `--primary` toggles the flag, the `openship domain verify <id>` next-step hint, and the `{ domain, records }` JSON shape.
- **preview** — `POST /domains/preview { hostname }`, DNS-record table, JSON.
- **verify** — `POST /domains/:id/verify` via `apiRaw`: 200 verified → exit 0; **422 (DNS not propagated) → exit 1 without throwing**; json mode prints the body and does **not** exit; a non-422 (500) surfaces and exits 1.
- **primary** — `POST /domains/:id/primary`, JSON.
- **records** — `GET /domains/:id/records`, DNS-record table, JSON.
- **renew** — `POST /domains/:id/renew`, SSL status, JSON.
- **verify-ssl** — `POST /domains/:id/verify-ssl`: valid → exit 0; **not-valid → exit 1**; json mode does **not** exit.
- **renew-all** — `POST /domains/renew-all`, per-domain results table vs the "Nothing needed renewal" note, JSON.
- **auth + errors** — an API `{error}` is surfaced and exits 1; the bearer token is sent when logged in.

## Design notes

- **Offline and fast.** No live API/DNS; body and exit assertions read `fetchStub.calls` and the exit code, so they hold regardless of spinner rendering. Each test is a real regression sentinel, not a coverage number — verified by mutation: removing both unverified-exit lines from `domain.ts` fails exactly the two exit tests and nothing else.
- **Documented behavior finding (not fixed here):** unlike `server`/`mail`/`service`, `domain.ts` has no pre-flight login guard — when logged out it issues the request with no `Authorization` header and lets the API reject it, rather than short-circuiting. One test pins that (request still made, no auth header, API 401 surfaced) so adding a pre-flight guard later is a deliberate, visible change. Left as-is; out of scope for a test-only PR.
- **Reuses the merged harness** (`test/helpers/*`) and vitest runner from #182 — no runner or dependency changes, lockfile untouched.

## Verification

Run twice, clean, both green:

- `bun run --cwd apps/cli test` → 12 files, **76 tests** passed (49 prior + 27 domain) ×2
- `bun run --cwd apps/cli lint` (`tsc --noEmit`) → clean (exit 0)
- `bun install --frozen-lockfile` → no changes (lockfile matches)
- mutation check: neutralizing the two `!verified → process.exit(1)` lines fails exactly the `verify` 422 and `verify-ssl` invalid tests, then reverted.

Reproduced against committed `HEAD`: working tree matches `HEAD`, full suite green.

## Commits

Ten atomic commits, one per subcommand plus the shared auth/error concern; each adds a self-contained `describe` block and is green in isolation:

```
test(cli): cover domain list over a mocked API
test(cli): cover domain add and its DNS-record hints
test(cli): cover domain preview DNS-record output
test(cli): cover domain verify 200/422/error exit paths
test(cli): cover domain primary
test(cli): cover domain records
test(cli): cover domain renew
test(cli): cover domain verify-ssl valid/invalid exit paths
test(cli): cover domain renew-all org-wide results
test(cli): cover domain auth header and API-error exit paths
```

## Not in scope

Commands that need a new test seam the harness doesn't yet provide: `lib/service` (ssh2/dockerode) for `stop`/`service`/`status`, a `child_process` seam for `deploy`/`doctor`/`install`/`update`, an interactive `@clack` stdin seam for `up`/`wizard`/`reset-admin`, and a `$HOME` config-file seam for `context`/`config`/`cache`/`backup`/`system`/`login`/`logout`/`init`. Each seam is a separate follow-up. `mail install` is covered on its own branch pending #141.

### DIFF
--- a/apps/cli/test/e2e/domain.test.ts
+++ b/apps/cli/test/e2e/domain.test.ts
@@ -220,3 +220,31 @@ describe("openship domain renew", () => {
     expect(JSON.parse(out)).toEqual(SSL.data);
   });
 });
+
+// ─── verify-ssl (recheck only, no reissue) ───────────────────────────────────
+
+describe("openship domain verify-ssl", () => {
+  it("POSTs /domains/:id/verify-ssl and exits 0 when the cert is valid", async () => {
+    fetchStub = stubFetch(() => ({ json: { data: { domain: "app.example.com", sslStatus: "valid", verified: true } } }));
+    const { err, code } = await runCommand(domainCommand, ["verify-ssl", "d1"]);
+    expect(code).toBe(0);
+    expect(fetchStub.calls[0].method).toBe("POST");
+    expect(fetchStub.calls[0].url).toBe(`${API}/domains/d1/verify-ssl`);
+    expect(err).toContain("status:  valid");
+  });
+
+  it("exits 1 when the certificate is not valid yet", async () => {
+    fetchStub = stubFetch(() => ({ json: { data: { domain: "app.example.com", sslStatus: "pending", verified: false } } }));
+    const { code } = await runCommand(domainCommand, ["verify-ssl", "d1"]);
+    expect(code).toBe(1);
+  });
+
+  it("does NOT exit 1 in json mode even when the cert is not valid", async () => {
+    setJsonMode(true);
+    const data = { domain: "app.example.com", sslStatus: "pending", verified: false };
+    fetchStub = stubFetch(() => ({ json: { data } }));
+    const { out, code } = await runCommand(domainCommand, ["verify-ssl", "d1"]);
+    expect(code).toBe(0);
+    expect(JSON.parse(out)).toEqual(data);
+  });
+});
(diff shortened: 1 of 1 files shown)


Hydralerne:
amazing fr, goat these are important tests

PR #189, opened 2026-07-24
fix(api/dashboard): add notification channel test/verify flow

### LINKED ISSUE #179: Notification channels created in the dashboard can never be verified, so they never deliver
### Summary

Every notification channel kind except `in_app` is created with `verified: false`, and the dispatcher skips unverified channels. There is no API route and no dashboard UI that can flip that flag, so channels created through the dashboard silently never deliver anything.

### Steps to reproduce

1. Settings → Notifications → Add channel
2. Pick Slack (or Discord / MS Teams / webhook / email), paste a valid webhook URL, save
3. Channel is listed with an **Unverified** badge
4. Subscribe it to `deploy.failed`
5. Trigger a failing deploy

**Expected:** the message arrives.
**Actual:** nothing arrives, and there is no error anywhere — see below, a `notification_delivery` row is never even created.

### Root cause

`createChannel` auto-verifies only in-app channels — `apps/api/src/modules/notifications/notifications.controller.ts:66-69`:

```ts
// In-app is always verified — nothing to prove. Other kinds require
// explicit verification (test email/webhook/slack), set via PATCH.
verified: body.kind === "in_app",
```

The dispatcher skips anything unverified — `apps/api/src/lib/notification-dispatcher.ts:64`:

```ts
if (!channel || !channel.enabled || !channel.verified) continue;
```

This `continue` runs *before* `repos.notificationDelivery.create(...)`, so the failure leaves no delivery row, no `lastError`, and no log line. It's invisible from the dashboard.

The comment says verification is "set via PATCH", and `updateChannel` does accept it (`notifications.controller.ts:102`):

```ts
if (typeof body.verified === "boolean") updates.verified = body.verified;
```

But nothing ever calls it:

- **No test/verify route.** `notifications.routes.ts` registers only categories, channels CRUD, subscriptions, defaults, and deliveries. There is no `/channels/:id/test`.
- **The dashboard never calls `updateChannel`.** `notificationsApi.updateChannel` is defined in `apps/dashboard/src/lib/api/notifications.ts` and accepts `verified`, but has zero callers in the cod
ebase. `NotificationsTab.tsx` only renders the badge — there is no button or action attached to it.

So the only way to make any channel deliver is hand-crafting `PATCH /api/notifications/channels/:id {"verified": true}` — and self-attesting the flag doesn't prove reachability, which is what the s
chema intended it to mean (`packages/db/src/schema/notification.ts`):

```
/** True only after we've proven the channel is reachable (email
 *  click-through, webhook returned 2xx, slack test message landed). */
```

None of those three proof paths are implemented.

### Impact

Affects every non-in-app kind: `email`, `webhook`, `slack`, `discord`, `msteams`. The recently added Discord (#164) and Microsoft Teams (#173) channels are both unreachable through the UI for this
reason.

### Proposed fix

Add `POST /api/notifications/channels/:id/test`, tagged `notifications:write`, which:

1. builds a synthetic delivery for the channel,
2. sends it through the channel's existing worker — the `WORKERS` registry in `notification-workers.ts` already keys `sendEmail` / `sendWebhook` / `sendSlack` / `sendDiscord` / `sendMsTeams` by kin
d, all with the same `(delivery, channel)` signature, so no worker changes are needed,
3. sets `verified: true` on success and returns the provider's error message on failure,

plus a **Send test** action on each unverified channel row in `NotificationsTab.tsx`.

### Open question for maintainers

Should `POST /channels/:id/test` become the *only* way to set `verified` — i.e. should `updateChannel` stop honouring a client-supplied `verified` boolean? Self-attestation defeats the purpose of the flag, but dropping it is a breaking API change for anyone currently scripting around this gap. Happy to go either way.

I'd like to pick this up if the approach sounds right.

santhiprakash:
## Problem

Every non-`in_app` notification channel is created with `verified: false`, and the dispatcher skips unverified channels. There was no API route or dashboard action that could flip `verified`, so Slack/Discord/Teams/email/webhook channels created through the UI silently never delivered anything.

Fixes #179.

## Solution

- Add `POST /api/notifications/channels/:id/test` (tag `notifications:write`).
- The endpoint builds a synthetic test delivery and routes it through the existing channel worker (`sendChannelTestMessage` in `lib/notification-workers.ts`) so the test uses the same transport path as real notifications.
- On success it sets `verified: true` and refreshes `lastDeliveredAt`; on failure it returns `{ ok: false, error: ... }` with the provider's error message.
- The Settings → Notifications UI now shows a "Send test" (play) button on unverified external channels and refreshes the list after a successful verify.

## Notes

- `in_app` channels are still always verified and do not show a test button.
- Self-attestation via `PATCH` is left as-is; this PR focuses on the missing verification path, not removing the old one.
- No worker changes were needed — the existing `email`/`webhook`/`slack`/`discord`/`msteams` workers already accept the synthetic test delivery.

## Verification

- `bun --cwd apps/api test` — 63 test files / 750 tests passed.
- `bunx tsc --noEmit` in `apps/dashboard` passed.
- Added `apps/api/test/modules/notifications/channel-test.test.ts` covering missing id, ownership check, `in_app` rejection, success path, and failure path.


santhiprakash:
Closing this because current upstream/main already contains the notification channel test/verify flow, and rebasing this branch now creates unnecessary conflict churn across unrelated files. If a new gap appears after current upstream changes settle, we can re-open with a fresh, narrower PR.

PR #196, opened 2026-07-24
fix: retry failed SSL renewals and preserve login context endpoints

Rish-it:
A transient SSL renewal failure changed the certificate row to `error`, which excluded it from both automatic and organization-wide renewal selection. Re-authenticating a named CLI context also replaced its saved endpoints with localhost when URL flags were omitted.

The renewal selectors now retry due certificates in either `active` or `error` state. Organization renewal uses the same configured renewal window and TLS ownership helper as the scheduler, respects per-domain access, and reports an unverified provider result as failed. Unissued, fresh, and externally managed certificates are excluded.

CLI login resolves each endpoint from the explicit flag, then the named context’s saved value, then the local default. It keeps validation through the shared SDK HTTP client and saves changes only after the replacement token is accepted.

Validation: 47 API/TLS/server-removal cases, 17 CLI cases, and a real PGlite selector regression pass. API and CLI TypeScript pass. Six regressions fail against current main.

This PR targets the bug integration branch for #892.


Baaaki:
Ran into this too. Fix looks solid, thanks.

PR #208, opened 2026-07-24
fix: preserve desktop MCP OAuth consent through login

### LINKED ISSUE #119: Desktop: external MCP clients can't complete OAuth — metadata advertises dev port 4000, and the only fix (OPENSHIP_PUBLIC_URL) disables zero-auth (catch-22)
Connecting an external MCP client (Claude Code: `claude mcp add --transport http openship http://127.0.0.1:<apiPort>/api/mcp`) to the desktop app fails at the OAuth step, and I believe **no working configuration exists** in v0.3.0 — the pieces block each other:

**1. Discovery metadata advertises the dev port, not the real API origin.**
`/.well-known/oauth-authorization-server` and `/.well-known/oauth-protected-resource` return issuer / authorization_endpoint / token_endpoint at `http://localhost:4000` (the `runtimeTarget.api` fallback) while the desktop API actually runs on the dynamic port (e.g. `54777`). Nothing listens on 4000, so the client aborts with "Unable to connect" right after successfully reaching `/api/mcp` itself.

```
$ curl -s http://127.0.0.1:54777/.well-known/oauth-authorization-server | jq '{issuer, authorization_endpoint}'
{
  "issuer": "http://localhost:4000",
  "authorization_endpoint": "http://localhost:4000/api/auth/mcp/authorize"
}
```

**2. The only knob that fixes the metadata kills desktop auth entirely.**
Launching the app with `OPENSHIP_PUBLIC_URL=http://127.0.0.1:54777` in its environment makes the metadata correct (the API env spawn inherits it) — but `zeroAuthAllowed()` in `apps/api/src/middleware/zero-auth-guard.ts` rejects on `env.OPENSHIP_PUBLIC_URL` (`"cli-managed-or-publicly-served"`), which disables zero-auth — the desktop's only session mechanism. `/api/auth/desktop-login` then bounces to the login page and the authorize flow can never find a session. Catch-22.

**3. Independent of (2): the no-session fallback 404s.**
The authorize endpoint redirects sessionless requests to `/login` **on the API origin** (`http://127.0.0.1:54777/login` → 404). The login UI lives on the dashboard port, so even a user willing to sign in has nowhere to do it.

```
$ curl -s -o /dev/null -L --max-redirs 6 -w "final: %{url_effective} [%{http_code}]\n" \
    "http://127.0.0.1:54777/api/auth/mcp/authorize?response_type=code&client_id=x&redirect_uri=http%3A%2F%2Flocalhost%3A12345%2Fcb&code_challenge=abc&code_challenge_method=S256&scope=openid"
final: http://127.0.0.1:54777/login?...&prompt=consent [404]
```

**Suggested fixes:**

- (a) Derive the advertised OAuth origin from the incoming request's Host header for loopback requests — or have `apps/desktop/src/main/services.ts` pass a dedicated env (e.g. `OPENSHIP_ADVERTISED_ORIGIN: apiOrigin`) that feeds URL construction but **not** the publicly-served auth gate.
- (b) In the MCP authorize handler, when `zeroAuthAllowed()` passes, auto-mint the local session (same path as `desktop-login`) instead of redirecting to `/login`. This alone would make external MCP clients work zero-config on desktop.
- (c) Prefer `127.0.0.1` over `localhost` in the advertised URLs: the API binds IPv4-only (`OPENSHIP_API_HOST=127.0.0.1`), and clients that resolve `localhost` → `::1` first (e.g. Bun's fetch) get connection-refused before OAuth even starts.

**Env:** macOS arm64 (Apple Silicon), Openship.app v0.3.0. Happy to test a fix build

oshtz:
Closes #119

## Summary

Current `main` now advertises the packaged desktop app's dynamic API and dashboard origins correctly. One part of the MCP OAuth flow remains broken in zero-auth mode: creating the local session drops the pending consent return path, and `/mcp/authorize` is not accepted by the existing safe `returnTo` allowlist.

This change:

- preserves the validated MCP consent path when the desktop login page requests a zero-auth session
- relays that path through `/api/auth/desktop-login` and back through the dashboard's existing redirect validator
- allows `/mcp/authorize` in the safe relative-path allowlist while continuing to reject external redirects
- adds focused regression coverage for the MCP consent return path

This leaves `main`'s `OPENSHIP_ADVERTISED_ORIGIN` implementation unchanged.

## Testing

- dashboard test suite: 27 tests pass
- API and dashboard TypeScript checks pass
- external and protocol-relative `returnTo` values remain rejected

### DIFF
--- a/apps/api/src/modules/auth/auth.controller.ts
+++ b/apps/api/src/modules/auth/auth.controller.ts
@@ -153,6 +153,16 @@ export async function desktopLogin(c: Context) {
   });
   await setSessionCookie(c, session.token, session.expiresAt);
 
+  // Return through the dashboard's login layout when an OAuth flow is pending.
+  // That layout owns the returnTo allowlist, so this endpoint only relays the
+  // value as an encoded query parameter and never redirects to it directly.
+  const returnTo = c.req.query("returnTo");
+  if (returnTo) {
+    const loginUrl = new URL("/login", `${dashboardUrl}/`);
+    loginUrl.searchParams.set("returnTo", returnTo);
+    return c.redirect(loginUrl.toString());
+  }
+
   return c.redirect(dashboardUrl);
 }
 
--- a/apps/dashboard/src/app/(auth)/login/page.tsx
+++ b/apps/dashboard/src/app/(auth)/login/page.tsx
@@ -122,9 +122,12 @@ function LoginPageInner() {
   if (authMode === "none") {
     const apiUrl = getApiOrigin(typeof window !== "undefined" ? window.location.origin : undefined);
     // Redirect to the desktop-login endpoint which creates a real
-    // session cookie and redirects back to the dashboard.
+    // session cookie and redirects back to the dashboard. Preserve a
+    // validated OAuth consent return path through that API round-trip.
     if (typeof window !== "undefined") {
-      window.location.href = `${apiUrl}/api/auth/desktop-login`;
+      const desktopLoginUrl = new URL("/api/auth/desktop-login", apiUrl);
+      if (postLoginUrl) desktopLoginUrl.searchParams.set("returnTo", postLoginUrl);
+      window.location.href = desktopLoginUrl.toString();
     }
     return (
       <AuthShell>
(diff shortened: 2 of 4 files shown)


Hydralerne:
cool

oshtz:
Closing as superseded by #215. Current `main` now advertises the desktop's dynamic API origin and `forceMcpConsent` mints the loopback zero-auth session before redirecting to `/mcp/authorize`, so the login `returnTo` round-trip added here is no longer part of the normal MCP OAuth flow.

PR #222, opened 2026-07-24
Webmail: block remote CSS fetches when remote images are off

### LINKED ISSUE #220: Webmail: CSS url()/@import bypasses remote-image blocking, leaking a read receipt with images off
With **remote images turned off**, a sender can still get a read receipt: the sanitizer allows `<style>`, and the image blocker only rewrites ``, so CSS `url()` and `@import` fetch remote resources anyway.

Verified on `main` at 5236172.

> **Note:** an earlier revision of this issue reported two further sanitizer findings. On closer investigation one was not exploitable (a percent-encoded `javascript:` scheme resolves as a relative path, not a scheme) and another was contained by the client's shadow DOM (selector-based exfiltration cannot reach the app's light DOM). This issue is now scoped to the one finding that reproduces.

## Reproduction

Send a message whose body contains both a CSS-referenced resource and a normal tracking image:

```html
<style>body { background: url(https://tracker.example/pixel?u=victim) }</style>
<img src="https://tracker.example/p.gif">
```

Open it in the webmail read pane with **"load remote images" off**. Output from `processEmailContent`:

```html
<style>body{background:url(https://tracker.example/pixel?u=victim)}</style>
<img src="data:image/gif;base64,R0lGODlhAQABAAAA…" />
```

The `` is neutralized exactly as designed. The CSS pixel loads regardless.

## Cause

`processEmailContent` — `apps/email/server/src/trpc/routes/mail.ts:256-266` — implements the privacy control with a regex over `` tags only:

```ts
const blocked = clean.replace(/<img\b([^>]*?)\bsrc\s*=\s*("[^"]*"|'[^']*')/gi, …);
```

`sanitize.ts:10-22` allows `style` in `ALLOWED_TAGS` and `style` in `allowedAttributes`, and nothing processes CSS text. So every CSS-referenced URL survives untouched: `url(…)` in a `<style>` block, the same inside an inline `style` attribute, and `@import url(…)`.

## Impact

The setting silently fails to do what it says. A sender learns the message was opened, when, and the recipient's IP address — the exact things "block remote images" exists to withhold. That is worse than not offering the control, because the UI reports the user is protected.

No CSP mitigates this: `server/src/main.ts:42-57` deliberately leaves CSP unset for the API, and the CSP meta tag in `client/lib/email-utils.client.tsx:160` applies to generated *outgoing* mail, not the read pane.

## Scope

The read pane renders into an **open ShadowRoot** (`client/components/mail/mail-content.tsx:97-114`), not the app's light DOM. That scopes sender selectors, so CSS-based exfiltration of app inputs is **not** a concern here. It does not stop network requests from CSS, which is why this bug stands.

## Suggested fix

Strip URL-fetching constructs — `url(…)` and `@import` — from both `<style>` contents and inline `style` attributes, leaving colors, fonts, spacing and layout intact so ordinary HTML mail still renders. Ideally gate the stripping on the same `shouldLoadImages` flag, so enabling images restores CSS-referenced resources too instead of the two controls disagreeing.

Two notes for whoever picks this up:

- `allowedStyles` alone does **not** work. It filters inline `style` attributes only and has no effect on `<style>` tag contents:

  | input | output |
  | --- | --- |
  | `<p style="color:red; background: url(javascript:alert(1))">` | `<p style="color:red">` ✅ |
  | `<style>.a{background:url(javascript:alert(1))}</style>` | unchanged ❌ |

- Removing `style` from `allowedTags` would fix it but visibly breaks HTML newsletters that depend on `<style>` blocks. Not worth it given the shadow-root scoping.

I have a failing repro ready to include as a regression test and am happy to send the PR.

**Coordination:** #219 currently pins the existing behavior in `sanitize.test.ts` ("retains style tags and inline style as current behavior"). If this fix is taken, that assertion needs updating. I will branch the fix from `main` and flag it in both PRs so whichever merges second is an expected small conflict rather than a surprise.

ahmedhesham6:
With remote images disabled, current webmail still fetches CSS tracking URLs through escaped `url()` names/schemes, `image-set()` string candidates, and CSS-variable substitution. Relative image URLs also bypass its HTTP-only check. This repair updates the existing contributor PR for the current mail server and integrates it into the bug batch in #892.

`blockRemoteContent` remains the single policy entry point after HTML sanitization. It uses the existing parsed HTML tree and a CSS syntax tree, preserving the current XSS protections instead of reinstating style blocks. Fetching URLs and image-set candidates are neutralized; inline attachments, literal text, ordinary layout/color declarations, and inline image-set MIME metadata survive. The srcset scanner preserves commas inside data URLs. CSS variables/attribute substitutions and unparseable styles fail closed while remote images are blocked; explicitly enabling images uses the original sanitized styles.

Validation: 77 mail-server regression cases and TypeScript pass; 11 regressions fail against main. An isolated Chromium read-pane probe reproduced four bypasses on main, then confirmed zero network requests with blocking enabled and successful positive controls with images enabled. Tracker requests were intercepted and aborted locally.

Fixes #220 when this reaches main. This PR targets the integration branch, not main.

### DIFF
--- a/apps/email/server/src/lib/sanitize.ts
+++ b/apps/email/server/src/lib/sanitize.ts
@@ -39,3 +39,212 @@ export function sanitizeMailHtml(input: string): string {
     },
   });
 }
+
+/* ─── Remote content blocking ──────────────────────────────────────────── */
+/*
+ * With "load remote images" off, nothing in the message may cause a network
+ * fetch — `<img src>` alone never covered that: `srcset`, `background:
+ * url(…)`, `@import` and `image-set()` all fetch too, and each is enough to
+ * collect a read receipt.
+ *
+ * Input contract: already through `sanitizeMailHtml`. sanitize-html balances
+ * tags, deduplicates attributes, and re-encodes only `&amp; &lt; &gt; &quot;`
+ * in attribute values — the attribute handling below relies on that.
+ *
+ * `<style>` bodies reach this code raw, so the comment and @import passes are
+ * linear scanners: the natural regexes backtrack quadratically on crafted
+ * input (`/*a/*a…`, `@import url(url(…`), a denial-of-service lever server-side.
+ */
+
+// data: is self-contained and cid: is already delivered — neither leaves the
+// client, which is what keeps inline attachments rendering while blocked.
+// Single definition so the img, srcset, and CSS passes can't drift.
+const INLINE_SCHEMES = 'data:|cid:';
+const INLINE_URL = new RegExp(`^(?:${INLINE_SCHEMES})`, 'i');
+
+// `\75 rl(…)` and `@\69 mport` are valid spellings of `url(` / `@import`.
+// Decode only escapes of ASCII alphanumerics: those are what spell a fetch
+// construct, and decoding an escaped quote would corrupt its string.
+const CSS_ESCAPE = /\\([0-9a-f]{1,6})(?:\r\n|[ \t\r\n\f])?|\\([^\r\n\f0-9a-f])/gi;
+
+function decodeCssEscapes(css: string): string {
+  return css.replace(CSS_ESCAPE, (match, hex?: string, ch?: string) => {
+    // Range-check before fromCharCode: it masks to 16 bits, so a huge escape
+    // like \10041 must not alias down to 'A'.
+    const cp = hex !== undefined ? parseInt(hex, 16) : -1;
+    const decoded = cp >= 0x30 && cp <= 0x7a ? String.fromCharCode(cp) : ch ?? '';
+    return /^[0-9a-z]$/i.test(decoded) ? decoded : match;
+  });
+}
+
+// A comment is whitespace to the CSS tokenizer, so replace with a space —
+// otherwise `url(/*x*/https://…)` hides the fetch from the patterns below.
+// An unterminated comment runs to end of input, per spec.
+function stripCssComments(css: string): string {
+  let out = '';
+  let i = 0;
+  while (true) {
+    const open = css.indexOf('/*', i);
+    if (open === -1) return out + css.slice(i);
+    out += css.slice(i, open) + ' ';
+    const close = css.indexOf('*/', open + 2);
+    if (close === -1) return out;
+    i = close + 2;
+  }
+}
+
+// An import exists only to fetch, so drop the whole statement: to the first
+// `;` (consumed) or `}` (kept, it closes the block) outside quotes/parens,
+// so `@import "a;b"` leaves no trailing garbage.
+function stripImports(css: string): string {
+  const lower = css.toLowerCase();
+  let out = '';
+  let copyFrom = 0;
+  let search = 0;
+  while (true) {
+    const at = lower.indexOf('@import', search);
+    if (at === -1) return out + css.slice(copyFrom);
+    // `-` continues a CSS ident: `@import-fake` is a different at-keyword.
+    if (/[\w-]/.test(css[at + 7] ?? '')) {
+      search = at + 7;
+      continue;
+    }
+    out += css.slice(copyFrom, at);
+    copyFrom = search = importEnd(css, at + 7);
+  }
+}
+
+function importEnd(css: string, i: number): number {
+  while (i < css.length) {
+    const c = css[i];
+    if (c === '"' || c === "'") {
+      for (i++; i < css.length && css[i] !== c; i++);
+      i++;
+    } else if (c === '(') {
+      for (i++; i < css.length && css[i] !== ')'; i++);
+      i++;
+    } else if (c === ';') {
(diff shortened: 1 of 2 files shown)


Hydralerne:
thank you for your support
how ever, i found some issues in the pr

- Bypasses (remote fetch survives, no banner)

1. image-set() bare string containing ; { or } - HIGH
<style>a{background:image-set("http://evil.com/a;b" 1x)}</style>   → unchanged, blocked=false
CSS_IMAGE_SET = /…image-set\([^;{}]*/ truncates its span at the first ;/{/} - but those are ordinary characters inside a CSS string token, so the browser parses the whole URL and fetches. Worst part: the realistic data: fallback + remote sibling hits this every time (the ; in ;base64 truncates the span before the remote candidate):
image-set("data:image/png;base64,AAA" 1x, "http://evil.com/a" 2x) → remote 2x leaks

2. image-set() via escaped hyphen - HIGH
<style>a{background:image\2dset("http://evil.com/x" 1x)}</style>   → unchanged, blocked=false
CSS_IMAGE_SET keys on the literal substring image-set(; image\2dset( doesn't match, but the browser resolves \2d→- while building the ident and forms a real image-set( function. (The url() form is safe here - CSS_REMOTE_URL is ident-independent - so only the bare-string form leaks.)

3. Decoy src= inside alt/title shields the real remote src - HIGH
x src='data:,'   → unchanged, blocked=false
The src rewrite /(\bsrc\s*=\s*)("[^"]*"|'[^']*')/i is non-global, and \b matches the src= inside the alt value (the space before it is a word boundary). It rewrites that first (inline, so it's left as-is) and stops - the genuine src="https://evil.com/track" is never inspected. This is the nastiest one: it breaks the plain  case, which is the whole original feature. (data-src is safe only because the allowlist strips it - this is the same class of bug the allowlist happened to cover, but alt/title are preserved verbatim.)

also Correctness (false positives / broken render)

- D1 (med): inline image-set("data:a" 1x, "data:b" 2x) → mangled to image-set("data:a"data:,"data:b" 2x) and falsely blocked=true. CSS_REMOTE_STRING re-anchors on the closing quote of an exempt string. Destroys a legit zero-fetch retina background - violates the PR's own "inline data: survives" invariant.
- D2 (low): content:"@import" (the literal substring in a string) → stripImports deletes to EOF, unbalancing the block; blocked=true false positive. stripImports doesn't skip string/url() contents.
- D3 (low): fill:url(#grad) (same-document fragment ref, never fetches) → rewritten to none + blocked=true false banner.


suggested fix :

- image-set: replace the [^;{}]* span regex with a quote-aware, balanced-paren scanner (like the existing importEnd) to find the function's true extent; and normalize the ident (decode escaped -, or match image-set allowing escapes) before comparing.
- : stop regexing src out of the raw tag string - parse the tag's attributes (or at minimum require a real attribute boundary and make it global) so a src= inside alt/title can't shadow the real one.
- false positives: exempt bare url(#fragment); make CSS_REMOTE_STRING not span across an exempt string's boundary; make stripImports skip string/url() contents.

ahmedhesham6:
Thanks for the feedback, will update the gaps

ahmedhesham6:
Thanks for the sharp review — all six reproduce on the previous revision, and all six are fixed in faa955d.

**Bypasses:**
1. **image-set() span truncation at `;`/`{`/`}`** — the span is now found by a quote-aware scan to the balancing paren (same approach as `importEnd`), so string contents can't end it early. Your `data:…;base64 fallback + remote 2x` case now blocks the remote candidate and keeps the data: one.
2. **`image\2dset(`** — escape decoding now includes `\2d` (`-`) alongside alphanumerics, so the ident normalizes to `image-set(` before matching. Same reasoning as the alnum escapes: `-` means the same thing in an ident and in a string, so decoding is safe there too.
3. **Decoy `src=` in alt/title** — `blockImgTag` no longer regexes `src=` out of the raw tag text. It walks the tag's attributes sequentially (name = value pairs), so only a real `src`/`srcset` attribute is inspected and a lookalike inside another attribute's value can't shadow it.

**False positives:**
- **D1** — image-set strings are rewritten by a sequential token scan instead of a regex, so it can't re-anchor on the closing quote of an exempt string; `image-set("data:a" 1x, "data:b" 2x)` now passes through byte-identical and unflagged.
- **D2** — `stripImports` skips string contents, so `content:"@import"` is left alone.
- **D3** — `url(#fragment)` is exempt in the lookahead; a same-document reference never fetches.

All previous behavior holds (34 checks including your five cases, the original issue repro, adversarial-input timing on the new scanners, and an end-to-end pass through the sanitizer).

Hydralerne:
perfect, gonna check last time then merge, ty

Hydralerne:
also are you on x ? would love to know your handle :"

ahmedhesham6:
I just DMed you

PR #223, opened 2026-07-24
fix(email): populate mailbox display name for webmail sender (#218)

### LINKED ISSUE #218: Bug: Webmail sends From with a bare address — mailbox display name is never populated
Sending mail from the bundled webmail (v0.3.0, `@zero/dist`) produces a `From` header with no display name, so recipients see the address local-part instead of the mailbox owner's name. In Gmail the message shows up as "bikramc" rather than "Bikram C".

The header on a sent message looks like this:

```
From: user@example.com <user@example.com>
```

The cause appears to be two places. `server/src/routes/auth.ts` hardcodes `name: null` when creating a session, so `ctx.session.name` is always null even though the mailbox record has a name:

```ts
const created = await createSession({
  email: parsed.data.email,
  name: null,
  ...
});
```

Then `server/src/trpc/routes/mail.ts` passes an address only into the send driver:

```ts
return driverSend(ctx.smtp, ctx.imap, input.fromEmail || ctx.session.email, {
```

`send()` in `server/src/lib/imap-driver.ts` takes that as `from: string` and forwards it verbatim to nodemailer, so whatever arrives becomes the header. Notably the same `mail.ts` already carries `name: ctx.session.name` when it builds the alias list, so the field is expected to be populated there but never is.

Since nodemailer accepts the `"Name" <address>` form natively, populating the session name at login and formatting it at send time would be enough to fix this. There is also no sender-identity setting anywhere in the webmail UI, so there is currently no way to work around it from the application side.

Environment: OpenShip 0.3.0 self-hosted on Ubuntu 24.04 aarch64, webmail running as the OpenShip-managed container against a local iRedMail (Postfix/Dovecot) backend.

ValenFelizia:
### Summary
Fixes #218.

When sending email from the webmail app (`apps/email`), the `From:` header was generated with a bare email address (e.g. `user@example.com <user@example.com>`) instead of incorporating the mailbox owner's display name (`"Name" <user@example.com>`).

### Problem
1. `signInSchema` rejected optional `name` in sign-in payload.
2. `auth.ts` hardcoded `name: null` when calling `createSession`.
3. `mailRouter.send` passed bare `fromEmail` directly to `driverSend` without formatting the display name from `ctx.session.name`.

### Solution
- **`apps/email/server/src/lib/schemas.ts`**: Updated `signInSchema` to accept `name: z.string().optional()`.
- **`apps/email/server/src/routes/auth.ts`**: Persisted `parsed.data.name` when creating sessions or updating active sessions on sign-in.
- **`apps/email/server/src/trpc/routes/mail.ts`**: Exported `formatFromAddress` helper to format sender addresses as `"Display Name" <email>` and applied it in `mailRouter.send`.
- **`apps/email/server/test/from-header.test.ts`**: Added unit tests importing and verifying `signInSchema` and `formatFromAddress` behavior directly.

### Verification
- `bun run --cwd apps/email/server build` (`tsc --noEmit`) passed with **0 errors**.
- `bun test apps/email/server/test/from-header.test.ts` passed **6/6 tests**.

> **Note on Verification**: Verified via unit tests (`bun test`) and TypeScript build check. Live end-to-end SMTP/IMAP email delivery against a running mail container was not performed.

### DIFF
--- a/apps/email/server/src/lib/schemas.ts
+++ b/apps/email/server/src/lib/schemas.ts
@@ -49,6 +49,7 @@ export const signInSchema = z
   .object({
     email: z.email(),
     password: z.string().min(1),
+    name: z.string().optional(),
   })
   .strict();
 
--- a/apps/email/server/src/routes/auth.ts
+++ b/apps/email/server/src/routes/auth.ts
@@ -226,13 +226,16 @@ authRoutes.post('/sign-in', async (c) => {
   let liveIds: string[];
 
   if (existing) {
+    if (parsed.data.name !== undefined) {
+      await db.update(schema.session).set({ name: parsed.data.name }).where(eq(schema.session.id, existing.id));
+    }
     activeId = existing.id;
     activeExpiresAt = existing.expiresAt;
     liveIds = [existing.id, ...existingIds.filter((id) => id !== existing.id)];
   } else {
     const created = await createSession({
       email: parsed.data.email,
-      name: null,
+      name: parsed.data.name ?? null,
       password: parsed.data.password,
       imapHost,
       imapPort,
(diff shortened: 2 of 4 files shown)


Hydralerne:
good one

Issue #123, opened 2026-07-22
Global rate-limiter always applies default-anon, and double-charges routes that set their own policy


shuvamk:
Two related problems with the global limiter mounted at `app.use("/api/*", rateLimiter)` (`apps/api/src/app.ts:95`). Both are consequences of the same ordering issue, and both are reproducible.

I have a fix in mind but it's an architectural change to how routes are mounted, so I'd rather agree a direction with you first than send an unsolicited refactor. Happy to open the PR once you point me at the approach you prefer.

## 1. `default-authed` is never applied

`rateLimiter` picks its policy by probing for the request context:

```ts
let policy: PolicyId = "default-anon";
try {
  getRequestContext(c);
  policy = "default-authed";
} catch { /* no ctx — anon route */ }
```

But `authMiddleware` — which sets `c.var.ctx` — is injected *inside* each route's chain by `secureRouter.mount()`, so it runs **downstream** of the global middleware. At the time `rateLimiter` executes, `ctx` is never set, `getRequestContext` always throws, and the policy is always `default-anon`.

Repro (mirrors the real mount order — global limiter, then a route chain that sets `ctx`):

```ts
const app = new Hono();
app.use("*", (c, n) => { c.set("clientIp", "1.2.3.4"); return n(); });
app.use("/api/*", rateLimiter);
app.get("/api/projects",
  (c, n) => { c.set("ctx", { userId: "u1", organizationId: "o1" }); return n(); },
  (c) => c.json({ ok: true }));

await app.request("/api/projects");
```

```
X-RateLimit-Limit: 300      // default-anon, subject "ip"
                            // expected 3000 (default-authed, subject "user")
```

**Effect:** every authenticated user is limited to **300 req/min shared per source IP** instead of 3000/min per user. `default-authed`'s own comment anticipates the opposite ("a dashboard render fans out many concurrent reads … so the per-user ceiling is generous").

That bites hardest exactly where self-hosting is common: a team behind one office NAT or VPN shares a single 300/min bucket, and a self-hosted dashboard whose SSR calls egress from one host does too. Inbound GitHub webhooks (`webhook-ingress`, 120/min) additionally drain the same anon bucket from GitHub's delivery IPs.

## 2. Routes with their own policy are charged twice

The deferral is documented as:

> Routes that set `rateLimit: <policyId>` … inject their own per-route limiter … when present, the global limiter defers to it (the per-route limiter has already run upstream).

[...]

PR #232, opened 2026-07-25
feat(api): add a pre-auth per-IP flood guard on /api (#123)

shuvamk:
Standalone API requests currently reach session lookup before any rate limit. Restore a coarse per-IP ceiling ahead of authentication without charging the anonymous or authenticated route buckets twice (#123).

The existing `secureRouter` and raw Better Auth policies remain authoritative. Cloud mode and explicit `OPENSHIP_TRUST_EDGE=true` delegate the coarse ceiling upstream; login and per-user policies still apply. Health probes retain their exemption. Configuration lives in the current shared platform environment schema and is documented in both environment examples.

Adapted Shuvam Kumar’s original contribution to the current architecture, retaining its history. This PR targets the bug integration branch in #892; main will be reviewed separately.

Validation: 55 rate-limit and invitation/auth cases pass, including rejection before authentication, separate IP/route buckets, trusted-edge behavior and retained login limits. API TypeScript passes. Disabling the guard makes three behavioral regressions fail.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -12,6 +12,11 @@ OPENSHIP_CLOUD_TARGET=local-saas
 # Set to "true" to enable cloud billing, metering, multi-tenant features.
 # Default: false (self-hosted single-tenant mode).
 CLOUD_MODE=false
+# Set to "true" when this API sits behind an edge/proxy that already rate-limits
+# (the Openship Edge, or your own reverse proxy). Skips the app-level pre-auth
+# flood guard as redundant. Auto-implied under CLOUD_MODE. Leave false on a
+# standalone self-hosted API so the flood guard still protects the session lookup.
+OPENSHIP_TRUST_EDGE=false
 DEPLOY_MODE=desktop
 # REQUIRED whenever DEPLOY_MODE=desktop — env.ts refuses to boot without it, so
 # a copied .env fails on the first `bun dev` if this is missing. The launcher
--- a/apps/api/src/app.ts
+++ b/apps/api/src/app.ts
@@ -3,7 +3,7 @@ import { cors } from "hono/cors";
 import { logger } from "hono/logger";
 import { env, trustedOrigins } from "./config/env";
 import { handleApiError } from "./middleware/error-handler";
-import { rateLimiterFor } from "./middleware/rate-limiter";
+import { rateLimiterFor, floodGuard } from "./middleware/rate-limiter";
 import { clientIpMiddleware } from "./middleware/client-ip";
 import { betterAuthShield } from "./middleware/better-auth-shield";
 import { forceMcpConsent } from "./middleware/mcp-consent";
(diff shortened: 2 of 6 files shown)


Hydralerne:
perfect, gj thank you gonna check this

Hydralerne:
BTW, I was wondering about something. Since our Edge has native rate limiting with default limits built in, wouldn't adding Flood be an unnecessary extra step?

Since OpenShip apps are deployed behind the OpenShip Edge by default, the Edge already handles that. I think Flood would mainly be useful if someone deploys the app separately without using the Edge.

That said, it's still a nice feature to have for those cases.

Hydralerne:
@shuvamk would love to hear your feedback about my comment

shuvamk:
Good question, and you're right about the default topology — behind the Edge, the Edge's native limiting is the first line and this guard will basically never fire there. So for Edge-fronted deployments it's a cheap, never-fires backstop rather than a second gate competing with the Edge (generous at 6000/min/IP and a separate bucket, so it adds no real cost to that path).

The case it actually earns its keep is exactly the one you named — the app deployed **without** the Edge. And there's a specific reason it matters right now: the #123 fix that moved rate limiting into the route chain **removed** the old global `/api/*` limiter, which used to run before `authMiddleware`. So on a self-hosted API that isn't behind the Edge, there's currently *no* pre-auth per-IP throttle on the session lookup at all — this guard restores what that path lost, rather than adding something new.

Given the docs lean hard on "self-host on any VPS," I'd lean toward keeping it as defense-in-depth for those operators. But it's your call on the posture — a few ways to go:

1. **Keep as-is** — silent backstop; no-op behind the Edge, protection for standalone deploys.
2. **Make it explicit** — gate it behind a config flag (e.g. default-off, or auto-skip when `OPENSHIP_PUBLIC_URL`/`TRUST_PROXY` signals the Edge is in front) so it's opt-in/opt-out rather than always-on. Small change.
3. **Drop it** — if you'd rather treat the Edge as the sole rate-limit layer and just document that non-Edge deploys should put their own proxy limits in front.

No attachment to landing it if you feel the Edge covers enough — happy to do whichever fits your mental model. Just say which and I'll adjust or close. My mild preference is (1) or (2), only because the #123 change did quietly drop the pre-auth throttle for the non-Edge case.

Hydralerne:
good design, I think config based, if edge exist or trust edge flag is on, we can disable it, but good to have lets do it then merge, thank you so much for your help and support

shuvamk:
Done — went with the config-based gating you described. 👍

**What changed since your review:**

1. **Edge-trust gate.** Added an `OPENSHIP_TRUST_EDGE` env flag. `floodGuard` now steps aside entirely when the edge is trusted:

   ```ts
   if (env.CLOUD_MODE || env.OPENSHIP_TRUST_EDGE) { await next(); return; }
   ```

   - Auto-implied under `CLOUD_MODE` (behind the Oblien Edge, which already has default limits) — no config needed for the cloud path.
   - `OPENSHIP_TRUST_EDGE=true` covers the "my own reverse proxy already rate-limits" case.
   - Left **on** only for a standalone self-hosted API with nothing in front of it — which is the exact path the #123 change left without a pre-auth throttle.

2. **Documented the flag** in `apps/api/.env.example`, right next to `CLOUD_MODE`, so operators can find it.

3. **Test.** Added a case asserting the guard is a no-op under both `CLOUD_MODE` and `OPENSHIP_TRUST_EDGE` (sets no rate-limit headers, passes through). The existing invariants still hold: separate bucket from the per-route policies (no double-charge), `/api/health` exempt, per-IP isolation.

**On the red Test check:** that was stale — the branch had fallen behind `main`, and the failure was the `apps/dashboard` i18n-parity test, unrelated to this API-only PR (it's green on current `main`). I've rebased onto latest `main`, so CI should come back clean on the re-run. `bun run --cwd apps/api test` is green locally (114 files, 1166 passed).

Ready to merge whenever you are — thanks for the thoughtful review!

Issue #216, opened 2026-07-24
Help: Heavy tests needed for the stable release


Hydralerne:
Hello everyone! We're very close to stabilizing the architecture now. From this point on, no new features (might be some improvments but without chainging the arch) will be added until we've fully covered the current functionality with tests and validated all edge cases.

To make that happen, we need everyone to thoroughly test the existing features and try as many edge cases as possible. Your feedback and bug reports will help us ensure everything is solid before we move forward.

farrasrayhand:
I've submitted a PR (#193) that addresses several items from this cleanup effort: https://github.com/oblien/openship/pull/193

**Fixes:**
- Command injection fix in mail provisioning (`sq()` shell quoting)
- SSRF fix in webhook notification URLs (`assertPublicWebhookUrl()`)
- Compose `${VAR:?error}` operator per Docker Compose spec
- WebSocket/shell safe helpers (removed 23 inline try/catch blocks)
- `process.env` bypass fix (use validated config)

farrasrayhand:
Follow-up PR #224 is up with 32 edge-case tests: https://github.com/oblien/openship/pull/224

**Tests added:**

- **`compose-parser.test.ts`** — 5 tests for `:?` / `?` operators: unset throws, empty throws (for `:?`), empty allowed (for `?`), happy paths
- **`terminal-helpers.test.ts`** — 12 tests for 4 safe helpers (`safeWsSend`, `safeWsClose`, `safeShellWrite`, `safeShellClose`): normal operation, binary data, error swallowing (new file)
- **`webhook-url.test.ts`** — 15 tests for `assertPublicWebhookUrl()`: valid public URLs, HTTP rejection, all private/loopback ranges (new file)

Also extracted `assertPublicWebhookUrl` to `webhook-url.ts` (zero dependencies) so tests can import it cleanly without loading the full notification module.

Hydralerne:
Reviewed the linked test PRs against the current architecture and integrated #219, #243 and #248 into #892, preserving the contributors’ original commits and GitHub merge credit.

- #219: 33 added webmail cases for crypto, client IPs, rate limits, schemas and sanitizer attacks, using the existing Bun runner. All 110 mail-server tests pass.
- #243: 23 added environment-parser/subdomain cases using the current dashboard harness.
- #248: 78 stream-processing and ten Button cases using the existing Happy DOM setup. Current install-phase coverage is retained; the full dashboard suite passes 1,462 tests.

[...]

PR #243, opened 2026-07-26
test(dashboard): cover environment parsing and subdomain editing

walidozich:
Adds 23 dashboard cases for pasted environment files and subdomain normalization, including quoted values, inline comments, CRLF, empty values, and preserving the hyphen a user is typing. CONTRIBUTING.md documents the workspace test conventions and local component cleanup.

Validation: 68 tests pass across the new suites and existing project-status and log-entry suites. Uses the dashboard’s existing Vitest configuration.

Related to #216. Integrates into the bug-review branch in #892; the contributor’s original commits are preserved.


walidozich:
Thanks, all three were correct. Addressed in 07c7dc2.

**1. `CONTRIBUTING.md` wording was inaccurate.** Verified both counts: non-colocated tests do exist (`packages/adapters/test/`, `apps/cli/test/`), and `apps/email/server/test/from-header.test.ts` does import from `bun:test`. Reworded to say *most* workspaces use Vitest and *most* colocate, called out the `test/` directory convention and the Bun runner by name, and scoped the rest of the section explicitly to `apps/dashboard`.

**2. Typo fixed.** `normalizeSubdomandInput` to `normalizeSubdomainInput`.

**3. The "exhaustiveness guard" comment overclaimed.** You were right that a manual array silently drifts. Rather than soften the wording, I made the guard real by keying it as a `Record<ProjectStatus, true>`:

```ts
const STATUS_COVERAGE: Record<ProjectStatus, true> = { live: true, /* ... */ draft: true };
const allStatuses = Object.keys(STATUS_COVERAGE) as ProjectStatus[];
```

Verified it actually fails now by temporarily adding `"archived"` to the `ProjectStatus` union:

```
apps/dashboard/src/utils/project-status.test.ts(150,9): error TS2741:
  Property 'archived' is missing in type '{ live: true; ... draft: true; }'
  but required in type 'Record<ProjectStatus, true>'.
```

So adding a status without covering it now breaks `tsc --noEmit`. Union reverted afterwards; suite is back to 8 files / 116 tests passing.

Issue #151, opened 2026-07-23
Publish prebuilt Docker images for self-hosting


merkys7:
Thanks for Openship!

I'd like to self-host with Docker Compose, but the `api` and `dashboard` currently have to be **built from source** — I couldn't find official images on GHCR or Docker Hub.

Building the whole monorepo at deploy time is heavy and slow (a multi-minute `bun install` + Next.js build), needs build tooling on the host, and depends on the repo building cleanly at that moment. A pull-based deploy (`docker compose pull` + `up`) is what most self-hosters expect.

Would you consider publishing official images per release — e.g. `ghcr.io/oblien/openship-api` and `ghcr.io/oblien/openship-dashboard` (or a single combined image), tagged by version? That'd make self-hosting a plain `docker compose up` with no local build.

Thanks for considering!

Hydralerne:
will work on it today, thank you

Hydralerne:
Official images will be published today, along with option to install the entire system on fresh new machines from the cli, including required docker demon, etc

alexanderadam:
duplicate of #11

Frtrillo:
Fresh-server verification on Ubuntu 24.04 ARM64 found that the documented Compose path is still not consumable end to end:

- npm `openship@0.3.0` does not expose `--compose` and selected bare mode despite a working Docker installation.
- Anonymous manifest requests for `ghcr.io/oblien/openship-api:latest`, `openship-dashboard:latest`, and `openship-edge:latest` return HTTP 403. A new server therefore cannot pull the documented stack without extra registry setup.
- `.github/workflows/docker-images.yml` currently inspects while authenticated and appends `|| true`, hiding both inaccessible manifests and inspection failures.

I opened draft PR #250 to log out before inspecting each published manifest and to fail the workflow when anonymous access does not work. The GHCR package visibility may still need to be changed to public in repository/package settings before that check can pass.

Hydralerne:
added ty

alexanderadam:
@Hydralerne can you please point me to the URL of the pre-built container image?
I can't find it in the documentation yet and also [the compose file](https://github.com/oblien/openship/blob/f5a2d8c01be040a37b0799ee999b2265b9d14b5b/docker-compose.yml#L79-L129) is still pointing to a build context instead of a pre-built image.

Hydralerne:
It's already available in the packages repo, and the CLI is using it. You're right though - we should update the Compose setup to use the package instead of the build context.

[...]

Issue #102, opened 2026-07-22
Fresh Ubuntu 24.04 installation does not configure domain or HTTPS automatically


tilwegener:
## Description

When installing Openship on a fresh Ubuntu 24.04 server hosted at Hetzner, the installation completes, but the configured domain is not reachable and no HTTPS configuration is created.

The standard installation flow does not clearly configure or prompt for:

- The public domain
- Ports 80 and 443
- A reverse proxy
- A Let's Encrypt SSL certificate
- HTTP-to-HTTPS forwarding

As a result, Openship may be running locally, but the dashboard cannot be accessed through the intended domain.

## Environment

- Provider: Hetzner
- Operating system: Ubuntu 24.04
- Installation: Openship CLI
- Fresh server installation

## Steps to reproduce

1. Create a fresh Ubuntu 24.04 server at Hetzner.
2. Point a domain's DNS A record to the server IP.
3. Install Openship using the CLI.
4. Start Openship using the standard installation command.
5. Open the configured domain in a browser.

## Actual behavior

- The domain does not connect to the Openship instance.
- No HTTPS endpoint is available.
- No SSL certificate is configured.
- There is no generated reverse-proxy or TLS configuration.
- It is unclear which ports Openship expects to be exposed.

## Expected behavior

During installation, Openship should either:

1. Ask for the public domain and automatically configure HTTPS, or
2. Clearly explain that the following command is required:

```bash
openship
```

Hydralerne:
going check into this

Hydralerne:
Please update to 0.3.0 and try again let us know if its fixed

d3vi0us:
Same  problem here and I'm on 0.3.0..

tilwegener:
@Hydralerne Thanks for the quick response.

I tested again with **Openship v0.3.0** on a completely fresh **Ubuntu 24.04** server at Hetzner, but unfortunately the issue is still present.

For privacy, I've redacted my personal information from the installation log.

```text
==> Installing the Bun runtime…
...

✔ Openship installed.

root@server:~# openship

Your name
  [REDACTED]

Email
  [REDACTED]

How should this instance be reachable?
  Public (server / VPS)

How do you want a domain + HTTPS?
  Custom domain

Your domain
  app.example-domain.com

Public IP
  [REDACTED]

DNS
  Add a DNS A record:
    app.example-domain.com → [REDACTED]

HTTPS is issued automatically once DNS resolves (it retries for a couple minutes).

...

Deployed

[...]

Issue #86, opened 2026-07-22
[Bug]: openresty install fails on Ubuntu 26.04 — apt repo has no Release file for non-LTS codename (resolute)


tizerluo:
**Describe the bug**

`server install -c openresty` (and `certbot`, which runs `apt-get update` afterwards) fails on Ubuntu 26.04 LTS because the installer pins the openresty apt repo to `$(lsb_release -sc)`, i.e. `resolute`:

```
deb [signed-by=/usr/share/keyrings/openresty.gpg] http://openresty.org/package/ubuntu resolute main
```

openresty.org only publishes repos for LTS releases (currently up to `noble`; see http://openresty.org/package/ubuntu/dists/), so `apt-get update` fails:

```
E: The repository 'http://openresty.org/package/ubuntu resolute Release' does not have a Release file.
[openresty] OpenResty installation failed
```

Because the installer runs `apt-get update` again after rewriting the sources file, this also breaks subsequent component installs (e.g. certbot) in the same run.

**To Reproduce**

1. Register an Ubuntu 26.04 (resolute) server.
2. `openship server install <id> -c openresty certbot`
3. Install aborts at the apt step.

**Expected behavior**

Installer should map non-LTS/unsupported codenames to the nearest supported LTS (e.g. `resolute` → `noble`), or probe the repo and fall back with a warning.

**Workaround used**

Temporarily rewrite the codename before installing, restore afterwards:

```bash
cp /etc/os-release /etc/os-release.bak
sed -i "s/^VERSION_CODENAME=resolute/VERSION_CODENAME=noble/" /etc/os-release
openship server install <id> -c openresty   # succeeds
mv /etc/os-release.bak /etc/os-release
```

The noble-pinned repo works fine on 26.04 (openresty 1.31.1.1 installs and runs). Note: `chattr +i` on the sources file does NOT work — the installer aborts on the failed write (set -e).

**Environment**

- openship CLI 0.2.1 (npm global), also reproduced via desktop app API
- Server: Ubuntu 26.04 LTS x86_64 (Contabo VPS)

Hydralerne:
fixed, thank you

PR #250, opened 2026-07-27
Guard fresh Compose installs in releases

Frtrillo:
## Summary

Make the documented fresh-Linux Compose path an enforced release contract. A CLI release must expose and test the Compose install path, and its published container images must be pullable by an unauthenticated server.

## Reproduction

I attempted fresh Openship installations on AWS EC2 using Ubuntu 24.04 and Ubuntu 26.04. In both cases setup completed, but the dashboard was not reachable through the configured custom domain.

The detailed Ubuntu 24.04 ARM64 reproduction used Docker Engine 29.6.2, Docker Compose 5.3.1, and the current npm release (`openship@0.3.0`):

1. Install Docker and verify that the non-root `ubuntu` user can access the Docker daemon.
2. Install Openship with `curl -fsSL https://get.openship.io | sh`.
3. Run `openship up --public-url https://openship.example.com`.
4. Point the domain at the EC2 public IP and allow inbound TCP 80 and 443.
5. Open the custom domain.

## Expected result

On Linux with working Docker, `openship up` should select Compose mode and start Postgres, Redis, the API, dashboard, and OpenResty edge. The edge should bind ports 80 and 443 so the custom domain can serve the dashboard with TLS.

## Actual result with the published CLI

- `openship@0.3.0` has no documented `--compose` or `--bare` options.
- It installs bare mode as a user systemd service even though Docker is available.
- The API and dashboard listen only on localhost ports 3001 and 4000.
- Nothing listens on ports 80 or 443.
- Cloudflare returns error 521 because it cannot connect to the origin.
- Anonymous pulls of `ghcr.io/oblien/openship-api:latest`, `openship-dashboard:latest`, and `openship-edge:latest` return HTTP 403.

The same user-visible result occurred on Ubuntu 26.04: installation appeared successful, but no working HTTPS edge was available for the custom domain. Ubuntu-specific edge handling is discussed in #86, and the Ubuntu 24.04 domain failure is documented in #102.

## Why both checks are required

The current `main` branch already contains the newer selection logic that defaults to Compose on Linux and attempts to install Docker when necessary. That logic was added after the published `v0.3.0` package, but the default no-flag behavior did not have an explicit regression test and the npm publication job did not verify that the built package exposed Compose support.

Even with the newer CLI, a fresh installation still depends on public GHCR images. The Docker image workflow currently inspects manifests while authenticated and suppresses failures with `|| true`, so it can succeed while ordinary servers receive HTTP 403.

## Changes

- Add an end-to-end CLI regression test proving that `openship up` with no mode flag starts Compose when Docker is available.
- Run the CLI suite immediately before npm publication.
- Smoke-test the built CLI and require `openship up --help` to expose `--compose`.
- Log out of GHCR before inspecting published image manifests.
- Stop suppressing manifest inspection failures so private, missing, or inaccessible images fail the workflow.

This PR prevents the broken combination seen in `v0.3.0` from being published again. A new npm release containing the existing Compose-selection implementation is still required, and the GHCR packages must be public before the anonymous manifest checks can pass.

## Validation

- CLI test suite: 19 files and 126 tests passed
- CLI production build completed successfully
- Built CLI help contains `--compose` when executed with Bun
- Prettier checks passed for all changed files
- Workflow YAML parsed successfully
- `git diff --check` passed

Relates to #151 and #102.

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -14,7 +14,7 @@ name: Docker images
 on:
   push:
     tags:
-      - 'v*.*.*'
+      - "v*.*.*"
   workflow_dispatch:
     inputs:
       tag:
@@ -148,7 +148,11 @@ jobs:
             $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$METADATA_JSON") \
             $(printf "${SRC}@sha256:%s " *)
 
-      - name: Inspect
+      - name: Verify public manifest
         run: |
+          # Compose installs pull these images without registry credentials.
+          # Remove the workflow login so private or otherwise inaccessible
+          # packages fail the release instead of appearing to publish cleanly.
+          docker logout ghcr.io
           docker buildx imagetools inspect \
-            "ghcr.io/${{ github.repository_owner }}/openship-${{ matrix.image }}:${{ steps.meta.outputs.version }}" || true
+            "ghcr.io/${{ github.repository_owner }}/openship-${{ matrix.image }}:${{ steps.meta.outputs.version }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -507,7 +507,7 @@ jobs:
       - name: Verify Compose install support
         run: |
           bun run --cwd apps/cli test
-          bun apps/cli/dist/index.js up --help | grep -F -- "--compose"
+          node apps/cli/dist/index.js up --help | grep -F -- "--compose"
 
       - name: Publish to npm (OIDC trusted publishing — no token)
         working-directory: apps/cli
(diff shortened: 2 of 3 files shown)


Hydralerne:
good, only One change before I merge: please use node instead of bun for the built-CLI smoke test.
node apps/cli/dist/index.js up --help | grep -F -- "--compose"

npm users run this under Node, and Bun papers over the ESM createRequire "Dynamic require of events" crash that broke a real release for us - a bun-only check would have gone green on that build. Verified node works on the current bundle.

Frtrillo:
> good, only One change before I merge: please use node instead of bun for the built-CLI smoke test. node apps/cli/dist/index.js up --help | grep -F -- "--compose"
> 
> npm users run this under Node, and Bun papers over the ESM createRequire "Dynamic require of events" crash that broke a real release for us - a bun-only check would have gone green on that build. Verified node works on the current bundle.

Thanks for the clarification! Ive updated the smoke test to run the built CLI with Node instead of Bun and pushed the change

PR #268, opened 2026-07-28
fix(api): honor linked project root directory

swingo35:
## What changed

- honor a project's configured `rootDirectory` when resolving Compose drift from GitHub
- reject a missing configured root instead of silently reconciling an unrelated repository root
- clear stale local source paths when a local project is linked to GitHub, including sibling environments
- add coverage for selecting an explicit nested Compose root when the repository also has a root Compose file

## Why

An adopted local Compose project can be linked to a GitHub repository whose deployment manifest lives in a subdirectory. Previously, the link kept `localPath`, while drift reconciliation auto-detected the repository root and could import unrelated services. This makes the linked GitHub repository and configured root the actual deployment source.

## Validation

- `bun run --cwd apps/api test -- test/modules/deployments/prepare.service.test.ts`
- `bun run --cwd apps/api lint`
- `git diff --check`

### DIFF
--- a/apps/api/src/modules/deployments/build.service.ts
+++ b/apps/api/src/modules/deployments/build.service.ts
@@ -458,6 +458,7 @@ async function reconcileComposeDrift(
       repo: project.gitRepo,
       branch,
       ctx,
+      rootDirectory: project.rootDirectory || undefined,
     });
     const services = info.services ?? [];
     if (services.length === 0) return;
--- a/apps/api/src/modules/deployments/local-source.ts
+++ b/apps/api/src/modules/deployments/local-source.ts
@@ -71,7 +71,10 @@ function createLocalReader(dirPath: string): ProjectReader {
   };
 }
 
-export async function resolveFromLocal(dirPath: string): Promise<ProjectInfo> {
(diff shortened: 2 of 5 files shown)


Hydralerne:
are we fine to merge ?

Hydralerne:
Reviewed against current main and the shared engine. This proposal is superseded by the current Compose-path flow: `composePath` selects the manifest or its directory, while `rootDirectory` remains a build setting. The shared resolver honors a declared nested path, reads its adjacent `.env`, and rejects missing/unreadable/invalid paths instead of importing an unrelated root Compose file. Deployment reconciliation passes the saved `composePath` through that same resolver for Git and local sources.

The other fix here is also on main: linking Git clears `localPath` and obsolete source metadata across the project's environment group, under the shared runtime mutation lock, while preserving ordinary build settings and sibling branches.

All 144 existing preparation, deployment and source-transition tests pass. Closing the obsolete API-local implementation rather than introducing a second root selector beside `composePath`. For a nested manifest in the current UI/API, configure `composePath` (file or directory); that is the supported source-selection field.

PR #269, opened 2026-07-28
feat(servers): one-click "Add This Server" for the host Openship runs on

DiogoDuart3:
## Problem

The `isLocal` "This Server" row — the machine Openship itself runs on, as a deploy target — is only ever created by the boot reconcile in `lib/startup/self-server.ts`. That hook no-ops when the box has no founding admin yet:

```ts
const adminId = await foundingAdminId();
if (!adminId) return; // no admin/org yet — retry next boot
```

On a fresh VPS install the admin is created *after* that boot, so the row only appears if the API happens to restart later. An instance that never restarts never gets one, and `/servers` gives no way to add it: the only affordance is "Add Server", which asks for SSH credentials — the wrong transport, since deploys to the local row resolve to the host executor, not SSH. Typing the box's own IP into that form works today only because `createServer` has a loopback branch that quietly adopts/creates the local row, which is not discoverable.

## Change

**`POST /system/servers/local`** — idempotent registration of the current host, guarded exactly like the loopback branch of `createServer`:

- not cloud (`assertNotCloud`)
- `DEPLOY_MODE` `docker`/`bare` only (desktop has its own "This Machine" flow)
- host control enabled — with `--no-host-control` the box is deliberately not a deploy target
- box-owning org only, since running on this box is code execution on the control plane (host executor + mounted docker socket)

An existing local row is returned as-is (200) rather than duplicated; a new one is created with the same display host the boot reconcile uses (`SERVER_IP || HOST_DOMAIN || 127.0.0.1`) and audited as `server.added`.

**Servers list** — an "Add This Server" button beside "Add Server", and in the empty state with a hint line. Rendered only when `isServerHost` is true *and* no `isLocal` row exists, and hidden again once the row is there.

Strings translated for all nine locales; the i18n parity ratchet is unchanged (`servers` still at 113).

## Verification

Ran against a local dev instance (dashboard + API, fresh DB) and driven in Chromium:

- Empty state → button renders with the hint; click → row appears as "This Server / Current host / Online", success toast, button gone.
- With a remote server already listed → button shows in the header next to "Add Server", disappears after the add.
- API directly: `POST /system/servers/local` → 201 with `isLocal: true`; posting again → 200, same id, still one row.
- `tsc --noEmit` clean for `apps/api` and `apps/dashboard`; `vitest run` green (API 1083, dashboard 28).


Hydralerne:
Hello, this was a bug, now fixed, it's by default should auto detect the vps without add this server, and it happens now

PR #277, opened 2026-07-28
fix: cross-server migration to the local server drops the wrong volume names and ignores resume overrides

jbermudez00:
## Summary

Migrating a Coolify/Dokploy/Dokku server **into the OpenShip host itself** ("This Server" as target) hits two bugs in how the direct-transfer volume plan is built and resumed, unrelated to the SSH-bridge reachability issue fixed in #271. A third bug (a rare Docker-API-over-SSH hang with no error) is bounded with a timeout rather than left to hang the whole scan. Found and verified against a real production instance running this exact scenario.

**Note on scope**: this PR originally also included a `docker` CLI binary + `/var/lib/docker` bind-mount to work around `direct-transfer.ts`'s local-target commands failing inside the container. I've since learned `v0.4.8` shipped the *actual* fix for that: `createHostExecutor()`, a proper SSH-based "host channel" for containerized installs to reach the real host for host-OS operations — and its own code comment explicitly warns against the CLI+mount approach I originally used. `moveDataDirect`'s executor resolution already routes through `createHostExecutor()` for an `isLocal` target, so I've dropped that part of this PR; provisioning the host channel is an install-time concern (documented in `createHostExecutor`'s own comments), not something this PR should touch.

## What's fixed here

- **Volume transfer fails with `no such volume`** for services with `namespaceVolumes=true`. `moveDataDirect()` hardcoded `containerId: null` when building each `ServiceHandle` for `listSources()`, which treats a null id as "not deployed yet" and *guesses* the volume name from `service.volumes + namespaceVolumes` — correct for the backup/restore use case `listSources` was built for, but wrong for an adopted source service (e.g. from Coolify), which was never namespaced by OpenShip. Fixed by threading `scannedContainerIds` (already captured at discovery time) through so `listSources` inspects the live container's actual `Mounts` instead of guessing. Extracted as `resolveScannedContainerId`, unit tested.
- **Resuming a parked migration with a manual source-name override silently ignores the override** for volume items — the volume branch called `transferVolume(item.source, …)` (the stale pre-override name) instead of the computed override. Bind/path items already handled this correctly. Fixed by centralizing the per-item transfer decision into `planResumeTransfer`, a single function every call site must use the return value of — the "computed the right thing but used the wrong variable at the call site" class of bug isn't reachable through this shape. Unit tested.
- **Migration scans can hang indefinitely on "Listing containers, volumes and networks…"** past the point the daemon was already proven reachable. Empirically reproduced (no trigger identified): a single Docker API call over the SSH bridge occasionally hangs with no error even on an otherwise-healthy bridge. `discoverServerStack()` bounds the initial reachability probe but nothing after it. Bounded the rest of discovery as one unit so a stuck call fails clearly and retryably instead of hanging forever.

## Test plan

- [x] Unit tests added for `resolveScannedContainerId` and `planResumeTransfer` (`migration.orchestrator.test.ts`), following the existing pure-function testing pattern in this module (see `buildAdoptedServiceRows` in `migrate.service.test.ts`)
- [x] `bun run --cwd apps/api lint` (tsc --noEmit): clean
- [x] `bun run --cwd apps/api build`: clean
- [ ] **`vitest run` did not execute in my sandbox** — it fails on a `zod` SSR-transform error (`TypeError: undefined is not an object (evaluating '...z.object')`) that reproduces identically on an already-existing test file (`migrate.service.test.ts`) with nothing more than importing `zod`, so it looks like a pre-existing environment/tooling issue unrelated to this change, not something I could root-cause in the time I had. Flagging this openly rather than claiming a green run I didn't get — a maintainer with a working local setup should confirm these pass.
- [x] Reproduced each bug against a real Coolify-managed server, then confirmed the fix against the same server
- [x] Completed a full end-to-end cross-server migration into the local server (5/5 volumes, deploy, health check, cutover) on the production instance that originally reported this

https://claude.ai/code/session_01LnWthiYX6w6K5E8rAREpLa

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -49,18 +49,6 @@ RUN apt-get update \
  && apt-get install -y --no-install-recommends openssh-client rsync git ca-certificates \
  && rm -rf /var/lib/apt/lists/*
 
-# Docker CLI binary — cross-server migration (direct-transfer.ts) shells out
-# `docker save|load|tag|volume create` via CommandExecutor. That executor is a
-# LocalExecutor (plain child_process, no SSH) whenever a migration's source or
-# target server is the API host itself ("is_local" server / server-host mode),
-# so those commands run INSIDE this container. The Docker Engine API calls
-# elsewhere (dockerode over the socket bridge) don't need the CLI, but these
-# raw shell-outs do — without it every such transfer failed with
-# "docker: not found" (exit 127), even though /var/run/docker.sock is mounted.
-# Copy just the static client binary from the official CLI-only image instead
-# of apt-installing docker.io (which pulls in containerd/dockerd as well).
-COPY --from=docker:27-cli /usr/local/bin/docker /usr/local/bin/docker
-
 # @repo/* are TypeScript-SOURCE workspace packages (package.json main =
 # src/index.ts) resolved via the ROOT package.json "workspaces" field — there
 # is NO node_modules/@repo; bun points straight at packages/*. tsup externalizes
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -49,6 +49,18 @@ RUN apt-get update \
  && apt-get install -y --no-install-recommends openssh-client rsync git ca-certificates \
  && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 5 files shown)


Hydralerne:
fixed ?

PR #285, opened 2026-07-29
fix(edge): materialize vhost symlinks before container startup

### LINKED ISSUE #284: fix(edge): bare migration must materialize vhost symlinks before container startup
## Summary

On a Linux self-hosted instance running in bare mode, Edge migration can leave a host-side vhost symlink in the directory mounted into the OpenResty edge container. The symlink often targets `/etc/nginx/sites-available/...`, which is not mounted inside the container.

## Reproduction

1. Start with a conventional Nginx layout where `sites-enabled/app.conf` is a symlink to `../sites-available/app.conf`.
2. Run the OpenShip Edge migration.
3. The migration passes the link through to the Edge state directory and starts the container.

The container then exits because the included vhost is a dangling absolute link inside the container:

```text
[emerg] open() "/usr/local/openresty/nginx/conf/sites-enabled/openship.conf" failed (2: No such file or directory)
```

## Expected behavior

Before mounting the vhost directory, migration should materialize valid symlinks as regular files and remove dangling/unreadable symlinks. The resulting Edge config must contain only container-readable files.

## Additional context

Imported static vhosts may also refer to host paths that are not mounted into the Edge container. That is a separate adoption/mounting concern, but link materialization is required to prevent immediate container startup failure.

I have a focused regression test and patch ready for this behavior.

dongyu-will:
## What changed

- Materialize valid Edge vhost symlinks before the state directory is mounted into the OpenResty container.
- Remove dangling or unreadable links so Nginx does not include inaccessible paths.
- Add a regression test covering both valid and dangling symlinks.

## Root cause

Bare migration can preserve a conventional host-side `sites-enabled` symlink whose target lives outside the mounted Edge state tree. It becomes dangling in the container and prevents Nginx from starting.

## Validation

- `bun test packages/adapters/src/system/proxy/edge-guards.test.ts`
- 8 tests passed

Fixes #284.

### DIFF
--- a/packages/adapters/src/system/proxy/detect.ts
+++ b/packages/adapters/src/system/proxy/detect.ts
@@ -132,7 +132,21 @@ export async function sanitizeEdgeVhosts(
   // POSIX sh, one pass, no per-file round trips (this runs over SSH).
   const script = [
     `for f in ${sq(sitesDir)}/*.conf; do`,
-    `  [ -f "$f" ] || continue;`,
+    // A host-side vhost may be a symlink into /etc/nginx/sites-available. That
+    // target is deliberately not mounted in the edge container, so leaving the
+    // link here makes nginx's include glob fail at startup. Materialize valid
+    // links before the container mounts this directory; remove dangling ones.
+    `  [ -e "$f" ] || [ -L "$f" ] || continue;`,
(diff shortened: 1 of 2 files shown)


Hydralerne:
thank you, very helpfull

PR #290, opened 2026-07-29
Fix Cloud sign-in on private HTTP dashboards

M-DGayle:
## Summary

- preserve PKCE S256 generation when a self-hosted dashboard is served from a private HTTP origin where `crypto.subtle` is unavailable
- keep the auth popup open with an actionable callback error instead of silently closing it when setup fails
- add RFC 7636 and native-vs-fallback regression coverage

## Root cause

The browser flow opens an `about:blank` popup synchronously, then prepares the PKCE handoff asynchronously. On a private-LAN HTTP origin, Web Crypto exposes secure randomness but not `crypto.subtle`. `computePkceChallenge()` therefore rejects, and the existing catch handler immediately closes the popup. To the user, the OpenShip Cloud sign-in window appears and disappears with no explanation.

The fallback uses the already-resolved `@noble/hashes` package to calculate SHA-256 while retaining the existing random verifier and PKCE binding. Native Web Crypto remains the preferred path whenever it is available.

## User impact

- self-hosted users no longer get a flashing, unexplained popup
- PKCE is not disabled or downgraded
- HTTPS and desktop flows continue to use their existing behavior
- any remaining setup failure is rendered in the popup and logged instead of being hidden

## Validation

- `bun run --cwd apps/dashboard test` — 30 tests passed
- RFC 7636 S256 example passes through the fallback
- fallback output matches native Web Crypto output
- production dashboard build passed with the release environment:
  - `NODE_ENV=production`
  - `CLOUD_MODE=false`
  - `OPENSHIP_TARGET=local`
  - `NEXT_PUBLIC_API_PROXY=true`
- reproduced before/after against a self-hosted OpenShip 0.4.8 instance

Contributed by Madd Technologies.

### DIFF
--- a/apps/dashboard/package.json
+++ b/apps/dashboard/package.json
@@ -16,6 +16,7 @@
     "clean": "rm -rf .next .turbo node_modules"
   },
   "dependencies": {
+    "@noble/hashes": "^2.0.1",
     "@radix-ui/react-slot": "^1.2.4",
     "@repo/core": "workspace:*",
     "@repo/onboarding": "workspace:*",
--- a/apps/dashboard/src/app/cloud-connect-callback/page.tsx
+++ b/apps/dashboard/src/app/cloud-connect-callback/page.tsx
@@ -42,6 +42,14 @@ function CloudConnectCallbackInner() {
   const [errorMessage, setErrorMessage] = useState<string>("");
 
   useEffect(() => {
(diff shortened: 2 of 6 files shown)


Hydralerne:
Are we good to merge? I noticed it's still marked as a draft.

PR #297, opened 2026-07-29
fix: guard compose archive on success status

ahmedhesham6:
## Summary

- archive the prior compose release only when the new deployment is `ready` or `partial_failure`
- keep the prior live release intact after failed, cancelled, reconciling, or rolled-back-to-failed compose deploys
- add focused regression tests for all status paths

## Root cause

The compose pipeline returns normally after recording a failed deployment, and its caller always runs finalization. The old finalization guard excluded only `reconciling`, so failed and cancelled deployments could still reach the prior-release retirement path.

## Impact

A failed compose deploy can no longer retire the release that is still serving production traffic.

## Validation

- `cd apps/api && bun run lint`
- `cd apps/api && bunx vitest run src/modules/deployments` — 45 tests passed


Hydralerne:
perfect, thnak you

PR #300, opened 2026-07-29
fix(core): PEP 621 dependencies are dropped when `[project]` holds another array

shuvamk:
## Problem

`parsePyprojectToml` finds the PEP 621 dependency list with a single regex:

```ts
content.match(/\[project\][^[]*?dependencies\s*=\s*\[([\s\S]*?)\]/)
```

Neither half survives contact with a real `pyproject.toml`:

1. **`[^[]*?` cannot cross a `[`.** Any array-valued key declared before `dependencies` — `authors`, `classifiers`, `keywords`, `maintainers` — breaks the match outright, and **every dependency is lost**.
2. **`([\s\S]*?)\]` stops at the first `]`.** A requirement carrying extras (`"uvicorn[standard]"`) closes the array early, so entries after it are lost — and when the extras entry comes first, the list comes back empty.

Both fail silently: the deps map is just empty, framework detection finds nothing, and the project falls back to the generic `python` stack.

| `[project]` manifest | detected deps (before) | after |
|---|---|---|
| `authors = [{ name = "Ada" }]` then `dependencies = ["fastapi", "uvicorn"]` | `{}` | `fastapi, uvicorn` |
| `classifiers = ["Private :: Do Not Upload"]` then `dependencies = ["flask"]` | `{}` | `flask` |
| `dependencies = ["uvicorn[standard]", "fastapi"]` | `{}` | `uvicorn, fastapi` |
| `dependencies = ["fastapi", "uvicorn[standard]", "sqlalchemy"]` | `fastapi` | `fastapi, uvicorn, sqlalchemy` |

The user-visible end of it — a stock FastAPI manifest that happens to declare `authors`:

```toml
[project]
name = "svc"
requires-python = ">=3.12"
authors = [{ name = "Ada", email = "ada@example.com" }]
dependencies = ["fastapi", "uvicorn"]
```

| | stack | start command |
|---|---|---|
| before | `python` | `python app.py` |
| after | `fastapi` | `uvicorn main:app --host 0.0.0.0 --port 8000` |

`python app.py` is not a file this project has, so the deploy starts and dies.

The existing test passes only because its fixture has no array before `dependencies` and no extras:

```ts
'[project]\nname = "svc"\ndependencies = ["fastapi", "uvicorn"]\n'
```

## Fix

Two steps instead of one regex:

1. Slice the `[project]` table with `(?=\n\[|$)` — the same idiom `parsePipfile` and the Poetry branch in this file already use, and it keeps the "don't wander into the next section" property the old `[^[]` was reaching for.
2. Read the array literal with a quote-aware bracket scan (`sliceArrayValue`), so `[` / `]` inside a quoted value can't terminate it.

Poetry, Pipfile and the `optional-dependencies` branches are untouched.

## Tests

Four cases in `apps/api/test/lib/language-detectors.test.ts`:

- array-valued key (`authors`, `classifiers`) before `dependencies`
- extras in first and middle position
- `[project]`-table boundary — a later `[tool.uv] dependencies` must **not** be read, and a trailing section must not truncate
- end-to-end: the manifest above resolves to the `fastapi` recipe

Verified the first three **fail** with `packages/core/src/languages/python.ts` stashed (3 failed / 25 passed) and pass with it restored. The boundary case pins existing behaviour and passes both ways, deliberately.

`bun run test` → 7/7 turbo tasks successful, `apps/api` 1186 passed (1182 on `main` + the 4 new). `bun run --cwd apps/api lint` clean.

`packages/core/src/languages/python.ts` is prettier-clean before and after. `language-detectors.test.ts` has pre-existing prettier drift on `main`; I hand-formatted my own additions to prettier's shape and left the pre-existing lines alone, so the diff stays surgical.

## Alternative

If you'd rather not carry a hand-rolled scanner, the same two bugs can be fixed by keeping the regex and just widening it (`[\s\S]*?` for the table, `["']\s*\]` for the terminator) — smaller diff, but it re-opens the "wanders into the next section" hole and still mis-handles a bare `]` inside a value. Happy to switch if you prefer that trade.

### DIFF
--- a/apps/api/test/lib/language-detectors.test.ts
+++ b/apps/api/test/lib/language-detectors.test.ts
@@ -4,7 +4,14 @@ import { fileURLToPath } from "node:url";
 import { join } from "node:path";
 
 import { detectStack, type RepoFile } from "../../src/lib/stack-detector";
-import { STACKS, LANGUAGES, getRuntimeImage, type StackId, type StackDefinition } from "@repo/core";
+import {
+  STACKS,
+  LANGUAGES,
+  collectDependencies,
+  getRuntimeImage,
+  type StackId,
+  type StackDefinition,
+} from "@repo/core";
 
 /**
  * Per-stack recipe validation. Loads the real fixtures in `fixtures/deploy/`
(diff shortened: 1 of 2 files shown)


Hydralerne:
your focus on stacks is amazing, thank you

PR #305, opened 2026-07-29
feat(email): alias, forwarding and catch-all management for mail server

DiogoDuart3:
## Summary

The `/emails` admin panel manages domains and mailboxes, but the underlying
iRedMail-style mail stack already supports per-address forwarding and
domain catch-alls via the `vmail.forwardings` table — there was just no way
to manage them without SSHing in and writing SQL by hand. This adds an
**Aliases** tab (after Mailboxes) so admins can create/toggle/delete
forwards and catch-alls from the dashboard.

## Implementation notes

- **`apps/api/src/modules/mail/admin/aliases.service.ts`** (new) —
  `listAliases`/`createAlias`/`updateAliasActive`/`deleteAlias`, mirroring
  `mailboxes.service.ts` exactly: same `psql-runner.ts` (SSH+psql)
  execution path, same `q()`/`qInt()` quoting for injection safety, same
  validate-then-INSERT convention.
- A `forwardings` row with `is_alias = 1` is either a **per-address alias**
  (`address = local@domain`) or a **catch-all** (`address` = the bare
  domain, matching `catchall_maps.cf`'s `WHERE forwardings.address =
  domain.domain` join) — no separate flag needed, the shape of `address`
  disambiguates it.
- Guards:
  - destination / local-part email format validation
  - reject an alias address that's already a real mailbox — `409` (since
    `virtual_alias_maps` is checked before `virtual_mailbox_maps`, an alias
    over a mailbox address would silently redirect that mailbox's mail)
  - reject an exact duplicate `(address, forwarding)` pair — `409`
  - catch-all creation is a delete-existing-then-insert transaction, so
    there's only ever one per domain
- 4 REST endpoints under `/admin/:serverId/aliases[/:id]`
  (`GET`/`POST`/`PATCH`/`DELETE`), same permission-tag convention as the
  existing mailbox routes (list/write/write/admin).
- Frontend: `aliases-tab.tsx` mirrors `mailboxes-tab.tsx`'s
  `DataTable`/`FormModalContent` pattern. The create dialog has its own
  domain select (unlike Mailboxes) so an alias for any domain can be added
  without leaving the current view, plus a catch-all checkbox that swaps
  the local-part field for a domain-wide note.
- `vmail` SQL (`forwardings`/`mailbox`/`domain` tables) is the only state
  touched — no control-plane schema changes, no migration.
- i18n: English strings only for the new `aliases.*` / `panel.tabs.aliases`
  keys; `i18n-parity.test.ts`'s ratchet baseline bumped to match the exact
  new ​count (consistent with how every other namespace in that file
  already carries an English-first backlog).

## Test plan

Verified live on a self-hosted install running a fork of `0.4.5` (api
rebuilt from source with this change, dashboard likewise) against a real
iRedMail mail server:

- [x] Created an alias via the dashboard UI and via the raw API
      (`local@domain` → external address)
- [x] Confirmed the row lands correctly in `vmail.forwardings`
      (`is_alias=1`, `is_forwarding=0`, `dest_domain` derived from the
      destination)
- [x] Confirmed it's **live with no restart**:
      `postmap -q <alias> pgsql:/etc/postfix/pgsql/virtual_alias_maps.cf`
      resolves to the destination immediately after creation
- [x] Guard: alias address colliding with an existing mailbox → `409`,
      mailbox untouched
- [x] Guard: exact duplicate `(address, forwarding)` → `409`
- [x] Guard: invalid destination email format → rejected
- [x] Toggle active on/off via `PATCH`
- [x] Delete via UI and via `DELETE` — confirmed row removed from
      `vmail.forwardings` and gone from the list endpoint
- [x] `tsc --noEmit` clean on `apps/api` and `apps/dashboard`
- [x] `i18n-parity.test.ts` passes with the updated baseline
- [x] All other hosted sites / unrelated services on the box unaffected
      throughout

### DIFF
--- a/apps/api/src/modules/mail/admin/admin.controller.ts
+++ b/apps/api/src/modules/mail/admin/admin.controller.ts
@@ -37,6 +37,15 @@ import {
   softDeleteMailbox,
   updateMailbox,
 } from "./mailboxes.service";
+import {
+  createAlias,
+  deleteAlias,
+  listAliases,
+  updateAliasActive,
+  AliasConflictsWithMailboxError,
+  AliasExistsError,
+  AliasNotFoundError,
+} from "./aliases.service";
 import { getMailServerStats } from "./stats.service";
 import { scanDns } from "./dns-scan.service";
 import { sendTestEmail, TestEmailError } from "./test-email.service";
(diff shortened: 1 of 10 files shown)


DiogoDuart3:
The single failing check (`Test`) is a pre-existing failure on `main`, unrelated to this PR's files.

It's `apps/cli`'s `test/e2e/up.test.ts > openship up --compose (edge chain) > imports migrated sites into the edge after the stack is healthy`, failing with `TypeError: exec.exec is not a function` in the test harness around `edge-import.ts`'s `waitForEdgeRunning`. This fails identically on `main`'s own tip, commit `1ea849a1` (the exact commit this branch is based on) — see run [30474891407](https://github.com/oblien/openship/actions/runs/30474891407). None of this PR's files touch `apps/cli`.

For what it's worth on the rest of the suite: `apps/dashboard` is fully green (8/8 test files), `apps/api` showed zero failures in everything that ran before turbo's fail-fast aborted the run on the `cli` failure, and `Typecheck` passed repo-wide.

PR #319, opened 2026-07-30
fix(mail): detect SpamAssassin via amavis when standalone unit is absent

### LINKED ISSUE #240: [Bug] Mail health check: three false positives (SpamAssassin daemon, desktop-resolver DNS scan, DKIM baseline)
## Summary

The mail server's **健康状态 / Health** tab flags four problems, but after verifying each one on the server, three of them are false positives in the health check itself (the fourth, missing PTR, is a genuine and correctly-reported gap). Details below with evidence.

## False positive 1 — "SpamAssassin not installed"

The check looks for a standalone `spamassassin` systemd unit, which indeed doesn't exist — but that's the standard iRedMail architecture: SpamAssassin runs **inside amavisd-new**, not as a separate daemon. Spam scoring is verifiably working; `amavis` log lines on the same server show SA scores and test lists for every message, e.g.:

```
Passed CLEAN {RelayedInbound}, ... Hits: 1.073, ... Tests: [DKIM_SIGNED=0.1, DKIM_VALID=-0.1, ...]
```

Expected: the check should probe SA-through-Amavis (e.g. presence of `@bypass_spam_checks_maps` config / SA version via amavisd, or a scored test message) instead of looking for `spamassassin.service`.

## False positive 2 — A / AAAA "resolves to the wrong address"

The DNS scan runs on the **desktop** through the local system resolver. On machines behind a fake-IP proxy (Clash / sing-box — common on macOS), the resolver returns synthetic addresses, which is exactly what the UI showed:

- A: expected `96.9.210.140`, got `198.18.1.202` (198.18.0.0/15 = fake-IP pool)
- AAAA: expected `2402:a7c0:8100:a017::8db:a`, got `fdfe:dcba:9876::11a` (ULA fake range)

Querying the authoritative nameservers via DNS-over-HTTPS shows both records are exactly the expected values — the zone is correct, the measurement is polluted.

Expected: run the DNS scan **server-side**, or resolve via DoH against the zone's authoritative NS, rather than trusting the desktop's resolver.

## False positive 3 — "DKIM TXT doesn't match the key generated at install"

We extracted the public key from the server's actual DKIM private key (`/var/lib/dkim/<domain>.pem`):

```
openssl rsa -in /var/lib/dkim/<domain>.pem -pubout -outform DER | openssl base64 -A
```

and compared it byte-for-byte with the TXT record published at `dkim._domainkey.<domain>`: **identical, all 392 base64 characters**. DKIM also verifies in real mail flow (receivers report `DKIM_VALID` / `dkim:AUTHOR`). So the install-time baseline the check compares against is wrong, or the comparison doesn't normalize whitespace/line-splitting in multi-string TXT records.

## The one genuine finding

**PTR missing** — correctly reported. 96.9.210.140 has no rDNS. Impact is low here because all outbound mail relays through Amazon SES, but the report itself is accurate and useful.

## Additional observations (not health-check items, hit during the same deployment)

- **Installer prompts have no CLI response channel.** When the installer hits an interactive gate (e.g. `edge_conflict` when ports 80/443 are taken), the CLI can only wait for the 5-minute timeout and fail. The response endpoint `POST /api/system/install/respond` exists and works (we answered the prompt via curl), but it isn't surfaced in the CLI or docs.
- **iRedMail disables UFW during install** and the wizard doesn't re-enable it afterwards — the server is left without a firewall until the operator notices.
- **`POST /api/system/check` is undocumented** — we located it via strings in the `openship-api` binary after the intuitive routes (`/api/system/servers/:id/check`, `/api/system/servers/check`) returned 404.

## Environment

- OpenShip desktop v0.3.0 on macOS (local resolver behind a fake-IP proxy)
- Mail server: Ubuntu 26.04 LTS, iRedMail 1.8.1 deployed by the OpenShip Email app, Amavis 2.13.0

Happy to re-run any check against our live server if useful.

santhiprakash:
## Summary

Fix a false "SpamAssassin not installed" health-check result on iRedMail deployments where SpamAssassin runs inside amavisd-new instead of as a standalone systemd unit.

## Motivation

On standard iRedMail installs (Debian/Ubuntu), the installer disables `spamassassin.service` because scoring runs inside amavis. The mail health probe only checked `systemctl show spamassassin`, so healthy servers reported SpamAssassin as missing. Reported in #240.

## Related issue

Closes #240 (first of three reported false positives — scoped to the SpamAssassin/amavis probe only).

## Changes

- **apps/api**
  - `mail-health.service.ts`: add `probeSpamAssassin()` fallback when the standalone unit is missing/inactive but `spamassassin` is on PATH and `amavis` is active; report `subState: "integrated"`.
  - `test/modules/mail/mail-health.service.test.ts`: regression tests for integrated, standalone-active, and truly-missing cases.

## Verification

```bash
bun run --cwd apps/api test test/modules/mail/mail-health.service.test.ts
# ✓ 3 tests passed

bun run --cwd apps/api lint
# tsc --noEmit (pass)
```

Preflight against `upstream/main` confirmed the bug marker (`unit: "spamassassin"`) is still present and no open/merged duplicate PR exists for this probe change.

## Screenshots

N/A — API health-probe logic only.

## Checklist

### DIFF
--- a/apps/api/src/modules/mail/mail-health.service.ts
+++ b/apps/api/src/modules/mail/mail-health.service.ts
@@ -118,16 +118,57 @@ export interface MailComponentHealth {
  * with `systemctl show <unit1> <unit2> …` but the result parsing gets
  * fiddly - keep it simple and parallel-friendly via Promise.all.
  */
-export async function checkMailHealth(
-  exec: CommandExecutor,
-): Promise<MailComponentHealth[]> {
-  const results = await Promise.all(
-    MAIL_COMPONENTS.map(async (comp) => probeUnit(exec, comp)),
-  );
+export async function checkMailHealth(exec: CommandExecutor): Promise<MailComponentHealth[]> {
+  const results = await Promise.all(MAIL_COMPONENTS.map(async (comp) => probeUnit(exec, comp)));
   return results;
 }
 
 async function probeUnit(
+  exec: CommandExecutor,
+  comp: MailComponentDef,
+): Promise<MailComponentHealth> {
+  if (comp.key === "spamassassin") {
+    return probeSpamAssassin(exec, comp);
+  }
+  return probeSystemdUnit(exec, comp);
+}
+
+/**
+ * iRedMail installs SpamAssassin but disables the standalone systemd unit —
+ * scoring runs inside amavisd-new. When the standalone unit is missing or
+ * inactive, fall back to "binary present + amavis active".
+ */
+async function probeSpamAssassin(
+  exec: CommandExecutor,
+  comp: MailComponentDef,
+): Promise<MailComponentHealth> {
+  const standalone = await probeSystemdUnit(exec, comp);
+  if (standalone.status === "active") {
+    return standalone;
+  }
+
+  try {
+    const raw = await exec.exec(
+      `command -v spamassassin >/dev/null 2>&1 && systemctl is-active amavis 2>/dev/null || echo inactive`,
+    );
+    if (raw.trim() === "active") {
+      return {
+        key: comp.key,
+        label: comp.label,
+        description: comp.description,
+        unit: comp.unit,
+        status: "active",
+        subState: "integrated",
+      };
+    }
+  } catch {
+    // fall through to standalone result
+  }
+
+  return standalone;
+}
+
+async function probeSystemdUnit(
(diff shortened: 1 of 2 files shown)


Hydralerne:
fixed already ty

PR #331, opened 2026-07-30
fix(i18n): translate the 90 untranslated deploy-namespace keys, restore the parity ratchet

AbdullahM07:
## Summary

Translates the 90 untranslated keys in the `deploy` namespace across all 8
locales and removes its `MISSING_BASELINE` entry, fixing the `i18n-parity` test
that currently fails on `main`.

## Motivation

`i18n-parity.test.ts` fails on `main` right now:

    deploy: 90 missing (baseline 18)

`MISSING_BASELINE` is a one-way ratchet — it never asks anyone to clear the
existing backlog, only to not add to it. That contract was broken: the
DNS-records modal landed 9 new English `deploy.dns.*` keys with no translations,
pushing the count from 18 to 90.

Nothing is functionally broken. `deepMerge` in `src/i18n/index.ts` falls back to
English, so no key renders blank. The costs are a red test that every unrelated
PR now inherits, and non-English users reading English mid-sentence in an
otherwise translated UI — #315 is a live example of the latter, a Chinese user
reporting raw English strings in the dashboard.

## Related issue

None — bug fix restoring a test that fails on `main`.

## Changes

Two commits, each green on its own.

**`fix(i18n)` — apps/dashboard/src/i18n/locales/{ar,de,es,fr,ja,pt,tr,zh}/deploy.json**

- `dns.*` — 9 keys × 8 locales (72), the DNS-records modal
- `serveMode.mode{Static,Process}` — 2 × 5 locales (10): de, es, ja, pt, zh
- `domainSettings.route{Free,Custom,None}{Label,Desc}` — 6, tr only
- `targetStep.thisServer{Badge,Host}` — 2, tr only

**`test(i18n)` — apps/dashboard/src/i18n/i18n-parity.test.ts**

- Deleted `deploy: 18` from `MISSING_BASELINE`, now that the namespace is at 0.

Translations reuse each locale's existing vocabulary rather than translating from
scratch, so the new strings read consistently beside their neighbours:
`despliegue` (es), `implantação` (pt), `alan adı` plus the file's own
untranslated `Host` (tr), `デプロイ` (ja), and 你 rather than 您 inside the `dns`
block (zh) to match the surrounding keys. The `{domain}` placeholder in
`domainSettings.routeFreeDesc` is preserved.

Keys are inserted in the English key order, so every locale diff is additive —
the only removed lines are 9 pre-existing lines that gained a trailing comma.

## Verification

```bash
# Before — on main
$ bun run --cwd apps/dashboard test
 ❯ src/i18n/i18n-parity.test.ts (2 tests | 1 failed)
   × introduces no NEW missing keys beyond the per-namespace baseline
AssertionError: New i18n drift vs English.
deploy: 90 missing (baseline 18)
 Test Files  1 failed | 7 passed (8)
      Tests  1 failed | 59 passed (60)

# After
$ bun run --cwd apps/dashboard test
 ✓ src/i18n/i18n-parity.test.ts (2 tests) 3ms
 Test Files  8 passed (8)
      Tests  60 passed (60)

# `deploy` is gone from the drift report; the total drops by exactly 90
$ bun run --cwd apps/dashboard i18n:check
i18n drift vs "en" — missing: 5571, extra: 14   # was 5661
# (no `deploy` line in the per-namespace breakdown)

# Interpolation placeholders match English in every locale; all 8 files parse
✓ all placeholders identical
✓ ar zh de es fr ja pt tr

# Diff is additive and scoped to i18n
$ git diff --numstat upstream/main..HEAD
+99 -10
```

I also checked the first commit in isolation: with the 72 `dns.*` keys translated
and the baseline still at 18, the count lands exactly on 18 and the test passes —
so that commit alone is enough to unblock CI if you'd rather take only part of
this.

## Checklist

Two notes on that unchecked box — both reproduce on a clean `main` and are
untouched here:

- `bun run --cwd apps/dashboard lint` fails with `Invalid project directory
  provided, no such directory: apps/dashboard/lint`, from the `next lint`
  invocation itself.
- CONTRIBUTING asks for `bun format` before committing, but on a clean checkout it
  rewrites roughly 51k insertions / 58k deletions across the repo — the tree isn't
  Prettier-clean, so running it would bury this diff. I deliberately skipped it;
  the JSON here matches the surrounding 2-space style and parses clean. Happy to
  run it scoped to just these files if you'd prefer.

## Follow-ups I found but deliberately left out

- The ratchet counts key *presence*, not whether the value differs from English.
  The whole `serveMode` block (`heading`, `subtitle`, `staticLabel`, `serverDesc`,
  `startCommandHint`, …) is English in all 8 locales and the test is satisfied;
  `ar`/`fr`/`tr` literally carry `"modeStatic": "Static"`. Worth its own issue.
- `settings.methodToken` is missing from `zh`, which is the confirmed cause of one
  symptom in #315. Different namespace, doesn't affect this test — a natural next
  PR.

### DIFF
--- a/apps/dashboard/src/i18n/i18n-parity.test.ts
+++ b/apps/dashboard/src/i18n/i18n-parity.test.ts
@@ -33,7 +33,6 @@ const MISSING_BASELINE: Record<string, number> = {
   emails: 42,
   projectDetail: 42,
   brand: 40,
-  deploy: 18,
   library: 119,
   billing: 6,
 
--- a/apps/dashboard/src/i18n/locales/ar/deploy.json
+++ b/apps/dashboard/src/i18n/locales/ar/deploy.json
@@ -40,7 +40,16 @@
     "verifySuffix": " للتحقق من الملكية.",
     "cloudInfoPre": "أضف ",
     "recordCname": "سجل CNAME",
-    "cloudInfoMid": " للتوجيه، ثم أضف "
+    "cloudInfoMid": " للتوجيه، ثم أضف ",
+    "modalTitle": "وجّه نطاقك، ثم انشر",
+    "modalSubtitle": "يُصدر HTTPS عند أول عملية نشر بعد أن يشير نطاقك إلى هنا — أو تجاوز الخطوة وأضف DNS لاحقًا من تبويب النطاقات.",
+    "descCnameWww": "يوجّه www إلى نطاقك الرئيسي",
+    "autoConfigTitle": "الإعداد التلقائي عبر مزوّد DNS الخاص بي",
+    "autoConfigDesc": "اربط مُسجّل النطاق وسنضيف هذه السجلات نيابةً عنك.",
+    "comingSoon": "قريبًا",
+    "deployAction": "نشر",
+    "cancel": "إلغاء",
+    "loadingRecords": "جارٍ تحميل سجلات DNS…"
   },
   "buildSummary": {
     "title": "ملخص النشر",
--- a/apps/dashboard/src/i18n/locales/de/deploy.json
+++ b/apps/dashboard/src/i18n/locales/de/deploy.json
@@ -40,7 +40,16 @@
     "verifySuffix": " zur Besitzbestätigung hinzu.",
     "cloudInfoPre": "Fügen Sie den ",
     "recordCname": "CNAME-Eintrag",
-    "cloudInfoMid": " zum Routing hinzu und fügen Sie dann den "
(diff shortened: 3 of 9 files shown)


AbdullahM07:
Related: the `@repo/api` failure on this PR's CI is a separate pre-existing break on `main`, not from this diff — fixed in #337. `main` has two independent red tests right now: this PR takes the `@repo/dashboard` i18n one, #337 takes the `@repo/api` one. Together they turn CI green. Independent workspaces, so merge order does not matter.

PR #359, opened 2026-07-31
fix(dashboard): mobile responsive fixes across sidebar, home, project, emails, deployments

DiogoDuart3:
A batch of small, independent mobile-responsive fixes across the dashboard, verified live via a headless screenshot pass at 390px and 1280px before/after each change.

- Sidebar: collapses to a drawer below `lg` (was always fixed-position, unusable on narrow viewports) - adds a mobile top bar with an open/close trigger.
- Home rows: narrow the name column and widen the hosting label on mobile; stop the hosting badge from hard-clipping mid-word; go icon-only for the hosting badge and fix a stack-badge clipping issue; hide the Stack badge below `sm` - it never actually fit at 390px.
- Deployments: stop the status badge and commit hash overlapping on a deployment row, make the filter chip row scrollable instead of wrapping/overflowing, tighten the commit hash display on mobile.
- Project detail: tab strip gets a fade instead of hard-clipping, stop label/value crowding in the overview tab, stop the page heading truncating on mobile.
- Emails admin: stop the header buttons wrapping awkwardly, fade the admin tab strip (same treatment as project detail), stop the mail hostname/CTA overlapping on mobile.

## Note on CI

Recent merged PRs against `main` (#322, #327) show `Typecheck: SUCCESS` / `Test: FAILURE` on their PR-context runs. Flagging in case it reproduces here too - doesn't look related to this change.


DiogoDuart3:
Rebased onto current `main`. Three conflicts, all in files `main` had since drifted:

- `sidebar.tsx` — nav sections moved to `lib/sidebar-nav` and the brand label switched to `useBrandName()`. Kept both, and kept the label's `lg:hidden` rather than `main`'s unconditional hide: `collapsed` is a desktop-only state, so hiding it outright blanks the mobile drawer's header.
- `DeploymentCard.tsx` — `main` added a dynamic z-index for the open dropdown; merged with the mobile layout change.
- `emails/page.tsx` — the page is now `<MailConsole />`; the header wrap fix moved with it into `_components/mail-console.tsx`.

Dashboard typecheck and 713 tests green.

DiogoDuart3:
Follow-up: 359 and 360 both touched the same line in `_components/mail-console.tsx` (this PR adds `truncate`, 360 drops the `/70`), so whichever merged second would have gone conflicted. Both branches now write that line identically, so either merge order stays clean.

PR #360, opened 2026-07-31
fix(dashboard): a11y contrast/ARIA fixes, dismissal persistence, servers skeleton

DiogoDuart3:
A batch of small, independent a11y and UX fixes to the dashboard, verified live via a headless screenshot + DOM-role pass at 390px before/after each change.

## Accessibility (Lighthouse-driven)
- Raised `--th-text-secondary`/`--th-text-muted` alpha (light/dark themes) to clear 4.5:1 against page/card backgrounds; the dim theme already passed.
- Eliminated `text-muted-foreground/70` double-dimming repo-wide in favor of the (now-passing) base token.
- Deepened `--st-warning-fg` on the light theme (amber-600 -> amber-800) so `text-warning` clears 4.5:1 on its tinted banner background; dark and dim already passed.
- The shared admin `DataTable` (aliases/mailboxes/domains lists) had `role="row"`/`role="cell"`/`role="columnheader"` on the data columns but the actions-column cells had no role at all - fails `aria-required-children`. Wrapped in `role="table"`/`role="rowgroup"` and added the missing roles on the actions column.
- Added `aria-label` to the unlabeled domain-picker `<select>` duplicated across the aliases, mailboxes, and dns admin tabs.
- Sidebar section labels (`MAIN`/`SETTINGS`/`INFRASTRUCTURE`/`Account`) used a `/60` double-dim on top of the already-tuned base token - same class of bug as the `/70` sweep above, different fraction.

## UX polish
- Persisted critical update-banner dismissal (was re-showing every session).
- Scoped the mail welcome-modal's dismissal to server+domain (was dismissing globally, so switching mail servers never showed it again even though the new server had never seen it).
- Swapped the servers list's loading spinner for a row skeleton.

## Note on CI

Recent merged PRs against `main` (#322, #327) show `Typecheck: SUCCESS` / `Test: FAILURE` on their PR-context runs. Flagging in case it reproduces here too - doesn't look related to this change.

### DIFF
--- a/apps/dashboard/src/app/(auth)/forgot-password/page.tsx
+++ b/apps/dashboard/src/app/(auth)/forgot-password/page.tsx
@@ -51,7 +51,7 @@ export default function ForgotPasswordPage() {
           <p className="mt-2 text-sm text-muted-foreground">
             {interpolate(t.auth.forgotPassword.sentDescription, { email })}
           </p>
-          <p className="mt-2 text-xs text-muted-foreground/70">
+          <p className="mt-2 text-xs text-muted-foreground">
             {t.auth.forgotPassword.spamHint}
           </p>
           <Button
--- a/apps/dashboard/src/app/(auth)/verify-email/page.tsx
+++ b/apps/dashboard/src/app/(auth)/verify-email/page.tsx
@@ -257,7 +257,7 @@ function VerifyEmailContent() {
           </Button>
         )}
 
-        <p className="mt-4 text-xs text-muted-foreground/70">
+        <p className="mt-4 text-xs text-muted-foreground">
           {t.auth.verifyEmail.spamHint}
         </p>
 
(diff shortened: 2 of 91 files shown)


Hydralerne:
amazing

DiogoDuart3:
Rebased onto current `main`. Ten conflicts, all in the contrast sweep — same `text-muted-foreground/70` → `text-muted-foreground` change, re-applied against the drifted files. Two of them (`ConnectionCard`, `ServiceDetailPanel`) `main` had already fixed, so those hunks dropped out. Decorative `KeyRound` icons in `GitHubConnection` were left at `/70` — not a text-contrast defect and not in the original PR.

The relocated `emails/page.tsx` subtitle now lives in `_components/mail-console.tsx` and was swept there.

One real fix the rebase surfaced: `main` narrowed `DesktopConfigKey` to a union and added the `RENDERER_CONFIG_KEYS` bridge allowlist since this branch was written, so `dismissedCriticalAdvisoryKeys` had to be registered in both plus `AppConfig` — otherwise the desktop bridge refuses the key and the dismissal silently never persists (the call sites swallow the rejection).

Full monorepo typecheck and test suite green.

DiogoDuart3:
Follow-up, two notes:

- 359 and 360 both touched the same line in `_components/mail-console.tsx` (359 adds `truncate`, this one drops the `/70`), so whichever merged second would have gone conflicted. Both branches now write that line identically, so either merge order stays clean.
- In `CleanDeployProgress.tsx` the sweep covers three `liveHost` paragraphs and an uppercase label where the original commit had only two lines — `main` has since duplicated that element across the success/failure/idle panels. Same defect, same fix; flagging it since it's more lines than the original diff.

PR #369, opened 2026-07-31
fix(api): normalize loopback MCP redirect URIs

soares-daniel:
## Summary

Fix MCP OAuth connections that complete browser consent but fail during token exchange when the client alternates between `localhost` and `127.0.0.1`.

Before Better Auth handles MCP authorize and token requests, Openship now maps an equivalent loopback callback back to the URI registered for that client. Every URL component except the hostname must remain identical.

## Motivation

I noticed this while connecting OpenCode to a self-hosted Openship MCP server. Browser authorization and consent completed successfully, but OpenCode remained disconnected.

Tracing the OAuth flow showed that the dynamically registered callback and the callback submitted during token exchange used different names for the same loopback listener:

| Stage | Redirect URI |
| --- | --- |
| Dynamic client registration | `http://localhost:<port>/mcp/oauth/callback` |
| Token exchange | `http://127.0.0.1:<port>/mcp/oauth/callback` |

Better Auth compares redirect URIs exactly, so the hostname difference caused it to reject the token exchange.

After applying the normalization, `/api/auth/mcp/token` returned `200`, `opencode mcp list` reported Openship as connected, and all 149 MCP tools were available.

## Related issue

None.

## Changes

**apps/api**

- Add an MCP redirect URI normalizer for authorize and token requests.
- Resolve the registered redirect URIs for the requesting OAuth client.
- Treat only `localhost` and `127.0.0.1` as interchangeable.
- Require protocol, port, credentials, path, query, and fragment to remain equal.
- Leave malformed requests for Better Auth to validate normally.
- Add regression tests for authorize and token normalization.
- Add negative cases for public hosts, alternate ports and paths, IPv6, other loopback addresses, and lookalike domains.

## Verification

With the loopback alias fallback disabled, the regression suite fails in the three affected behaviors:

```bash
$ bunx bun@1.3.10 run --cwd apps/api test -- test/lib/oauth-redirect.test.ts

 Test Files  1 failed (1)
      Tests  3 failed | 5 passed (8)
```

```text
× aliases localhost and 127.0.0.1 for the same callback
× rewrites the token request to the registered loopback URI
× rewrites the authorize request before the redirect URI is persisted
```

With the change enabled:

```bash
$ bunx bun@1.3.10 run --cwd apps/api test -- test/lib/oauth-redirect.test.ts

 Test Files  1 passed (1)
      Tests  8 passed (8)
```

Full monorepo suite, with Docker-dependent tests skipped because no local daemon was available:

```bash
$ DOCKER_HOST=unix:///tmp/openship-no-docker.sock \
    bunx bun@1.3.10 run test -- --concurrency=1

```

The API workspace result within that run:

```text
 Test Files  171 passed | 4 skipped (175)
      Tests  1849 passed | 15 skipped (1864)
```

Typecheck and formatting:

```bash
$ bunx bun@1.3.10 run --cwd apps/api lint
$ tsc --noEmit

$ bunx bun@1.3.10 x prettier --check \
    apps/api/src/lib/oauth-redirect.ts \
    apps/api/src/modules/auth/auth.routes.ts \
    apps/api/test/lib/oauth-redirect.test.ts
All matched files use Prettier code style!
```

## Checklist

### DIFF
--- a/apps/api/src/lib/oauth-redirect.ts
+++ b/apps/api/src/lib/oauth-redirect.ts
@@ -1,0 +1,129 @@
+/**
+ * Return the registered redirect URI when the requested URI is an equivalent
+ * loopback address. OAuth normally compares redirect URIs exactly, but native
+ * MCP clients commonly alternate between localhost and 127.0.0.1 for the same
+ * callback listener.
+ *
+ * This deliberately only aliases loopback hostnames and requires every other
+ * URL component to match. Public redirect URIs remain exact-match only.
+ */
+export function findEquivalentLoopbackRedirectUri(
+  requestedUri: string,
+  registeredUris: readonly string[],
+): string | undefined {
+  const exact = registeredUris.find((uri) => uri === requestedUri);
+  if (exact) return exact;
+
+  const requested = parseRedirectUri(requestedUri);
+  if (!requested || !isLoopbackHostname(requested.hostname)) return undefined;
+
+  return registeredUris.find((registeredUri) => {
+    const registered = parseRedirectUri(registeredUri);
+    if (!registered || !isLoopbackHostname(registered.hostname)) return false;
+
+    return (
+      registered.protocol === requested.protocol &&
+      registered.port === requested.port &&
+      registered.username === requested.username &&
+      registered.password === requested.password &&
+      registered.pathname === requested.pathname &&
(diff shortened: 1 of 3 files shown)


Hydralerne:
thank you

Issue #346, opened 2026-07-31
Add first-class Codex setup to the MCP client picker


M-DGayle:
## Problem

**Settings → MCP** currently provides first-class setup recipes for Claude Code, Cursor, VS Code, Claude Desktop, Windsurf, and Zed, plus a generic fallback. OpenAI Codex supports Streamable HTTP MCP servers and OAuth, but Codex users must currently infer the setup from the generic option.

## Proposed change

Add a **Codex** option to the existing MCP client picker with the native local-client flow:

```bash
codex mcp add openship --url https://<your-openship>/api/mcp
codex mcp login openship
```

The recipe should note that local Codex surfaces share MCP configuration (desktop app, CLI, and IDE extension) and may need to restart after the server is added.

Update the MCP documentation to include the same native Codex flow and clearly distinguish local Codex clients from hosted/cloud execution. A private self-hosted OpenShip endpoint should not be presented as reachable by a hosted Codex task unless that environment actually has network access to it.

## Why this should fit the existing architecture

OpenShip already provides the required protocol and authorization surfaces:

- Streamable HTTP JSON-RPC at `/api/mcp`
- OAuth protected-resource and authorization-server discovery
- dynamic client registration
- authorization-code flow with S256 PKCE
- read-only/full-control and resource-scoped grants
- per-call permission enforcement

A compatibility probe against a current self-hosted OpenShip instance confirmed the expected unauthenticated `401` challenge and advertised OAuth metadata. The current Codex CLI exposes the matching `mcp add --url` and `mcp login` commands.

This proposal should therefore require **no new endpoint, schema, dependency, or authentication behavior**.

## Suggested implementation scope

- Add a Codex brand entry and setup recipe in `apps/dashboard/src/app/(dashboard)/settings/_components/McpConnection.tsx`.
- Update `apps/web/content/docs/mcp.mdx` and the dashboard settings documentation where the supported client list is described.
- Keep OAuth as the recommended path; retain the existing static-token fallback for users who intentionally choose it.
- Avoid unrelated MCP server or authorization changes.

## Validation approach

- Complete the real Codex OAuth flow against self-hosted OpenShip.
- Confirm Codex can initialize, list tools, and make an allowed read-only call.
- Confirm the OpenShip grant remains scoped and revocation disconnects the client.
- Run the existing test suite, dashboard typecheck/lint, and formatting checks.
[...]

PR #376, opened 2026-08-01
fix(dashboard): rewrite proxied OAuth redirects to public origin

M-DGayle:
## Summary

- rewrite same-origin `Location` responses from the private dashboard API upstream to the configured public OpenShip origin
- leave redirects to external origins unchanged, including native MCP clients' loopback OAuth callbacks
- add focused regression coverage for the OpenShip self-redirect and the client callback boundary

## Root cause

The same-host dashboard proxy streamed upstream response headers verbatim. During MCP OAuth, Better Auth adds `prompt=consent` by redirecting back to the authorize endpoint using the API request's internal origin (`http://127.0.0.1:4000`). A remote browser therefore followed the private VM address and failed before the consent page could load.

## Impact

External MCP clients such as Codex can complete OAuth against a self-hosted OpenShip instance served through the dashboard proxy. The API and dashboard may remain bound to loopback; this change does not require opening another port or weakening authentication.

## Validation

- reproduced the released v0.5.0 behavior: the public authorize request returned `Location: http://127.0.0.1:4000/...`
- deployed the same rewrite to a loopback-only self-hosted instance and verified the redirect now stays on its HTTPS public origin with no internal port
- completed `codex mcp login openship`; `codex mcp list` reports the server as `OAuth`
- `bun run --cwd apps/dashboard test` — 123 tests passed
- `bunx tsc --noEmit -p apps/dashboard/tsconfig.json` — passed
- `bunx prettier --check ...` for all changed files — passed
- `NEXT_PUBLIC_API_PROXY=true INTERNAL_API_URL=http://127.0.0.1:4000 OPENSHIP_PUBLIC_URL=https://ops.example.com bun run --cwd apps/dashboard build` — passed

The repository's current `bun run --cwd apps/dashboard lint` command was also attempted, but Next 16 interprets `next lint` as a project-directory argument and exits with `Invalid project directory .../apps/dashboard/lint`. The TypeScript check and production build both pass.

## Exclusions

- no OAuth scopes, consent behavior, cookies, or session handling changed
- no new endpoint, dependency, database change, or public network exposure
- no unrelated formatting changes

## Risk

The rewrite applies only when the upstream `Location` resolves to the same origin as the configured private API upstream. Redirects to a Codex/Claude loopback callback or any other external origin are preserved byte-for-byte.

Related to #346 and #119.

### DIFF
--- a/apps/dashboard/src/app/api/proxy/[...path]/route.ts
+++ b/apps/dashboard/src/app/api/proxy/[...path]/route.ts
@@ -26,6 +26,7 @@
  */
 
 import type { NextRequest } from "next/server";
+import { rewriteUpstreamLocation } from "@/lib/proxy-response";
 
 // Hop-by-hop headers — MUST NOT be forwarded by a proxy.
 // Per RFC 7230 §6.1 (HTTP/1.1) — Next.js's fetch automatically handles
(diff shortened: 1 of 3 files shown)


Hydralerne:
Verified against current main (`4fefe217`): the reported MCP consent self-redirect is already handled in the API's `forceMcpConsent` middleware, which builds its URL with `publicRequestUrl` before the Better Auth handler. The middleware is mounted before `/api/auth` in `apps/api/src/app.ts`.

The real middleware passes four request/replay cases for signed-in and signed-out users behind the dashboard proxy. The redirect stays on `https://ops.example.com`, and the registered client callback remains unchanged, including when its loopback port happens to equal the private API port. The existing public-origin, metadata and redirect-URI tests also pass.

Closing this as superseded by the API-side fix on main. An additional blanket `Location` rewrite would duplicate that correction and could rewrite a legitimate client callback that shares the private upstream's origin. Thank you for the original reproduction and verification.

PR #377, opened 2026-08-01
Add MiniMax email compose provider

octo-patch:
Reason: The registered email compose and subject-generation endpoints need a configurable MiniMax implementation instead of inert stubs.

This change:

- adds MiniMax-backed compose and subject generation
- supports MiniMax-M3 and MiniMax-M2.7 through the global and CN OpenAI-compatible endpoints
- validates and forwards the provider configuration for managed email deployments
- adds provider request and route coverage

Checks:

- `git diff --check origin/main..HEAD`
- secret scan against the committed upstream diff


octo-patch:
Updated this PR for the current main branch and pushed commit 9d8558f. MiniMax configuration now flows through the catalog-based webmail installer while retaining both supported models and the global/CN endpoints. I ran the webmail catalog contract test, core catalog validation, MiniMax tests, and API/email TypeScript checks.

octo-patch:
Correction: I reverted the broad current-main merge to keep this PR limited to the original MiniMax changes. The focused MiniMax test suite passes (5 tests).

octo-patch:
Correction complete: I restored the synced base tree so the final PR diff is limited to the 14 MiniMax provider, model, and endpoint files, with no Japanese or Chinese locale changes. The webmail contract (7), core catalog (26), and MiniMax (5) tests pass, along with the API and email-server TypeScript checks.

Issue #256, opened 2026-07-28
Support ACME External Account Binding (EAB / HMAC) for certificate issuance


tristanbob:
# Support ACME External Account Binding (EAB / HMAC) for certificate issuance

**Type:** Feature request
**Area:** Domains & SSL / edge (OpenResty)

## Summary

Openship currently issues certificates from Let's Encrypt, which allows anonymous account creation. Most other public ACME CAs — and essentially every internal/enterprise ACME CA — require **External Account Binding** (RFC 8555 §7.3.4): a key identifier (`kid`) plus an HMAC key that is used to sign the new-account request, tying the ACME account to an existing account at the CA.

Without EAB support, Openship can only ever talk to Let's Encrypt, and users who need a different CA have to bypass the built-in TLS management entirely.

## Motivation

Concrete cases where Let's Encrypt-only is a blocker:

- **ZeroSSL, Google Trust Services, SSL.com, Sectigo, DigiCert** — all require EAB credentials for their ACME endpoints.
- **Let's Encrypt rate limits.** Teams deploying many preview environments per repo hit the certificates-per-registered-domain limit; a second CA is the usual escape hatch.
- **Internal PKI.** `step-ca`, EJBCA, and Vault's ACME endpoint are commonly configured to require EAB so that only provisioned clients can enroll. This matters for anyone running Openship on a private network or in an air-gapped environment where the Let's Encrypt HTTP-01 challenge can't complete anyway.
- **Compliance.** Some orgs are contractually required to use a specific commercial CA.
- **Redundancy.** A configurable fallback CA means a Let's Encrypt outage or account suspension doesn't stop renewals.

## Current behavior

TLS is described as "Auto Let's Encrypt, wildcards, unlimited domains, auto-renewal", with no documented way to point the edge at a different ACME directory or to supply EAB credentials. As far as I can tell there's no config surface for this in `.env.example` or the CLI.

## Proposed solution

Introduce a **certificate authority profile** as a first-class config object, with EAB as optional fields on it.

### Configuration

```
OPENSHIP_ACME_DIRECTORY_URL=https://acme.zerossl.com/v2/DV90
OPENSHIP_ACME_EMAIL=ops@example.com
OPENSHIP_ACME_EAB_KID=<key identifier>
OPENSHIP_ACME_EAB_HMAC_KEY=<base64url HMAC key>
OPENSHIP_ACME_KEY_TYPE=ec256          # ec256 | ec384 | rsa2048 | rsa4096
[...]

PR #380, opened 2026-08-01
feat: support alternate ACME CAs and EAB

tristanbob:
## Summary

Adds optional support for ACME certificate authorities other than Let's Encrypt — alternate directory URL, External Account Binding credentials, key type, and a private trust bundle — and fixes renewal for certificates whose issuing CA has changed. With none of the new variables set, nothing changes.

## Motivation

Openship can only talk to Let's Encrypt today. ZeroSSL, Google Trust Services, and essentially every internal CA (`step-ca`, EJBCA, Vault) require External Account Binding (RFC 8555 §7.3.4), so anyone who needs a different CA has to bypass the built-in TLS management entirely. #256 has the full case, including Let's Encrypt rate limits and air-gapped installs.

This is the smallest slice that helps: one instance-wide CA, configured by environment variable. That's deliberately the same tier `lib/mail.ts` already uses for SMTP — operator settings first, static env vars as the deployment fallback.

## Related issue

#256. **No maintainer has agreed to scope or approach yet**, and per CONTRIBUTING that makes this premature — I opened it to make the proposal concrete rather than to jump the queue. Happy for it to sit until someone weighs in, or to close it.

## Changes

**`packages/adapters` — `infra/nginx.ts`**

- `NginxProvider` gains `acmeDirectoryUrl`, `acmeEabKid`, `acmeEabHmacKey`, `acmeKeyType`, `acmeCaBundle`, and `acmeTosAgreed`, applied to both issuance and renewal.
- **EAB credentials never reach argv.** They go into a mode-`0600` file inside a fresh `0700` directory — the same contract `runtime/git-ssh-material.ts` documents for SSH key material — passed to certbot via `--config` and removed as a unit afterwards.
- **HMAC redaction lives in `_execCertbot`**, not in each caller's catch. That matters: the scheduled renewal sweep has no catch of its own, and a failed renewal emits an `ssl.renewal_failed` notification carrying the certbot error to email and webhooks.
- **Renewal survives a CA switch.** A lineage issued by one directory cannot be renewed against another — `certbot renew` neither registers the new account nor carries EAB — so `renewCert` compares the lineage's recorded `server` against the configured directory and reissues on mismatch. One-time, self-healing. Without this, every pre-switch certificate fails renewal silently in the nightly sweep.

**`apps/api`** — env schema plus boot-time validation, so a malformed EAB pair or a relative CA-bundle path fails at startup naming the variable, not mid-deploy. Threaded into every provider-construction path (deploys, edge takeover, self-edge).

**`apps/cli`** — `openship up` preserves the variables across upgrades, so an upgrade cannot silently discard operator config.

**`docs/`** — `acme.md` (Let's Encrypt staging, ZeroSSL, private CA) and `acme-testing.md` (a throwaway EAB-requiring CA in Docker).

## Verification

Unit suites, all passing: `packages/adapters` 750, `apps/cli` compose-preserve 20, proxy-takeover 6. Typechecks clean for `packages/adapters`, `apps/api`, and `apps/cli`.

More usefully, the whole path was exercised against **a real EAB-requiring CA** — Pebble plus `pebble-challtestsrv` plus real certbot, all in Docker. `docs/acme-testing.md` reproduces it.

- A certificate was issued end to end using the exact argv `provisionCert` composes, including the EAB config file: `issuer=CN=Pebble Intermediate CA`, key `NIST CURVE: P-384` — confirming the `ec384` key-type mapping actually takes effect. All four key-type mappings were checked against real certbot.
- The EAB file format is genuinely consumed, not silently ignored: registering with a deliberately wrong HMAC is rejected with `external account binding JWS verification error`.
- Renewal succeeds with the argv `renewCert` now composes — no `--server`, no EAB config — confirming EAB is needed only at account registration.
- The renew-versus-reissue decision was checked against the real `renewal/<domain>.conf` certbot wrote, not a fixture: the recorded `server` parses correctly and both branches resolve as designed.

That exercise also caught a bug the unit tests could not. The directory probe sent no `User-Agent`, which RFC 8555 §6.1 requires and real servers enforce, so it failed against any conforming CA.

Not covered: issuance driven through `NginxProvider` against a live OpenResty edge. The argv, EAB file, key mapping, and renewal contract are all verified above; the provider's file handling around them is unit-tested only.

## Checklist

On the unchecked box: `bun format` was deliberately not run — it reformats roughly 1,580 pre-existing files, which CONTRIBUTING says to drop rather than bundle, so new code matches surrounding style instead. The full `bun run test` leaves failures in `apps/cli` `ports.test.ts` and three `apps/api` e2e tests that reproduce identically on the unmodified base branch here; nothing in this PR touches them.

---

*Follow-up layers — DB-backed CA profiles, API, CLI, dashboard, and a per-domain pin — exist as draft PRs on my fork ([DB](https://github.com/tristanbob/openship/pull/1), [API](https://github.com/tristanbob/openship/pull/2), [CLI](https://github.com/tristanbob/openship/pull/3), [dashboard](https://github.com/tristanbob/openship/pull/4), [per-domain](https://github.com/tristanbob/openship/pull/5)), so the question this PR really raises — is this the right configuration shape — can be judged concretely. None are proposed for merge, and this PR stands alone without them.*

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -81,6 +81,19 @@ INTERNAL_TOKEN=change-me-32-byte-random-hex
 # shown — override to avoid conflicts with something already on the host.
 # API_PORT=4000
 # DASHBOARD_PORT=3001
+
+# Certificate authority / ACME (self-hosted managed edge)
+# Defaults are unchanged when these are omitted: Let's Encrypt production,
+# Certbot's default EC key, and automatic renewal.
+# OPENSHIP_ACME_EMAIL=ops@example.com
+# OPENSHIP_ACME_DIRECTORY_URL=https://acme.zerossl.com/v2/DV90
+# EAB credentials must be set together. The HMAC key must be base64url encoded.
+# OPENSHIP_ACME_EAB_KID=
+# OPENSHIP_ACME_EAB_HMAC_KEY=
+# OPENSHIP_ACME_KEY_TYPE=ec256       # ec256 | ec384 | rsa2048 | rsa4096
+# OPENSHIP_ACME_CA_BUNDLE=/etc/ssl/private/acme-root.pem
+# OPENSHIP_ACME_TOS_AGREED=true
+# See docs/acme.md for ZeroSSL, private-CA, and container mount examples.
 # WEB_PORT=3000   # landing site — root control-plane compose (docker-compose.yml) only
 
 # ─── Image source (self-hosted pull-based compose) ───────
--- a/apps/api/src/config/env.ts
+++ b/apps/api/src/config/env.ts
@@ -414,6 +414,42 @@ const envSchema = z.object({
     .min(16 * 1024)
     .max(8 * 1024 * 1024)
     .default(512 * 1024),
+}).superRefine((cfg, ctx) => {
+  // Fail ACME misconfiguration HERE, at boot, with the variable named — not at
+  // the first deploy, where NginxProvider's constructor re-checks the same rules
+  // (that check stays: the adapter also serves non-env callers). Empty/whitespace
+  // values count as unset, matching resolveAcmeProviderOptions.
+  const kid = cfg.OPENSHIP_ACME_EAB_KID?.trim();
+  const hmac = cfg.OPENSHIP_ACME_EAB_HMAC_KEY?.trim();
+  const bundle = cfg.OPENSHIP_ACME_CA_BUNDLE?.trim();
+  if (!!kid !== !!hmac) {
+    ctx.addIssue({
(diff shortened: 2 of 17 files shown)


tristanbob:
Thanks for reviewing my PR!

To fully implement this feature, we need this PR and the stacked PRs on my fork that are fully tested end-to-end for creating and renewing certs using EAB auth.

-- Tristan

Hydralerne:
Good job, i reviewed it, really nice thank you

Issue #75, opened 2026-07-21
[Feature]: First-class GitLab integration (connect, push-to-deploy, MR previews)


ThomasRbn:
## Summary

Add first-class GitLab support alongside the existing GitHub integration - connect an account or self-hosted instance, browse groups/projects/branches, bind a project, clone, and auto-deploy on push - wired the same way users already know from GitHub.

Roadmap intent: repositories, branches, merge-request preview environments, and push-to-deploy.

This proposal intentionally keeps the existing deploy architecture unchanged:

```text
Git provider → webhook / poll → Openship API → clone → build → deploy runtime
```

GitHub remains fully supported. GitLab becomes a peer source provider, not a replacement.

This is not a proposal to:

- Remove or replace the GitHub module.
- Sync GitLab issues or full merge-request timelines into Openship.
- Require extracting a shared `GitProvider` abstraction before shipping v1.
- Bundle Bitbucket in the same change.
- Change the deploy / build / runtime pipeline for existing GitHub projects.
- Mirror repositories into Openship storage.

The only functional change is which Git host a project can connect to and deploy from.

## Why this is possible

Openship already reserves GitLab in the project source model:

```ts
// packages/core/src/project-source.ts
export const SOURCE_PROVIDERS = [
  "github",
  "gitlab",
  "bitbucket",
  "local",
  "upload",
  "release",
] as const;
```

The `project.gitProvider` column and `git_installation.provider` column already accept provider discriminators. Installation tracking is not GitHub-exclusive at the schema level:

```text
packages/db/src/schema/github.ts   → git_installation.provider (default "github")
packages/db/src/schema/project.ts  → gitProvider / gitOwner / gitRepo / gitBranch / …
```

What is missing is the GitLab auth, API, webhook, clone, and dashboard surfaces. GitHub today lives entirely under:

```text
apps/api/src/modules/github/
```

GitLab can ship as a peer module with the same product surface, without rewriting GitHub.

## Current Openship behavior

GitHub integration covers:

1. **Connect** - GitHub App, Better Auth OAuth, local `gh` CLI, PAT, per-server auth.
2. **Browse** - live repository / branch listing (not a mirrored repos table).
3. **Bind** - project fields: `gitProvider`, `gitOwner`, `gitRepo`, `gitBranch`, `gitUrl`, `installationId`, webhook fields.
4. **Deploy** - clone credentials via `clone-auth.ts` / token dispatcher.
[...]

PR #386, opened 2026-08-01
refactor:vcs strategy refactor

### LINKED ISSUE #379: Help Wanted: Replace GitHub-Only Integration with Multi-Git Provider Support
Hello guys! I know this is a pretty big surgery, but right now OpenShip only supports GitHub. Before the LTS release, supporting self-hosted Git and GitLab will be mandatory

This requires a proper abstraction layer and a major refactor across the codebase. If anyone is interested in taking this on, it would be amazing. I'm currently focused on fixing bugs and improving the platform's stability, so I don't have the bandwidth to work on it myself.

mohamedadel96e:
## Motivation & Context
Currently, the codebase is tightly coupled to GitHub as the sole source code provider, primarily through direct dependencies on `github.service.ts` across deployment and project modules. As we aim to support additional providers (such as GitLab and Self-Hosted options), this hard dependency creates friction and scaling challenges.

This PR resolves this technical debt by introducing a provider-agnostic Version Control System (VCS) abstraction layer using the Strategy pattern. This allows the core application to interact with git repositories purely through interfaces, decoupling it from any specific provider implementation while maintaining full backwards compatibility for existing GitHub projects.

## Related issue

Closes #379 

### 1. Strategy Pattern Abstraction
- Introduced a `VcsProviderStrategy` interface that defines the contract for all repository operations (e.g., getting branches, commits, fetching files, resolving webhooks).
- Implemented `VcsStrategyFactory` which serves as the central resolver, dynamically returning the correct implementation (e.g., `github`, `gitlab`) based on the project's configured `gitProvider`.

### 2. Generic API Endpoints (`/api/vcs/:provider`)
- Extracted generic repository endpoints out of `github.routes.ts` into a dedicated set of generic `vcs.routes.ts` and `vcs.controller.ts`.
- The dashboard frontend (`apps/dashboard/src/lib/api/`) has been migrated to consume the new `/api/vcs/github/` routes, removing hardcoded assumptions about GitHub from the client-side API layer.

### 3. Core Modules Migration
Refactored the core domain modules to rely entirely on `VcsStrategyFactory`:
- **Deployments:** `build-pipeline.ts`, `preflight.ts`, and `prepare.service.ts` now resolve the appropriate strategy using `project.gitProvider` before executing git operations.
- **Projects:** `project-crud.service.ts` and `project.controller.ts` dynamically handle project creation, repo linking, and tear-downs using the generic VCS interface.
- **Webhooks:** `vcs.webhook.processor.ts` now accepts a generic `VcsPushPayload` and dynamically handles incoming push events regardless of the origin.

### 4. Database Schema Alignment
- Adapted schema typing to account for abstract provider configurations (`gitProvider`, etc.) to ensure long-term flexibility without losing any existing GitHub data.

## Verification & Testing
To guarantee strict backwards compatibility and ensure no regressions were introduced, the following checks were performed locally:

- **Strict Diffing:** Formatting drift (`prettier` output on unmodified lines) was manually scrubbed from the diff to ensure strict compliance with the contributing guidelines. Only logical changes are included in this PR.
- **Type Checking:** Ran `bun run tsc --noEmit` locally on the modified packages (`apps/api`, `apps/dashboard`, `packages/db`). All packages compile perfectly with zero TypeScript errors.
- **Unit & Integration Tests:** Executed the root `bun run test` suite across all workspaces. All 1,877 tests passed flawlessly.
- **Production Build:** Ran the root `bun run build` command, successfully compiling the optimized assets and static routes for both the API and the Dashboard without issue.

## Screenshots
image

these changes don't fully close the issue if we want to support another git provider right now, we should implement each strategy.


Hydralerne:
huge, gonna review

mohamedadel96e:
@Hydralerne, this is ready for another look. It changed quite a bit since your “huge, gonna review” comment, so rather than making you go through the entire diff again, here’s a summary of what changed.

## Rebase and cleanup

The branch was **217 commits behind `main`**, so I rebased it onto the current base.

Before rebasing, I squashed the two original commits because the second one completely reverted the first one's changes in:

* `build-pipeline.ts`
* `preflight.ts`
* `prepare.service.ts`
* `github.service.ts`

Replaying both separately would have meant resolving conflicts in code that ultimately gets discarded anyway.

For every conflicted file, I started from `main` and reapplied only the actual semantic changes. As a result, there are now **no whitespace-only hunks against `main`**.

## Four regressions fixed

I also found and fixed four regressions that could have affected existing GitHub users. They were all missed by TypeScript because values were being passed through the adapter using `as any` / `as unknown as`, so the previous version still type-checked and passed its tests.

### 1. Webhook ID was never persisted

`registerWebhook` returned GitHub's `{ hookId }`, cast as `VcsWebhook`, while `ensureSharedWebhook` reads `.id`.

That meant the guard always failed, `syncSharedWebhookId` never ran, and `project.webhookId` was never persisted. As a result, repo webhook auto-deploy could silently stop being wired up.

The existing test had also been changed to return `{ id }`, which masked the issue.

This is now fixed and covered by a regression test.

### 2. Directory listing was routed through file-content logic

`listFiles` was going through `getFileContent`, which base64-decodes `data.content`.

That works for files, but GitHub returns an array for directories, so the directory-listing endpoint could no longer list directories correctly.

This now has a dedicated test.

### 3. Tree entry types did not match the controller

`getTree` passed `"file"` / `"dir"` entries through a type declaring `"blob"` / `"tree"` / `"commit"`.

The controller checks:

```ts
entry.type === "tree"
```

So that condition was never true, causing directories to be interpreted as files.

The mapping is now corrected and covered by a test.

### 4. Check runs lost their Details link

`detailsUrl` had been commented out on both check-run paths, while `service-checks.ts` still passed it.

As a result, check runs lost their Details link.

Both paths are restored and pinned by tests.

## Removed duplicated and dead code

I removed a few things rather than keeping them around.

### `vcs.webhook.processor.ts`

This was effectively a verbatim copy of `webhook-push.ts`, leaving the original orphaned with zero importers.

That was risky because `webhook-push.ts` contains logic such as the cross-tenant fan-out guard. Any future fix made there could merge cleanly while having no effect on the code that actually runs.

There is now a single path instead of two copies.

### `github.routes.ts`

The route was rebinding `c.req.param` on the live Hono request just to inject a parameter that was not actually part of the route path.

Handlers now receive the provider explicitly instead.

### `github.controller.ts`

I also removed roughly **360 lines of unrouted handlers** that were no longer part of the active flow.

## The VCS seam is now real

Previously, the abstraction was mostly nominal.

For example:

```ts
export type VcsPushPayload = GitHubPushPayload
```

That means a future GitLab strategy would still be required to produce GitHub-shaped data.

The VCS types are now provider-neutral rather than aliases for GitHub types.

I also changed the factory behavior. Unknown providers previously fell back to GitHub. It now throws explicitly, following the existing `UnknownDnsProviderError` pattern.

The GitLab and self-hosted stubs remain registered, but their methods now throw instead of returning plausible fake values.

One important example was `compareCommits`, which returned:

```ts
{ files: [] }
```

instead of `null`.

That could bypass this logic in `smart-route.ts`:

```ts
if (!compare) {
  resolvedForceAll = true
}
```

and potentially skip services that had actually changed.

## Scope was intentionally kept limited

`TODO.md` describes the broader follow-up work, including:

* Generalizing `GitHubSource` to `GitSource`
* Moving provider dispatch into `resolveBuildGitToken` / `tokenFor`

I deliberately did **not** include those changes here. The goal of this PR is to establish the seam without turning it into a much larger refactor.

I also left these untouched:

* Making `gitProvider` a checked union
* Using `gitUrl` as the clone source of truth
* The `clone-plan.ts` `repoIsGithub` bug caused by `!!gitOwner`
* `projectGitUrl()` hardcoding `github.com`

Several of those overlap with @Farahat612's groundwork from #75. I'm happy to rebase on top of that if you'd prefer those changes to land first, or fold them into this PR if that makes more sense.

## UI was intentionally unchanged

I did not rename **“Connect GitHub”** to **“Connect Version Control”**.

GitHub is still the only supported provider, and changing the copy now would increase the missing-translation baseline across all 7 locales without providing a user-facing benefit yet.

## Local verification

CI has not reported on this PR yet. I suspect the fork workflows need approval to run.

Everything is green locally:

```text
bun run --cwd apps/api lint
✓ clean

bun run --cwd packages/core lint
✓ clean

bun run --cwd packages/db lint
✓ clean

bun run --cwd packages/core test
✓ 53 files, 913 passed

bun run --cwd apps/api test
✓ 428 files, 5241 passed / 3 skipped
```

The PR is currently split into **five independent commits**, and each one is green on its own. If reviewing the full PR is still too much in one go, I'm also happy to split it into smaller PRs.

Hydralerne:
checking this

Farahat612:
Sorry for the slow reply. Fold them into #386. My groundwork was never opened as a PR, and
yours is the one in review, so it shouldn't wait on mine.

If it's useful as reference, the branches are on my fork:
[refactor/git-provider-checked-union](https://github.com/Farahat612/openship/tree/refactor/git-provider-checked-union),
with [refactor/persist-project-remote](https://github.com/Farahat612/openship/tree/refactor/persist-project-remote)
stacked on it. They predate the platform refactor, so treat them as notes, not a base.

The piece most worth carrying over is still live on main: `repoIsGithub` is derived from
owner presence rather than the provider column (`!!project.gitOwner` in
`build-pipeline.ts:911`, `!!opts?.gitOwner` in `preflight.ts:1578`), so a GitLab row with an
owner would take the GitHub tarball fast path.

PR #413, opened 2026-08-03
fix(adapters): detect untagged database images in backup autodetection

shuvamk:
## Summary

Backup auto-detection required a tag on the service image, so `image: postgres` took a crash-consistent volume tar instead of `pg_dump`. All four database producers now accept the untagged spelling.

## Motivation

`image: postgres` with no tag is valid compose and is stored verbatim on the service row, but every producer gated on a colon after the image name. Measured through `autoDetectProducer` with the real registry:

| service image | main | this branch |
| --- | --- | --- |
| `postgres:16` | pg_dump | pg_dump |
| `postgres` | **volume** | pg_dump |

Same mechanism, same result for `mysql`, `mariadb`, `mongo`, `redis`, `postgis/postgis` and `percona/percona-server(-mongodb)`: tagged detects, untagged falls through.

A tar of a live PGDATA is crash-consistent, not app-consistent — not the backup the policy asked for, and nothing reports the downgrade. The policy editor already renders "Detected PostgreSQL — Auto backs it up with pg_dump" for `image: postgres` while the run tars the volume, and it offers only auto / volume / custom_command, so there is no way to ask for the dump.

## Related issue

None.

## Changes

- `producers/{pg-dump,mysql-dump,mongo,redis}.ts` — each image regex gains `$` as an alternative terminator to `:`. Each `Detection:` header line is updated to the regex it describes.
- `producers/detect.test.ts` — new.

`/` is deliberately **not** added to the postgres, mysql and mongo patterns: that would admit the whole `<db>/<repo>` namespace, and `mysql/mysql-router` is a proxy with no data and no `mysqldump` binary whose documented compose env carries `MYSQL_PASSWORD` — it would satisfy the credential guard, fail the dump, and turn a green zero-artifact run red. Redis keeps the `/` it already had, which `redis/redis-stack` relies on.

So the image match widens by exactly the untagged spelling. The env guards are untouched (untagged `postgres` with no `POSTGRES_DB`/`PGDATABASE` still falls back to volume); a digest reference `postgres@sha256:…` still resolves to volume; and `postgrest`, `postgresql`, `postgres-exporter`, `myapp/postgres`, `redis-sentinel`, plus every `<db>/<repo>` form, resolve to `volume` before and after.

## Verification

Five tests in `detect.test.ts`. The untagged case fails with only the source files reverted:

```bash
$ npx vitest run src/backup/producers/detect.test.ts   # fix reverted, tests kept
 ❯ src/backup/producers/detect.test.ts (5 tests | 1 failed)
     × routes an untagged database image to its dump producer
AssertionError: expected { postgres: 'volume', …(7) } to deeply equal { postgres: 'pg_dump', …(7) }
-   "postgres": "pg_dump",   +   "postgres": "volume",

$ npx vitest run src/backup/producers/detect.test.ts   # fix restored
      Tests  5 passed (5)
```

The `<db>/<repo>` test passes on main too, on purpose — it pins the boundary this change could have opened, and I checked it by inversion: widening the three terminators to `[:/]` makes that test, and only that test, fail.

Full suite green, `tsc --noEmit` clean in `packages/adapters` and `apps/api`.

## Checklist

- [x] `bun run test` and `bun run --cwd packages/adapters lint` pass locally. I did not run the repo-wide `bun format` — `prettier --check` already fails on three of these files on `origin/main` and I did not want to churn that drift; I checked the files individually instead, and every line this PR adds is prettier-clean.

### DIFF
--- a/packages/adapters/src/backup/producers/detect.test.ts
+++ b/packages/adapters/src/backup/producers/detect.test.ts
@@ -1,0 +1,130 @@
+import { describe, expect, it } from "vitest";
+import "../index";
+import { autoDetectProducer } from "./detect";
+import type { PayloadKind, ServiceHandle } from "../types";
+
+const POSTGRES_ENV = { POSTGRES_DB: "app", POSTGRES_USER: "app", POSTGRES_PASSWORD: "pw" };
+const MYSQL_ENV = { MYSQL_ROOT_PASSWORD: "pw", MYSQL_DATABASE: "app" };
+const MYSQL_ROUTER_ENV = {
+  MYSQL_HOST: "db",
+  MYSQL_PORT: "3306",
+  MYSQL_USER: "app",
+  MYSQL_PASSWORD: "pw",
+};
+const MONGO_ENV = { MONGO_INITDB_ROOT_USERNAME: "root", MONGO_INITDB_ROOT_PASSWORD: "pw" };
+const REDIS_ENV = { REDIS_PASSWORD: "pw" };
+
+function serviceWith(image: string, env: Record<string, string>): ServiceHandle {
+  return {
+    id: "svc-1",
+    projectId: "proj-1",
+    name: "db",
+    image,
+    env,
+    volumes: [],
+    containerId: "c1",
+    projectSlug: "app",
+    namespaceVolumes: true,
+  };
+}
+
+function kindsFor(cases: Array<[string, Record<string, string>]>): Record<string, PayloadKind> {
+  const out: Record<string, PayloadKind> = {};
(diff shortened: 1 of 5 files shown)


Hydralerne:
Verified against current main: the shared `packages/core/src/backup-image-detect.ts` registry already handles untagged images, and all four database producers use it through `isDbImage`. The same registry supplies the dashboard hint, avoiding the duplicated regex definitions this PR originally modified.

I ran this PR's original untagged-image test against the current producers: all eight database image cases select their logical backup producer. Current core tests also cover tagged/untagged forms, namespaced images, false-positive boundaries, registry aliases and pinned digests.

Closing as superseded without merging the older per-producer regex changes. The fix is already on main; this decision is recorded in the #928 review ledger.

PR #414, opened 2026-08-03
fix(dashboard): enable rollback for success-mapped deployment rows

### LINKED ISSUE #410: Rollback action permanently disabled on every deployment — status string mismatch ("ready" vs "success")
## Summary

On a self-hosted instance (v0.5.0, "you're on the latest version" per Settings → Instance), the **Rollback** action ("Reverter para esta versão" in the pt locale) is permanently disabled/greyed-out on **every** deployment in a project's history — including old, non-active releases that show the "Snapshotted" chip (so a retained artifact clearly exists) and have a real commit hash.

## Environment

- Self-hosted, Docker deploy mode
- Version: **0.5.0** (Settings → Instance reports "Você está na versão mais recente · Atual: 0.5.0")
- Project type: Docker Compose, 3 services (app + postgres + postgres-exporter)
- Deploys are triggered via a project-level **incoming webhook** (Settings → Webhooks, `actionType: "deploy"`) called from an external CI pipeline (GitHub Actions `curl -X POST .../api/webhooks/incoming/:id`), not via the native GitHub App push integration (that integration is intentionally left "Desligado"/off in this project so deploys only fire after the CI pipeline's own gates pass).

## Steps to reproduce

1. Deploy a project via the project's incoming deploy webhook (`POST /api/webhooks/incoming/:id`, `actionType: "deploy"`) a few times, so the Deployments list has several "Implantada"/"Deployed" rows with the "Snapshotted" chip on the non-active ones.
2. Open **Deployments** for the project.
3. Click the **⋮** menu on any row that is *not* the currently active version (e.g. a "v8" row while "v9" is active).
4. Observe **"Reverter para esta versão" / "Rollback"** is disabled (greyed out), regardless of which older version is picked.

## Expected

Per `apps/api/src/modules/deployments/rollback/PENDING.md`'s "Shipped" section and the 0.4.9 changelog entry ("Roll back any release, on every project ... available on every successful release"), Rollback should be enabled on any non-active, non-in-flight deployment that has either a retained artifact or a known commit.

## Actual

Rollback stays disabled on every single row in the project's history, even ones clearly showing the "Snapshotted" (retained-artifact) chip.

## Root cause (best guess from reading the source)

Three places disagree about what a deployment's *display* status is called, and the rollback button's eligibility check ends up checking a value the list never actually sends:

- **Where the DB status is written** — `apps/api/src/modules/deployments/deployment-lifecycle.ts`, `onSuccess()`, sets `status: "ready"` (line ~434). The schema doc-comment in `packages/db/src/schema/deployment.ts` also lists `ready` as the terminal-success value.
- **Where the list renders the green "Deployed" badge** — `apps/dashboard/src/app/(dashboard)/deployments/components/DeploymentCard.tsx` (`statusLabelMap`) and `apps/dashboard/src/app/(dashboard)/deployments/utils.ts` (`getStatusConfig`) both key off **`"success"`**, not `"ready"`. Neither has a `case "ready"` — anything not in the map falls back to the "Pending" label. Since these projects' rows *do* render as "Implantada"/"Deployed" (not "Pending"), the value actually reaching the frontend for a finished deploy must be `"success"`.
- **Where the rollback button decides if it's clickable** — `apps/dashboard/src/app/(dashboard)/deployments/components/DeploymentMenu.tsx`:
  ```ts
  const canRollback =
    (deployment.status === "ready" || deployment.status === "partial_failure") &&
    !deployment.isActive &&
    !isInFlight &&
    (!!deployment.artifactRetainedAt || hasCommit);
  ```
  This checks for `"ready"` — which, per the above, the list payload apparently never actually contains for a finished deploy (it's `"success"` there). So `canRollback` can never be true for a normally-completed deployment, no matter how old, snapshotted, or eligible it is.

Whether the "correct" fix is teaching `DeploymentMenu` to also accept `"success"`, or making the list stop presenting `"success"` and have `DeploymentCard`/`utils.ts` handle `"ready"` instead, is a call for whoever owns the canonical status vocabulary — I didn't chase down the exact place `"ready"` becomes `"success"` on the wire (likely the "orchestrator-aware listing endpoint" `DeploymentCard.tsx` refers to in its comments, which I didn't locate).

## Workaround confirmed

The server-side `rollbackDeployment()` / `previewRestore()` (`apps/api/src/modules/deployments/deployment.service.ts`) have **no `status` gate at all** — they only require a retained artifact or known commit via `resolveRestorePlan`. Calling `POST /api/deployments/:id/rollback` directly (CLI `openship deployment rollback <id>` or a raw authenticated request) works fine and bypasses the frontend's disabled button entirely. So this is a dashboard-only bug, not a broken capability.

Happy to attach screenshots of the disabled menu and the Instance/version screen if useful — didn't include them here since they show a private repo/project name.

santhiprakash:
Fixes #410

## Problem

On the deployments list, finished deploys show status `success` because `mapRowToDeployment` normalizes the API's terminal `ready` value for display/filtering. `DeploymentMenu` only enabled Rollback and Pin when `deployment.status === "ready"`, so `canRollback` was always false for normally completed deployments — even when the row showed the Snapshotted chip and had a retained artifact or commit SHA.

## Triage / Root cause

- API persists `ready` (`deployment-lifecycle.ts`).
- List mapping converts `ready → success` in `apps/dashboard/src/app/(dashboard)/deployments/utils.ts`.
- Rollback/pin gating in `DeploymentMenu.tsx` still checked only for `ready`.

## Fix

- Add `isDeploymentRollbackEligible()` beside the status map in `utils.ts`, accepting `ready`, `success`, and `partial_failure`.
- Use it for rollback enablement and pin visibility in `DeploymentMenu.tsx`.

## Verification

```bash
bun run --cwd apps/dashboard test src/utils/project-status.test.ts
```

11 tests passed, including new coverage that a `ready` API row mapped to `success` is rollback-eligible.

## Notes / Risks

- Dashboard-only change; server-side `rollbackDeployment()` was already ungated.
- No behavior change for in-flight or failed rows.

### DIFF
--- a/apps/dashboard/src/app/(dashboard)/deployments/components/DeploymentMenu.tsx
+++ b/apps/dashboard/src/app/(dashboard)/deployments/components/DeploymentMenu.tsx
@@ -14,6 +14,7 @@ import {
 import { generateIcon } from "@/utils/icons";
 import { deployApi, getApiErrorMessage } from "@/lib/api";
 import { useI18n, interpolate } from "@/components/i18n-provider";
+import { isDeploymentRollbackEligible } from "../utils";
 
 interface Deployment {
   id: string;
@@ -77,7 +78,7 @@ export const DeploymentMenu: React.FC<DeploymentMenuProps> = ({
   // "Redeploy this commit" fallback is gone: it was the same operation behind a
   // second label.
   const canRollback =
-    (deployment.status === "ready" || deployment.status === "partial_failure") &&
+    isDeploymentRollbackEligible(deployment.status) &&
     !deployment.isActive &&
     !isInFlight &&
     (!!deployment.artifactRetainedAt || hasCommit);
@@ -261,7 +262,7 @@ export const DeploymentMenu: React.FC<DeploymentMenuProps> = ({
 
           {/* Pin / Unpin — toggles the artifact's exemption from
               retention prune. Available for any ready deployment. */}
-          {!isInFlight && deployment.status === "ready" && (
+          {!isInFlight && isDeploymentRollbackEligible(deployment.status) && (
             <button
               onClick={handleTogglePin}
               disabled={!deployment.pinned && !deployment.artifactRetainedAt}
--- a/apps/dashboard/src/app/(dashboard)/deployments/utils.ts
+++ b/apps/dashboard/src/app/(dashboard)/deployments/utils.ts
@@ -1,6 +1,13 @@
 import type { Deployment } from "./types";
 import type { Dictionary } from "@/i18n";
 
+/** Terminal success statuses eligible for rollback/pin in the dashboard.
+ *  The API persists `ready`, but `mapRowToDeployment` normalizes that to
+ *  `success` for list rendering — menu gating must accept both shapes. */
+export function isDeploymentRollbackEligible(status: string): boolean {
+  return status === "ready" || status === "success" || status === "partial_failure";
+}
+
 export const mapRowToDeployment = (row: any): Deployment => {
   const statusMap: Record<string, Deployment["status"]> = {
     ready: "success",
--- a/apps/dashboard/src/utils/project-status.test.ts
+++ b/apps/dashboard/src/utils/project-status.test.ts
@@ -5,6 +5,8 @@ import {
   calculateDeploymentStats,
   filterDeployments,
   getStatusConfig,
+  isDeploymentRollbackEligible,
+  mapRowToDeployment,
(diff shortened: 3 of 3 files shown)


santhiprakash:
Closing — upstream main now types DeploymentMenu rows from the shared list type and gates rollback on deployment.isActive (see the vocabulary-drift comment in apps/dashboard/src/app/(dashboard)/deployments/components/DeploymentMenu.tsx), which is the fix this PR carried for #410.

Issue #391, opened 2026-08-02
[Bug] Relay SMTP settings not applied to webmail container + logout redirects to 404


farrasrayhand:
### Description

After configuring SMTP relay (Brevo/custom) in the Emails → Sending tab, the webmail container still uses the default `mail.example.com:465` instead of the configured relay settings.

### Steps to Reproduce

1. Install mail server via Openship
2. Go to Emails → Sending
3. Enable relay, set custom SMTP (e.g. Brevo `smtp-relay.brevo.com:587`)
4. Save changes — UI shows "Relay active" ✅
5. Go to Emails → Test → send test email

### Expected Behavior

Test email should be sent via the configured relay (`smtp-relay.brevo.com:587`).

### Actual Behavior

Error: `SMTP submission check failed against mail.example.com:465: Connection timeout`

Inspecting the webmail container environment variables confirms it still uses old defaults:

```
DEFAULT_SMTP_HOST=mail.example.com
DEFAULT_SMTP_PORT=465
```

### Root Cause (suspected)

The webmail container is created with hardcoded default SMTP env vars and does not get recreated or updated when relay settings are changed via the UI. The relay config appears to be saved (UI shows "Relay active") but is not propagated to the webmail container's environment.

### Workaround

Sending directly via `swaks` using Brevo credentials works fine, confirming the relay config itself is correct:

```bash
swaks --to user@gmail.com \
      --from postmaster@example.com \
      --server smtp-relay.brevo.com \
      --port 587 \
      --auth LOGIN \
      --auth-user YOUR_BREVO_LOGIN \
      --auth-password "YOUR_SMTP_KEY" \
      --tls
```

Output confirms successful delivery:

```
235 2.0.0 Authentication succeeded
250 2.0.0 OK: queued
221 2.0.0 Bye
```

---

### Description

After logging out from the webmail interface, the user is redirected to an invalid URL containing `undefined` instead of the correct login path, resulting in a 404 Page Not Found error.

### Steps to Reproduce

1. Log in to the webmail interface at `https://mail.example.com/mail`
2. Click logout
3. Observe the redirect URL

### Expected Behavior

User should be redirected to:

```
https://mail.example.com/mail/login
```

### Actual Behavior

User is redirected to:

```
https://mail.example.com/mail/undefined/login
```

This results in a 404 Page Not Found error.

### Workaround

Manually navigate to:

```
https://mail.example.com/mail
```

Or clear browser cookies and access the URL directly.

---

## Environment

[...]

PR #420, opened 2026-08-04
fix(email): resolve webmail logout redirect to /mail/undefined/login

santhiprakash:
Partially addresses #391 (Bug 2 only — logout redirect; SMTP relay is out of scope).

## Problem

After logging out of the self-hosted webmail UI, the post-logout session guard redirect used `${import.meta.env.VITE_PUBLIC_APP_URL}/login`. Self-hosted release builds omit that env var, so the template literal becomes the relative URL `undefined/login`. From `/mail/inbox`, the browser resolves that to `/mail/undefined/login` (404).

## Triage / Root cause

`clientLoader` auth guards in the email SPA (notably `apps/email/client/app/(routes)/mail/[folder]/page.tsx`) built redirect targets from build-time `VITE_PUBLIC_APP_URL`. The same class of bug was already fixed for API calls via `backend-url.ts`; redirects still used the stale env pattern.

Logout flow: `nav-user.tsx` → hard-navigate `/mail/inbox` → folder `clientLoader` sees no session → broken redirect.

## Fix

- Add `apps/email/client/lib/app-url.ts` with `getAppUrl()` / `absoluteAppUrl()` — runtime `window.location.origin` in the browser, optional `VITE_PUBLIC_APP_URL` fallback off-browser (mirrors `backend-url.ts`).
- Route all SPA `Response.redirect` call sites through `absoluteAppUrl()`.

## Verification

- `bun test apps/email/client/lib/app-url.test.ts` — passes; asserts logout/login redirect is `/login`, not `undefined/login`, when no origin env is set.
- Preflight: issue #391 open, no duplicate PRs, bug marker still on upstream.

## Notes / Risks

- Scoped to Bug 2 only per CONTRIBUTING one-change-per-PR.
- Also fixes the same latent bug in settings/mailto/create loaders that used the same env pattern.

### DIFF
--- a/apps/email/client/app/(routes)/mail/[folder]/page.tsx
+++ b/apps/email/client/app/(routes)/mail/[folder]/page.tsx
@@ -2,6 +2,7 @@ import { useLoaderData, useNavigate } from 'react-router';
 
 import { MailLayout } from '@/components/mail/mail';
 import { useLabels } from '@/hooks/use-labels';
+import { absoluteAppUrl } from '@/lib/app-url';
 import { authProxy } from '@/lib/auth-proxy';
 import { useEffect, useState } from 'react';
 import type { Route } from './+types/page';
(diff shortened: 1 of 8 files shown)


santhiprakash:
Closing — the mail client on main has been restructured and the logout-redirect surface this PR patched (apps/email/client/lib/app-url.ts) no longer exists in that form.

PR #421, opened 2026-08-04
fix(dashboard): surface analytics overview timeouts instead of empty stats

### LINKED ISSUE #396: Overview page shows 0 requests / no traffic data on some projects (silent AbortError in root layout)
## Summary

On a self-hosted instance (v0.5.0), the Project Overview page's summary cards ("Server Requests", "Unique IPs", "Avg Response", "Bandwidth Out") and the "Traffic" chart show `0` / `N/A` / "No traffic data yet" on **some** projects, while other projects on the same instance/session render correctly with real numbers.

This is **not a data problem** — the underlying analytics endpoint returns correct, non-zero data for the affected projects. It's a client-side rendering issue.

## Reproduction

Across 5 projects on the same instance/org/session:
- `mesh-api`, `eyris` → Overview page renders real traffic numbers correctly.
- 3 other projects → Overview page consistently shows `0`/`N/A`/"No traffic data yet", on every reload.

For an affected project, calling the exact same endpoint the page itself calls, from the page's own JS context (same session cookies, same fetch call the frontend issues):

```js
await fetch('/api/proxy/api/analytics/overview?projectId=<id>&domain=<domain>', {credentials:'include'})
  .then(r => r.json())
  .then(j => j.data.summary)
// => { totalRequests: 3057, uniqueVisitors: 3010, avgResponseTimeMs: 63970, ... }
```

...returns correct real data. So the API layer and the exact request the frontend makes both work fine. The page just doesn't render it.

## Console error (only fires on affected projects)

On every load of an affected project's Overview page, and never on a working one, the console throws:

```
AbortError: signal is aborted without reason
    at app/layout-<hash>.js:31:3830
```

This fires from the root `layout.js` chunk (shared across all routes), suggesting a shared data-fetching hook's `AbortController` is being aborted before the request completes — likely a race condition tied to some per-project state, since it's consistent per-project (always fires for the same broken projects, never for the working ones) rather than random.

## Impact

Users lose visibility into real traffic/usage for affected projects directly in the dashboard, even though the data is being collected and is available via the API. Currently working around it by querying `/api/analytics/overview` directly / using external monitoring instead of the dashboard widget.

## Environment

- Self-hosted, release `v0.5.0`
- Reproduces consistently across multiple browser sessions/reloads on the same affected projects

santhiprakash:
Fixes #396. Analytics failures now show a retry action instead of zero traffic. The error is confined to overview statistics and traffic, so Monitoring keeps independent resource and geography data visible. Both tabs use the same error component.

Overview derives the selected domain during rendering and waits for the current project/domain before fetching. This avoids an unnecessary all-domain request on initial load or navigation. The shared cache rejects responses from invalidated requests, including polling; the overview request allows 60 seconds for aggregation.

Validated with 99 relevant dashboard cases across request lifecycle, rendered Monitoring panels, and connection discovery (98-case suite plus the added polling regression). Ten new cases fail against main. Dashboard TypeScript check passes. Targeting the bug integration branch for PR #892.


Hydralerne:
MonitoringTab blanks the whole tab (should-fix)
The MonitoringTab change is a full early return ([MonitoringTab.tsx:309](vscode-webview://0b81592da317tc1nu9cumdtma4747grfifg2i5s05u5oc8qmuocn/apps/dashboard/src/app/(dashboard)/projects/%5Bid%5D/components/MonitoringTab.tsx#L309)) that replaces the entire MonitoringView with the error card when /analytics/overview fails. But that view renders several independent data sources:

geo map + top paths → useAnalyticsGeo (separate cache)
CPU/memory resource charts → useProjectUsageHistory (a DB read, not edge aggregation)
live traffic feed → useProjectUsageStream (SSE, already has its own usageError/reconnect)
So an overview timeout now hides the resource charts and live feed that loaded fine. That's a resilience regression: before this PR, an overview failure left analytics null and the TrafficChart showed its empty state, but geo/usage/live still rendered. The fix should scope the error to the traffic chart/summary block — exactly the way OverviewTab already scopes it — not nuke the tab.

santhiprakash:
Pushed a follow-up commit to address the feedback.

The early return in `MonitoringTab` is gone. Instead, `MonitoringView` now accepts an `analyticsError` prop and scopes the retry card to the stats strip / traffic chart block only. The resources card, visitor map, live hits, top paths, response mix, and usage history all keep rendering when `/analytics/overview` fails, which matches the `OverviewTab` pattern.

What changed:
- Removed the whole-tab `if (showAnalyticsError)` early return from `MonitoringTab`.
- Added `analyticsError` and `onRetryAnalytics` props to `MonitoringView`.
- When `analyticsError` is set and not loading, `MonitoringView` replaces the stats strip and `CollapsibleCard` traffic block with the existing error card + retry button.
- The no-data prompt and traffic `CollapsibleCard` are suppressed while the error is shown.

Local verification:
- `bun run --cwd apps/dashboard lint` (tsc --noEmit) passes.
- `bun run --cwd apps/dashboard test` passes (447 tests, including `MonitoringView.test.tsx`).
- `bunx prettier --write` applied to the two modified files.

Please take another look when you have a moment.

PR #422, opened 2026-08-04
fix(adapters): probe Docker daemon with docker version --format '{{.Server.Version}}'

### LINKED ISSUE #408: Docker health check shows Unhealthy on external server added via SSH (password auth) despite Docker running normally
When adding an external server via SSH (password auth) to a self-hosted Openship instance, the Components tab shows Docker as Unhealthy with the message "Docker is installed but the daemon is not running", even though Docker is fully functional and accessible on the target server.

**Environment**
- Openship: self-hosted (web dashboard)
- Target server: Ubuntu, Docker v29.7.1
- SSH Auth: Password, port 22
- Target user is in the `docker` group

**Steps to reproduce**
1. Set up a self-hosted Openship instance (web dashboard)
2. Add an external server via SSH (password auth)
3. Go to Components tab
4. Docker shows as Unhealthy: "Docker is installed but the daemon is not running"

**Evidence that Docker is working correctly**
From Openship's own Terminal tab on the same server:

\```
$ docker info 2>&1 | head -3
Client: Docker Engine - Community
 Version:    29.7.1
 Context:    default

$ docker ps
CONTAINER ID   IMAGE   COMMAND   CREATED   STATUS   PORTS   NAMES

$ groups $(whoami)
muha : muha adm cdrom sudo dip plugdev users lxd docker

$ ls -la /var/run/docker.sock
srw-rw---- 1 root docker 0 Aug 3 16:57 /var/run/docker.sock
\```

**Expected behavior**
Docker should show as Healthy since the daemon is running and accessible.

**Actual behavior**
Docker shows as Unhealthy despite being fully functional. The Terminal tab (which uses the same SSH connection) works fine, suggesting the issue is in the health check logic, not the SSH connection itself.

**Additional context**
- `docker info` works without sudo from Openship's own Terminal tab
- Socket permissions are correct (`root:docker`, `660`)
- User is in the `docker` group
- Clicking Re-check does not resolve the issue
- Clicking Reinstall installs Docker successfully (v29.7.1) but status remains Unhealthy
- This is an external server added to Openship, not the host server running Openship itself

santhiprakash:
## Summary

Fix Docker health check false negatives on SSH-added external servers running Docker 29 by replacing the `docker info --format '{{.ServerVersion}}'` probe with `docker version --format '{{.Server.Version}}'`.

## Motivation

Closes #408.

On SSH-added servers running Docker 29, the Components tab reported Docker as unhealthy ("daemon not running") even though `docker info` and `docker ps` worked from the interactive Terminal tab. The root cause is that Docker 29 no longer exposes the `.ServerVersion` template field in `docker info`, so the daemon probe exited with a template error (`template: no such field: .ServerVersion`).

## Related issue

Closes #408

## Changes

- **packages/adapters/src/system/catalog.ts** — change `daemonCommand` from `docker info --format '{{.ServerVersion}}'` to `docker version --format '{{.Server.Version}}'`, which is the documented, stable server-version field across Docker 24-29+.
- **packages/adapters/src/system/docker-check.test.ts** — update test needles and the daemon-command assertion to match the new probe and still verify that the probe targets a real server version.
- **packages/adapters/docs/SYSTEM.md** — update the setup-flow diagram to show the new `docker version` probe.

## Problem

Concrete repro from #408: add an external server over password-auth SSH where Docker 29.7.1 is installed and running. The Components tab health check calls `checkDocker()` -> `tryExec(executor, recipe.daemonCommand!)` with `docker info --format '{{.ServerVersion}}'`. On Docker 29 this produces `template: no such field: .ServerVersion`, so `tryExec()` returns an empty output and `checkDocker()` reports the daemon as not running.

## Triage / Root cause

Docker 29 removed the `.ServerVersion` field from the `docker info` Go-template context. `docker version --format '{{.Server.Version}}'` is the documented cross-version way to read the server version.

## Fix

Use `docker version --format '{{.Server.Version}}'`. It contacts the daemon and returns a positive server-version signal, so `checkDocker()` treats any non-empty output as "daemon running" and surfaces the underlying error on failure.

## Verification

```bash
cd packages/adapters && bun test src/system/docker-check.test.ts
# 7 pass, 0 fail
```

Also verified locally on Docker 29.1.3:

```bash
$ docker --version
Docker version 29.1.3, build 29.1.3-0ubuntu3~24.04.2
$ docker info --format '{{.ServerVersion}}'
docker: template: no such field: .ServerVersion
$ docker version --format '{{.Server.Version}}'
29.1.3
```

## Notes / Risks

- Scope is limited to the daemon probe string and its tests; no changes to `SshExecutor` or `checkDocker()` control flow.
- The old command still works on Docker <29, but `docker version --format '{{.Server.Version}}'` is available across Docker 24-29+.

## Checklist

### DIFF
--- a/packages/adapters/docs/SYSTEM.md
+++ b/packages/adapters/docs/SYSTEM.md
@@ -32,7 +32,7 @@ Dashboard Setup Wizard
 SystemManager.setup(onLog, config)
     │
     │  Phase 1: CHECK - run health checks
-    │     checkDocker(executor)  → "docker --version" + "docker info"
+    │     checkDocker(executor)  → "docker --version" + "docker version --format '{{.Server.Version}}'"
     │     checkGit(executor)     → "git --version"
     │     checkNode(executor)    → "node -v"
     │     checkTraefik(executor) → "traefik version" + systemctl status
--- a/packages/adapters/src/system/catalog.ts
+++ b/packages/adapters/src/system/catalog.ts
@@ -89,12 +89,11 @@ export const systemCatalog = {
   checks: {
     docker: {
       versionCommand: "docker --version",
-      // A POSITIVE signal, not a bare exit code: only a daemon that answered can
-      // name its version. `docker info >/dev/null && echo ok` prints ok on ANY
-      // exit-0, so a daemon answering nothing would read as healthy — trading #408's
-      // false negative for a false positive. Leave stderr alone too; `checkDocker`
-      // puts the failure text in the component message.
-      daemonCommand: "docker info --format '{{.ServerVersion}}'",
+      // Docker 29 removed the `.ServerVersion` template field from `docker info`,
+      // so `docker info --format '{{.ServerVersion}}'` fails with "no such field"
+      // on healthy Docker 29+ daemons. `docker version --format '{{.Server.Version}}'`
+      // is the documented, stable server-version probe across Docker 24-29+.
+      daemonCommand: "docker version --format '{{.Server.Version}}'",
       parseVersion: (output: string) =>
         output.match(/Docker version ([^\s,]+)/)?.[1] ?? output,
       missingMessage: "Docker is not installed",
--- a/packages/adapters/src/system/docker-check.test.ts
+++ b/packages/adapters/src/system/docker-check.test.ts
@@ -6,7 +6,7 @@ import type { CommandExecutor } from "../types";
 
 /**
  * #408: an SSH-added server reported "Docker is installed but the daemon is not
- * running" while `docker info` worked from the Terminal tab on the same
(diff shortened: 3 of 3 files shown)


Hydralerne:
Reviewed against main and the later reproduction in #408. Changing docker info to docker version does not refresh supplementary groups in an existing SSH login. Main already preserves the daemon error and distinguishes permission denied from a stopped daemon. The remaining stale-login bug is fixed in #892 by refreshing the pool only after observing changed account groups, preserving active sessions and rechecking once.

Closing this probe-command alternative because it does not address the confirmed failure. The fix was also verified against a real SSH login and group-protected probe.

PR #423, opened 2026-08-04
fix(mail): scope outbound-relay TLS to the relay transport

### LINKED ISSUE #392: [Bug] smtp_tls_security_level set to "encrypt" breaks inbound mail delivery via Amavis
## Description

Openship configures Postfix with `smtp_tls_security_level = encrypt` in `/etc/postfix/main.cf`. This forces TLS for **all** outbound SMTP connections including internal ones, which breaks delivery from Postfix to Amavis on localhost (127.0.0.1:10024).

## Error in /var/log/mail.log

```
postfix/amavis/smtp: to=<user@example.com>, relay=127.0.0.1[127.0.0.1]:10024, 
status=deferred (TLS is required, but was not offered by host 127.0.0.1[127.0.0.1])
```

## Root Cause

`smtp_tls_security_level = encrypt` enforces TLS for every outbound connection including loopback. Amavis on port 10024 does not offer TLS, so all inbound mail gets deferred indefinitely.

## Fix

Change in `/etc/postfix/main.cf`:

```
# Before (incorrect)
smtp_tls_security_level = encrypt

# After (correct)
smtp_tls_security_level = may
```

Then restart Postfix and flush the queue:

```bash
systemctl restart postfix
postqueue -f
```

## Impact

All inbound email is silently deferred — users receive no mail without any visible error in the UI.

## Environment

- Openship version: 0.5.0
- VPS: Oracle Cloud Free Tier (ARM64, Ubuntu 24.04 Noble)
- Architecture: arm64

santhiprakash:
## Summary

Stop setting `smtp_tls_security_level=encrypt` globally in `configureOutboundRelay`; that forces TLS on the local `smtp-amavis` delivery to `127.0.0.1:10024`, which does not speak TLS, causing all inbound mail to be silently deferred.

## Motivation

Reported in #392. `apps/api/src/modules/mail/admin/outbound-relay.service.ts` uses `postconf -e smtp_tls_security_level=encrypt` whenever an outbound relay is enabled. `smtp_tls_security_level` is a global Postfix client directive, so it also applies to the `smtp-amavis` transport and any other local SMTP client delivery. Amavis on `127.0.0.1:10024` does not offer STARTTLS, so Postfix logs `status=deferred (TLS is required, but was not offered by host 127.0.0.1...)` and inbound mail never arrives.

## Related issue

Fixes #392

## Changes

- **apps/api**
  - `outbound-relay.service.ts`:
    - Write a per-destination `/etc/postfix/tls_policy` map (`[<host>]:<port> encrypt`) and reference it via `smtp_tls_policy_maps`.
    - Change the global `smtp_tls_security_level` from `encrypt` to `may`.
    - For port 465, still add `smtp_tls_wrappermode=yes` globally, but pin the `smtp-amavis/unix` transport to `smtp_tls_wrappermode=no` via `postconf -P`.
    - Pin `smtp-amavis/unix` to `smtp_tls_security_level=none` via `postconf -P`.
    - On `disableOutboundRelay`, remove `smtp_tls_policy_maps` and `smtp_tls_wrappermode`, reset `smtp_tls_security_level=may`, and delete the `tls_policy` map files.
  - `outbound-relay.service.test.ts`: add/extend assertions for the `tls_policy` map, the `may` security level, the Amavis per-transport overrides, the port-465 wrapper-mode split, and disable cleanup.
- **apps/email/engine/samples/postfix/master.cf**: add `-o smtp_tls_security_level=none` and `-o smtp_tls_wrappermode=no` to the `smtp-amavis` transport so the sample config matches the runtime override.

## Verification

```bash
python3 /home/ubuntu/Projects/open-source/scripts/preflight_ship.py \
  --repo oblien/openship --local openship --branch main --issue 392 \
  --file apps/api/src/modules/mail/admin/outbound-relay.service.ts \
  --must-contain 'smtp_tls_security_level=encrypt' \
  --must-not-contain 'smtp_tls_policy_maps'
# PREFLIGHT CLEAR

bun run --cwd apps/api test outbound-relay.service.test.ts
# ✓ src/modules/mail/admin/outbound-relay.service.test.ts (18 tests)

bun run --cwd apps/api lint
# tsc --noEmit (pass)
```

The targeted test includes new assertions that fail against the original code (which emits `smtp_tls_security_level=encrypt` and no `smtp_tls_policy_maps` / no Amavis override).

## Screenshots

N/A — backend Postfix config change only.

## Checklist

Note on the unchecked box: `bunx prettier --check` on the two touched `apps/api` files reports pre-existing formatting issues (long lines) unrelated to this change; running `bun format` would reformat many lines I didn't touch, so I scoped the diff to the functional change. `bun run --cwd apps/api lint` and the targeted test pass.

### DIFF
--- a/apps/api/src/modules/mail/admin/outbound-relay.service.test.ts
+++ b/apps/api/src/modules/mail/admin/outbound-relay.service.test.ts
@@ -64,12 +64,14 @@ beforeEach(() => {
 describe("configureOutboundRelay", () => {
   const base = { provider: "ses" as const, region: "us-east-1", port: 587, username: "AKIASMTPUSER", password: "s3cr3tPass" };
 
-  test("writes creds via SFTP writeFile, NEVER through a shell command", async () => {
+  test("writes creds + per-destination TLS policy via SFTP, NEVER through a shell command", async () => {
     const { exec, execCalls, writes } = makeExec();
     await configureOutboundRelay(exec, base);
 
     expect(writes[0].path).toBe("/etc/postfix/sasl_passwd");
     expect(writes[0].content).toContain("[email-smtp.us-east-1.amazonaws.com]:587 AKIASMTPUSER:s3cr3tPass");
+    expect(writes[1].path).toBe("/etc/postfix/tls_policy");
+    expect(writes[1].content).toContain("[email-smtp.us-east-1.amazonaws.com]:587 encrypt");
 
     // SECURITY INVARIANT: no shell command may contain the password or username.
     for (const cmd of execCalls) {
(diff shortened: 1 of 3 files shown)


santhiprakash:
Closing — upstream main now scopes relay TLS via per-nexthop smtp_tls_policy_maps and document-transport overrides in apps/api/src/modules/mail/admin/outbound-relay.service.ts (see the TLS design comments there), which is the fix this PR proposed for #392.

PR #425, opened 2026-08-04
fix(adapters): use POSIX path module for remote nginx/certbot paths

### LINKED ISSUE #381: Openship Windows Path Bug
# Bug report: Windows path separators used for remote Linux paths

**Impact: certificates are issued successfully but never recorded, so renewal is
never scheduled. TLS silently expires ~90 days later.**

## Environment

| | |
|---|---|
| Openship control plane | Windows |
| Target server | Ubuntu 24.04, deployed over SSH |
| Project | `framework: docker-compose`, `productionMode: host` |
| Domain | custom, `externalIngress: false`, `manualSsl: false` |
| Cert flow | certbot, standalone HTTP-01 on port 49180, proxied via the edge |

## What happens

Adding a custom domain gets all the way to a valid certificate and then fails on
the last step. certbot's own output, surfaced in the Openship UI:

```
Requesting a certificate for app.example.com (standalone HTTP-01 via the edge)…
Successfully received certificate.
Certificate is saved at: /etc/letsencrypt/live/app.example.com/fullchain.pem
Key is saved at:         /etc/letsencrypt/live/app.example.com/privkey.pem
This certificate expires on <date>.
```

Immediately followed by:

```
certbot exited without an error but no readable certificate is at
\etc\letsencrypt\live\app.example.com. certbot said:
...
```

Note the separators. certbot wrote to `/etc/letsencrypt/live/...`; Openship
looked for `\etc\letsencrypt\live\...` on a Linux host and found nothing.

The certificate is genuinely there and valid — the site serves it once a vhost
exists:

```
subject=CN=app.example.com
issuer=C=US, O=Let's Encrypt, CN=<ca>
notBefore=<issue date>
notAfter=<issue date + 90d>
```

## Why this matters more than the failed step

The domain record is left in a state that suggests nothing was ever issued:

```
hostname       app.example.com
status         pending
sslStatus      provisioning
sslExpiresAt   null
verified       false
sslIssuer      certbot
```

`sslExpiresAt: null` is the dangerous part. If renewal scheduling keys off the
recorded expiry or off `sslStatus == issued`, it will never fire — while a
perfectly valid certificate sits on disk quietly counting down. Nothing else on
the host covers it: there is no certbot binary, no certbot container, no systemd
timer and no cron entry. `/etc/letsencrypt/renewal/*.conf` exists and is correct,
so any `certbot renew` would work; nothing runs one.

So on a Windows control plane the failure mode is not "domain setup failed" —
it's "domain setup appeared to half-fail, works fine, and loses TLS in 90 days".

## Second instance of the same bug

The generated nginx vhost has the same problem. Rather than landing in
`/var/lib/openship/edge/sites-enabled/`, it is written to the current directory
as a file literally named:

```
\var\lib\openship\edge\sites-enabled\<site>.conf
```

nginx never sees it, so the site does not serve until a vhost is installed by
hand. Same root cause, different path.

## Suggested fix

1. Build remote paths with POSIX joining regardless of control-plane OS. Any
   path destined for the target server should never pass through the host's
   native path module.
2. Do not infer failure from certbot's stdout. The run above exited 0 and the
   text reported as `lastVerifyError` was certbot's donation footer:
   `* Donating to ISRG / Let's Encrypt: https://letsencrypt.org/donate`.
   Exit code plus a stat of the (correctly-joined) live directory is decisive.
3. Consider recording `sslExpiresAt` by reading the certificate itself, so a
   post-issue bookkeeping failure cannot silently disable renewal.

## What worked

Worth saying, because most of the flow is sound: DNS verification, the ACME
challenge proxied through the edge on port 49180, certbot itself, and the
certificate and renewal config it produced. Only the path handling after
issuance failed.

santhiprakash:
## Summary

Windows control planes were building remote Linux paths with the OS-dependent `node:path` module, producing backslash-separated paths. This caused certbot-issued certs to be reported missing (`sslExpiresAt: null`, breaking renewal scheduling) and nginx vhosts to be written to a literal backslash-named file in the CWD.

## Motivation

On a Windows control plane, `join('/etc/letsencrypt/live', 'app.example.com')` becomes `\\etc\\letsencrypt\\live\\app.example.com`. The remote SSH/SFTP executor then looks for a file at that backslash path, which fails, even though certbot correctly wrote the cert to the POSIX path. The same thing happens for vhost paths under `sitesDir`.

## Related issue

Fixes #381

## Changes

- `packages/adapters/src/infra/nginx.ts`: import `dirname` and `join` from `node:path/posix` so every path constructed for the remote target uses forward slashes, regardless of the control plane OS.
- `packages/adapters/src/infra/nginx-posix-paths.test.ts`: add a regression test that mocks `node:path` to `win32` and asserts the generated vhost path stays POSIX.

## Verification

- `bun run --cwd packages/adapters lint` — clean (`tsc --noEmit`)
- `bun run --cwd packages/adapters test` — 757 passed, including the new regression test
- Manually confirmed the new test fails when `nginx.ts` is reverted to `node:path` (path becomes `\\var\\lib\\openship\\edge\\sites-enabled\\app-example-com.conf`) and passes with `node:path/posix`.

```bash
cd packages/adapters
bun run lint
bun run test
```

## Checklist

- [x] `bun run test`, `bun run --cwd <workspace> lint` pass locally

### DIFF
--- a/packages/adapters/src/infra/nginx-posix-paths.test.ts
+++ b/packages/adapters/src/infra/nginx-posix-paths.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, test, vi } from "vitest";
+import { NginxProvider, type NginxProviderOptions } from "./nginx";
+import { OPENRESTY_DEFAULT_PATHS } from "./openresty-lua";
+import type { CommandExecutor, RouteConfig } from "../types";
+
+// Windows control planes must still produce remote (target Linux) paths with
+// forward slashes. The source already switched to `node:path/posix`; this test
+// makes sure a `node:path` default of `win32` cannot reintroduce backslashes.
+const win32Path = vi.hoisted(() => {
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const path = require("node:path/win32") as typeof import("node:path");
+  return { ...path, default: path };
+});
+
+vi.mock("node:path", () => win32Path);
+
+const SITES = "/var/lib/openship/edge/sites-enabled";
+const CERT = "/etc/letsencrypt/live";
+
+interface FakeExecutor extends CommandExecutor {
+  writes: Array<{ path: string; content: string }>;
+}
+
+function makeExecutor(): FakeExecutor {
+  const files = new Map<string, string>();
+  const writes: Array<{ path: string; content: string }> = [];
+
+  function recordWrite(p: string, c: string) {
+    writes.push({ path: p, content: c });
+    files.set(p, c);
+  }
+
+  const executor = {
+    writes,
+    exec: async (command: string) => {
+      const singleMv = command.match(/^mv '([^']+)' '([^']+)'$/);
+      if (singleMv) {
+        const [, src, dest] = singleMv;
+        const c = files.get(src);
+        if (c !== undefined) {
+          files.set(dest, c);
+          recordWrite(dest, c);
+          files.delete(src);
+        }
+        return "";
+      }
+      if (command.startsWith("mv -f ")) {
+        const paths = [...command.matchAll(/'([^']+)'/g)].map((m) => m[1]!);
+        const dir = paths.pop()!;
+        for (const src of paths) {
+          const c = files.get(src);
+          if (c !== undefined) {
+            const dest = `${dir.replace(/\/$/, "")}/${src.split(/[\\/]/).pop()!}`;
+            files.set(dest, c);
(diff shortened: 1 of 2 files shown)


santhiprakash:
Closing as superseded — the same POSIX-path fix is already in `upstream/main` via commit `5230af32` (uses `node:path.posix` for remote path construction).

PR #435, opened 2026-08-04
fix(backups): fail fast and allow cancel when backup restore-apply hangs with no managed container

### LINKED ISSUE #434: Backup restore-apply hangs forever in "applying" when the service row has no managed container (v0.5.0); cancel refused mid-apply
### Environment
- Openship **v0.5.0** (installed via `openship update` from v0.3.0), self-hosted CLI service (systemd) on Debian, Docker runtime
- Deploy target: self-hosted SSH server = the same box (`127.0.0.1`)

### Setup that triggers it
A single-app **git-deploy (docker) project** originally created on v0.3.0 — its running container is deployment-managed (`openship-<slug>-dep_*`). Because backup policies fan out over *service rows* only, we retrofitted one:

1. `openship services create hello --project <id> --build . --dockerfile Dockerfile --port 20000:3000`
2. `PATCH /projects/:pid/services/:sid` with `{"volumes":["trialdata:/data"]}`

So the service row exists and is enabled, but it has **no service-managed container** — the app's container belongs to the deployment.

### What works
- Backup policy create + `openship backup policy run <policyId>` → run **succeeded**, artifact `volume-<name>-0.tar.zst` + manifest written to the local destination. The artifact is valid (manual `zstd -dc | tar -x` reproduces the exact volume contents).
- `openship backup run restore <runId>` (in_place) → restore reaches `prepared` with a confirmation token.

### Bug
`openship backup restore apply <restoreId> --token <token>` → the restore transitions to **`applying` and never progresses** (observed >4 minutes; `bytesRestored` already set from prepare).

- Journal shows the apply POST return 200 and then nothing further from the restore worker.
- `docker events` shows **no stop attempt on any container** during the whole window.
- Reading `apps/api/src/modules/backups/restore.orchestrator.ts`, apply = `transition("applying")` → `executor.stopService(serviceHandle)` → wipe+extract → `startService`. `stopService` appears to target the *service-managed* container (`openship-<project>-<service>`), which does not exist in this setup — the call never completes and never times out.
- `openship backup restore cancel <id>` → **"Cannot cancel a applying restore"**, so there is no user-side recovery.
- Only way out: `systemctl restart openship` — the runtime backstop then marks the restore `server_error: "API restart while restore in flight"`. (The app container was unaffected throughout.)

### Expected
1. Fail fast with a clear error when the target service has no managed container ("service has no container to stop/restore into"), or a timeout on `stopService`.
2. Allow cancelling a restore that is stuck in `applying`.

### Notes
- Volume backup itself handled this hybrid shape fine; only restore-apply assumes a service-managed container.
- Same reporter as #203 (the v0.3.0 SSH-deploy hang) — this one is unrelated to that fix, which works great in v0.5.0.

santhiprakash:
Fixes #434.

## Summary

- Fail fast in `runApply()` before calling `stopService()` when the target docker/cloud service has no live managed `containerId`. This prevents restore-apply from hanging in `applying` for hybrid deployments where a service row exists for backup policy fan-out but the app container is deployment-managed (`openship-<slug>-dep_*`) and not a service-managed restore target.
- Wrap `stopService()` and `startService()` with a 60-second client-side timeout so a stuck Docker/SSH call fails the restore instead of leaving it frozen in `applying`.
- Extend `cancel()` to accept `applying` restores, and schedule a best-effort `startService()` afterwards so a cancelled mid-apply restore attempts to bring the service back up.
- Add a regression test covering the no-managed-container fail-fast path, normal apply, bare runtimes, mail-server restores, and mid-apply cancellation.

## Test plan

- `bunx vitest run src/modules/backups/restore.orchestrator.test.ts` — 7/7 passing.
- `bun run lint` (`tsc --noEmit`) in `apps/api` passes.
- `npx prettier --check apps/api/src/modules/backups/restore.orchestrator.ts apps/api/src/modules/backups/restore.orchestrator.test.ts` passes.

### DIFF
--- a/apps/api/src/modules/backups/restore.orchestrator.test.ts
+++ b/apps/api/src/modules/backups/restore.orchestrator.test.ts
@@ -1,0 +1,253 @@
+import { describe, it, expect, beforeEach, vi } from "vitest";
+
+const h = vi.hoisted(() => {
+  const transitions: Array<[string, string, Record<string, unknown> | undefined]> = [];
+  const fakeExecutor: {
+    runtimeName: "docker" | "bare" | "cloud";
+    stopService: ReturnType<typeof vi.fn>;
+    startService: ReturnType<typeof vi.fn>;
+    isRunning: ReturnType<typeof vi.fn>;
+  } = {
+    runtimeName: "docker",
+    stopService: vi.fn(async () => {}),
+    startService: vi.fn(async () => {}),
+    isRunning: vi.fn(async () => false),
+  };
+  const fakeDestination = { get: vi.fn(), head: vi.fn() };
+
+  return {
+    transitions,
+    fakeExecutor,
+    fakeDestination,
+    buildApplyTargetResult: null as {
+      executor: unknown;
+      serviceHandle: Record<string, unknown>;
+    } | null,
+    currentRestore: null as Record<string, unknown> | null,
+    currentRun: null as Record<string, unknown> | null,
+    currentDestination: null as Record<string, unknown> | null,
+    repos: {
+      backupRestore: {
+        findById: vi.fn(async (_id: string) => h.currentRestore),
+        transition: vi.fn(async (id: string, status: string, patch?: Record<string, unknown>) => {
+          transitions.push([id, status, patch]);
+          if (h.currentRestore && h.currentRestore.id === id) {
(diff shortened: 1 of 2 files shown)


Hydralerne:
thank you but already fixed in main

PR #443, opened 2026-08-05
fix(api): accept framework display labels in project ensure schema to…

### LINKED ISSUE #427: [Bug] Cannot create project — POST /api/projects/ensure returns 400 Bad Request
## Bug Description

Creating a new project from the dashboard fails immediately with a 400 Bad Request
error on `POST /api/proxy/api/projects/ensure`. The request is retried automatically
by the frontend but all attempts fail. The dashboard also throws a React hydration
error (#418) alongside this.

## Steps to Reproduce

1. Open the Openship dashboard
2. Navigate to Projects
3. Attempt to create a new project
4. Observe the error in the UI and browser console

## Expected Behavior

Project is created successfully and appears in the dashboard.

## Actual Behavior

Console errors:

```
/api/proxy/api/projects/ensure — 400 Bad Request (repeated 3x)
Deployment error: ApiError: API 400: Bad Request
Uncaught Error: Minified React error #418 (text hydration mismatch)
```

The project creation flow fails completely. No project is created.

## Browser Console Output

```
content-script.js:657 [GD Media] attribution started in frame https://ops.example.com
Uncaught Error: Minified React error #418; visit https://react.dev/errors/418?args[]=text&args[]=
/api/proxy/api/projects/ensure:1 Failed to load resource: the server responded with a status of 400 (Bad Request)
Deployment error: ApiError: API 400: Bad Request
    at H (9241-998ea2829f58eae7.js:1:20147)
    at async 6900-bf288b9bfaf6b303.js:1:52105
    at async page-ca5b570dccb86577.js:1:81496
```

## Environment

- Deployment: Self-hosted (Docker Compose)
- Dashboard URL: https://ops.example.com
- Server: root@127.0.0.1 (localhost), Port 22, Password auth, Status: Healthy

## Additional Notes

The server shows as **Healthy** in the dashboard. The issue is specifically
with project creation, not server connectivity. No additional error detail
is surfaced in the UI — the 400 response body (if any) is not shown to the user,
making this difficult to debug without access to API logs.

## Workaround

None found so far.

farrasrayhand:
---

## Summary

Fix 400 Bad Request on `POST /api/projects/ensure` caused by `tbValidator` rejecting framework display labels (e.g. `"Static Site"`) that don't match any entry in `STACK_IDS`.

## Motivation

Since v0.5.0, the `/ensure` route gained a `body: EnsureProjectBody` declaration as part of MCP tool wiring. This caused `secureRouter` to auto-mount `tbValidator("json", EnsureProjectBody)` before the controller runs. `FrameworkEnum` is a strict TypeBox union of `STACK_IDS` — but the dashboard sends human-readable display labels like `"Static Site"` instead of stack IDs like `"static"`. TypeBox rejects these immediately with 400, before any business logic is reached.

In v0.4.8 the route had no `body` declaration so `tbValidator` was never mounted and the value passed through to the controller, which ignored unrecognized framework strings anyway.

## Related issue

Closes #427

## Changes

**`apps/api/src/modules/projects/project.schema.ts`**
- Relaxed `framework` field in `CreateProjectBody` from strict `FrameworkEnum` to `Type.Optional(Type.String({ maxLength: 100 }))` — validation/normalization is handled at the service layer, not the schema boundary

## Verification

Before (v0.5.0):
```bash
POST /api/projects/ensure 400 8ms
# API log confirms rejection at middleware level — too fast to have touched controller or DB
```

After this fix:
```bash
POST /api/projects/ensure 200 42ms
# Request passes through, project created successfully
```

Reproduced with [`[farrasrayhand/script-collection](https://github.com/farrasrayhand/script-collection)`](https://github.com/farrasrayhand/script-collection) — plain static site, auto-detected as "Static Site" by the dashboard, which triggered the 400 every time.

## Checklist

- [ ] A test fails without this change and passes with it (or I explained above why there isn't one) — no existing test covered this path; the schema change is self-evident from the TypeBox rejection

### DIFF
--- a/apps/api/src/modules/projects/project-crud.service.ts
+++ b/apps/api/src/modules/projects/project-crud.service.ts
@@ -18,6 +18,7 @@ import {
   normalizeRollbackWindow,
   normalizeAliasStrict,
   aliasConflictsWithSiblings,
+  normalizeFramework,
   type ReleaseSource,
   type UpdatableIdentity,
 } from "@repo/core";
@@ -357,7 +358,7 @@ function buildProductionProjectInput(
     releaseSource: source.releaseSource,
     installationId: data.installationId,
     autoDeploy: !!(env.CLOUD_MODE && source.gitOwner && source.gitRepo),
-    framework: data.framework ?? "unknown",
+    framework: normalizeFramework(data.framework),
     packageManager: data.packageManager ?? "npm",
     installCommand: data.installCommand,
     buildCommand: data.buildCommand,
@@ -890,7 +891,7 @@ export async function ensureProject(
       throw new NotFoundError("Project", data.projectId ?? desiredSlug);
     }
     const update: Record<string, unknown> = {};
-    if (data.framework !== undefined) update.framework = data.framework;
+    if (data.framework !== undefined) update.framework = normalizeFramework(data.framework);
     if (data.packageManager !== undefined) update.packageManager = data.packageManager;
     if (data.installCommand !== undefined) update.installCommand = data.installCommand;
     if (data.buildCommand !== undefined) update.buildCommand = data.buildCommand;
--- a/apps/api/src/modules/projects/project.schema.ts
+++ b/apps/api/src/modules/projects/project.schema.ts
@@ -17,9 +17,10 @@ import {
 
 // ─── Shared enums (derived from registry) ────────────────────────────────────
 
-export const FrameworkEnum = Type.Union(
-  STACK_IDS.map((id) => Type.Literal(id)) as [TLiteral<string>, ...TLiteral<string>[]],
-);
+export const FrameworkEnum = Type.String({
+  maxLength: 100,
+  description: "Framework stack ID or display label (e.g. 'static', 'nextjs', 'Static Site').",
+});
 
 export const PackageManagerEnum = Type.Union(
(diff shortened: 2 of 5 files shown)


Hydralerne:
thank you

Hydralerne:
Nice fix for #427 — the approach is right, but relaxing the shared FrameworkEnum while only normalizing 2 of its write paths leaves the same 400-class gap open on the others. Three things before merge:

PATCH /projects/:id — updateProject stores framework raw (PROJECT_UPDATE_KEYS loop, project-crud.service.ts:1106-1114). UpdateProjectBody inherits the relaxed enum, so "Static Site" now persists verbatim. MCP-exposed.

Service / monorepo sub-app writes — same shared enum feeds the service create/update bodies; service.service.ts:488 and project-crud.service.ts:436 store raw. A sub-app labeled "Static Site" fails isStaticService's framework in STACKS guard → deployed as a running server instead of static.

normalizeFramework: lower in STACKS → Object.hasOwn(STACKS, lower) — in walks the prototype chain, so framework:"constructor" / "__proto__" pass and are stored verbatim, then getRuntimeImage("constructor") throws a TypeError in the build path.

Cleanest fix for 1 & 2: normalize at the schema boundary (a Transform on FrameworkEnum) instead of at hand-picked call sites — then every consumer of the shared enum is covered and a new route can't reopen it. Dashboard callers are unaffected either way; this only bites raw API/MCP clients.

PR #451, opened 2026-08-05
fix(adapters): folder deploy takes down the API; SSH channel drops mid-build

### LINKED ISSUE #448: [Bug] Folder deploy crashes the entire control plane: ENOENT lstat race in docker build context tar (Windows, pnpm monorepo)
## Summary
Deploying a local folder (large pnpm/turbo monorepo) on Windows crashes the **whole API process** — not just the build. The context packer hits a deterministic `ENOENT` on the same file every time, the tar-fs Pack emits an unhandled 'error' event, and the control plane exits.

## Environment
- Openship CLI **0.5.0** (npm install), bare mode via `openship up --foreground`, Node v24.11.1
- Windows 11 Pro, repo: pnpm workspaces + turbo monorepo (~85k files w/o node_modules), Dockerfile at repo root
- Project: `type docker`, local-folder source

## Repro
1. `openship project create --type docker --local-path <monorepo>` + `openship init`
2. `openship deploy`
3. Build starts → ~seconds later the API process dies. Reproduced twice back-to-back, same file both times.

## Log
```
[ioredis] Unhandled error event: AggregateError [ECONNREFUSED]
node:events:486
      throw er; // Unhandled 'error' event
Error: ENOENT: no such file or directory, lstat 'C:\Users\<u>\AppData\Local\Temp\openship-docker-context-AEUppo\.mcp.json.example'
Emitted 'error' event on Pack instance at:
    at Pack.destroy (...\node_modules\tar-stream\pack.js:202:17)
    at onstat (...\node_modules\tar-fs\index.js:101:26)
```
(second run: identical, temp dir `openship-docker-context-6Xqs2t`, same `.mcp.json.example`)

Notes:
- `.mcp.json.example` is a plain 345-byte tracked JSON file (not a symlink). `.mcp.json` (sibling) is gitignored; the file is NOT in `.dockerignore`. Looks like a race between the ignore-copy into the temp context dir and the tar walk (file listed by the walker, gone at lstat).
- Tested on a build that includes the v0.5.0 `.dockerignore` semantics fix (#370) — still crashes.

## Expected
1. A build-context error should fail the deployment, **never take down the API** (unhandled 'error' on the Pack stream).
2. The copy/tar pipeline shouldn't race on ignored-sibling patterns.

Happy to provide more logs or test a patch — this repo reproduces it 100%.

---

**Update:** root cause found and fixed in #451 — `applyDockerignore`'s parallel prune races the directory listing, and dockerode's tar-fs pack then `lstat`s a name that is already gone, emitting an unhandled `'error'` that takes the process down. Verified on the same repo: the control plane now survives the whole build.

cengalishen:
Two independent failures we hit deploying a local folder (large pnpm/turbo monorepo, Windows orchestrator → Ubuntu server). Both are reproducible and both are fixed here; each commit stands alone.

Fixes #448.

## 1. A missing context file kills the whole API

`prepareSourceTree` stages the repo, then `applyDockerignore` prunes it with `Promise.all` — a wide `.dockerignore` (every `*.md`, `.github/`, `.gitignore` …) fans out to hundreds of concurrent `rm`s. The directory listing can lag behind those deletes, so `readdir` still reports a name whose inode is gone. dockerode's tar-fs pack then `lstat`s it, emits an **unhandled** `'error'`, and the process exits:

```
Error: ENOENT: no such file or directory, lstat 'C:\…\openship-docker-context-AEUppo\.mcp.json.example'
    at Pack.destroy (tar-stream/pack.js:202:17)
    at onstat (tar-fs/index.js:101:26)
```

Not a failed build — the control plane goes down with it. Two narrow changes: prune sequentially so the tree and its listing stay in step, and verify each name still resolves before the list is handed to dockerode. A test pins the invariant (every returned entry exists after a pruning `.dockerignore`).

## 2. `remote channel closed without an exit status`

With #1 fixed the build ran, then died in `COPY --from=deps` and again during `next build`. `keepaliveCountMax: 10` (~150s) is short for a real build: a large `node_modules` copy or a Next.js compile pins disk/CPU for minutes with no channel traffic, sshd misses ten consecutive keepalive replies, and the client tears down a link that is only busy. Raised to 240 (~60m); the exec paths keep their own idle timeouts so liveness is still bounded.

## One more, not fixed here

After both fixes the image built successfully, but a build longer than **5 minutes** still lost its connection until I raised `DEFAULTS.idleTimeoutMs` in `apps/api/src/lib/ssh-manager.ts` locally. `touchIdleTimer` skips the timer while `retainCounts > 0`, so the mechanism exists — it looks like the server-build exec simply isn't retaining the connection for its duration. Bumping the default is the wrong fix, so I've left it out; you'll know the right place to `retain`/`release`. Happy to send it as a follow-up if you point me at it.

## Verification

Real deploy of a ~2000-package pnpm/turbo monorepo to a self-hosted Ubuntu server:

| | before | after |
|---|---|---|
| control plane during folder deploy | died within seconds, every attempt | stayed up through a 6-minute build |
| build | never reached Docker | image built and finalized |
| a failing build | took the API with it | reported as a build failure |

Also ran the isolated race repro (400 files + a wide `.dockerignore`): the old parallel prune leaves stale names in the listing, the new path doesn't.

### DIFF
--- a/packages/adapters/src/runtime/docker-build-context.test.ts
+++ b/packages/adapters/src/runtime/docker-build-context.test.ts
@@ -151,4 +151,31 @@ describe("createDockerBuildContext", () => {
     expect(context.dockerfileName).toBe("Dockerfile");
     expect(context.usesRepositoryDockerfile).toBe(true);
   });
+
+  /**
+   * dockerode's tar-fs pack `lstat`s every name in `contextEntries` and emits an
+   * unhandled 'error' event on a miss — which kills the API process, not just the
+   * build. A wide `.dockerignore` prunes many paths right before this list is
+   * read, so every returned name must still resolve.
+   */
+  it("only lists context entries that still exist after pruning", async () => {
+    const repo = await makeRepo("*.md\n.gitignore\ndocs\n");
+    await writeFile(join(repo, "README.md"), "# readme\n");
+    await writeFile(join(repo, "NOTES.md"), "# notes\n");
+    await writeFile(join(repo, ".gitignore"), "nothing\n");
+    await mkdir(join(repo, "docs"), { recursive: true });
+    await writeFile(join(repo, "docs", "guide.md"), "# guide\n");
+    await exec("git", ["-C", repo, "add", "-A", "-f"]);
+    await exec("git", ["-C", repo, "commit", "-q", "-m", "pruned files"]);
+
+    const context = await createDockerBuildContext(config(repo), {
+      requireRepositoryDockerfile: true,
+    });
+    cleanups.push(context.cleanup);
+
+    const onDisk = new Set(await readdir(context.contextDir));
+    for (const entry of context.contextEntries) expect(onDisk.has(entry)).toBe(true);
+    expect(context.contextEntries).not.toContain("NOTES.md");
+    expect(context.contextEntries).not.toContain("docs");
+  });
 });
--- a/packages/adapters/src/runtime/docker-build-context.ts
+++ b/packages/adapters/src/runtime/docker-build-context.ts
@@ -180,17 +180,21 @@ async function applyDockerignore(contextDir: string, config: BuildConfig): Promi
 
   const prune = async (currentPath: string): Promise<void> => {
     const entries = await readdir(currentPath, { withFileTypes: true });
-    await Promise.all(
-      entries.map(async (entry) => {
-        const absolutePath = join(currentPath, entry.name);
-        const relativePath = toPosixPath(relative(contextDir, absolutePath));
-        if (!buildFiles.has(relativePath) && matcher.ignores(relativePath)) {
-          await rm(absolutePath, { recursive: true, force: true });
-          return;
-        }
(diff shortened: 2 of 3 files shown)


cengalishen:
Thanks for closing #448 — just flagging that `main` still has the parallel prune, so I don't think the crash is fixed yet:

https://github.com/oblien/openship/blob/main/packages/adapters/src/runtime/docker-build-context.ts#L183

```js
await Promise.all(
  entries.map(async (entry) => {
    ...
    await rm(absolutePath, { recursive: true, force: true });
```

That's the window: the fan-out races the directory listing, `readdir` then names an entry whose inode is gone, and dockerode's tar-fs pack `lstat`s it and emits an unhandled `'error'` — which exits the process rather than failing the build. Last commit touching that file is from Jul 31.

This PR closes it (sequential prune + verify each entry before handing the list to dockerode, with a test), and also fixes the SSH channel drop that surfaced right after. No rush — happy to rebase or split it if that's easier to review.

PR #470, opened 2026-08-06
feat(core): pin the Ruby image to the version the project declares

a-abdellatif98:
> **Stacked on #468** — it branches from that work, so its two commits show here until #468 merges. Opened as a draft; review the top commit only, or wait for #468. Happy to rebase once that lands.

## Problem

`LANGUAGES.ruby` hardcodes `ruby:3.3-slim`. Nothing reads `.ruby-version`, the lockfile's `RUBY VERSION` stanza, or the Gemfile's `ruby` directive, so any app pinning a different Ruby fails bundler's version check before it installs a single gem. The pin is the one piece of the toolchain the project genuinely gets to choose.

## Changes

`getBuildImage` / `getRuntimeImage` take an optional version and swap the version segment of a `ruby:<version><suffix>` tag, preserving the variant. Both new parameters are optional, so every existing call site is untouched. `stack-detector.ts` resolves the version in the order bundler trusts: `.ruby-version`, then the lockfile stanza, then the Gemfile directive.

**The part that needed the most care.** Build and runtime images have to move together — bundler installs into a version-scoped path under `BUNDLE_PATH`, so a runtime on a different Ruby finds no gems at all. But the two are resolved in different places: `buildImage` is persisted on the project from detection, while `runtimeImage` is recomputed at deploy time from the DB row (`build.service.ts:598` `resolveRuntimeImage`). Pinning only the build image would have produced exactly that mismatch.

Rather than add a column and a migration, `resolveRuntimeImage` now derives the tag from the already-persisted `project.buildImage` via a new `runtimeVersionFromImage` helper. The two are in lockstep by construction, and there is no schema change.

**Security note:** the version originates in a file inside the repo being deployed, so it is attacker-controlled input to an image reference. It is validated against `^\d+\.\d+(?:\.\d+)?$` before it reaches a tag; anything else (`latest`, `3.3-slim && rm -rf /`, a traversal) falls back to the language default. There is a test covering that.

`.ruby-version` is added to `rubyLanguageDetector.manifestFiles` so callers fetch it; `parseManifest` returns `{}` for it, so it contributes no phantom deps. It has no effect on project-root detection — roots come from `STACK_ROOT_MARKERS`, and the `manifestFiles` use at `project-root-detector.ts:271` is against `WORKSPACE_DETECTORS`, a separate registry.

## Testing

10 new tests, including the round-trip that keeps build and runtime aligned and the tag-injection rejection cases. Full `packages/core` suite: **437 passing, 0 failures**. `tsc --noEmit` clean in `packages/core` and `apps/api`.

### DIFF
--- a/apps/api/src/lib/stack-detector.ts
+++ b/apps/api/src/lib/stack-detector.ts
@@ -26,6 +26,7 @@ import {
   OUTPUT_DIRECTORIES,
   getProjectType,
   getBuildImage,
+  parseRubyVersion,
   LANGUAGE_MANIFEST_FILES,
   collectDependencies,
   detectPort as detectPortFromLanguages,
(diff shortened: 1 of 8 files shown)


a-abdellatif98:
@Hydralerne #468 and #469 landed in #892, but this one isn't mentioned there. Is the Ruby version pin still wanted, or should it be reworked against the integration branch?

PR #472, opened 2026-08-06
Restore Star History section in README

### LINKED ISSUE #260: Star history chart is not displaying
The **Star History** section in the README does not display a chart. It appears as an empty/broken image on GitHub.

Could you please check the embedded Star History image/link?

Mubelotix:
Fixes #260

## Verification

```bash

```

## Checklist

- [ ] One change per PR — one bug, or one agreed feature, with nothing unrelated bundled in
- [ ] The diff is scoped — no reformatting or lint fixes on lines I wasn't otherwise changing
- [ ] A test fails without this change and passes with it (or I explained above why there isn't one)
- [ ] I understand every line of this diff and can explain it in review

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,6 +246,30 @@ Good-faith security research is **authorized** under our
 [safe-harbor policy](SECURITY.md#safe-harbor), and we're happy to credit valid
 first reports.
 
+---
+## ⭐ Star History
+
+<p align="center">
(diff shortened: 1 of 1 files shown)


Mubelotix:
https://star-history.dera.page provides the OG Star History experience

PR #476, opened 2026-08-07
fix(webmail): use request origin fallback for redirects and loopback host

farrasrayhand:
### Description
Fixes issue #391:
1. **Webmail Redirects**: Uses `request.url` origin fallback when `VITE_PUBLIC_APP_URL` is undefined, preventing `/undefined/...` 404 redirects in production Docker builds.
2. **Loopback Host**: Configures `DEFAULT_SMTP_HOST` & `DEFAULT_IMAP_HOST` to use `127.0.0.1` for self-hosted deployments to avoid NAT Hairpinning / Connection timeout on Cloud VPS.
3. **Test Email Fallback**: Adds `127.0.0.1` fallback for SMTP verification if public DNS/IP connection times out.

### DIFF
--- a/apps/api/src/modules/mail/admin/test-email.service.ts
+++ b/apps/api/src/modules/mail/admin/test-email.service.ts
@@ -274,45 +274,61 @@ export async function sendTestEmail(
   // listens, and every send fails with `ECONNREFUSED 127.0.1.1:465` while
   // Postfix is healthy on 0.0.0.0:465. `servername` keeps TLS validating
   // against the hostname, so certificate checking is unchanged.
-  const connectHost = await resolveSubmissionAddress(smtpHost);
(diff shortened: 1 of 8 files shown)


Hydralerne:
Reviewed against current main while auditing #391. The current webmail already uses same-origin authentication and explicit `/login` redirects, so the `VITE_PUBLIC_APP_URL` failure addressed here no longer exists.

The proposed `127.0.0.1` backend fallback does not fit the current architecture: webmail can run in a separate container or on another server, and managed webmail must submit to its mail server's Postfix before Postfix uses the external relay. Closing this superseded PR rather than introducing a second routing path. The unresolved submission-timeout portion of #391 remains open with a request for current diagnostics.

Issue #429, opened 2026-08-04
[Bug + Feature Request] Deleted emails stuck with TRASH label & no auto-refresh on new email


farrasrayhand:
## Bug Report

**Describe the bug**
When deleting an email from the inbox in the OpenShip webmail interface (`mail.example.com`), the email is not permanently removed. Instead, it remains visible in the inbox list with a `TRASH` label tag. Clicking the delete action again has no effect — the email cannot be permanently deleted through the UI.

**To Reproduce**
1. Log in to OpenShip webmail
2. Open the Inbox
3. Hover over an email and click the delete (trash) icon
4. Observe that the email remains in the inbox with a `TRASH` label instead of being moved to Bin/Trash folder or deleted
5. Attempt to delete again — no action occurs

**Expected behavior**
- Email should be moved to the Bin/Trash folder upon first delete action, OR
- Emails already labeled `TRASH` should be permanently deletable via a second action (e.g. "Delete permanently")

**Actual behavior**
- Emails remain in the Inbox with a `TRASH` label badge
- No way to permanently remove them through the UI

**Additional context**
This may be related to how OpenShip handles IMAP flags/labels vs. actual folder moves. The delete action appears to only apply a label (`TRASH`) rather than issuing an IMAP `MOVE` to the Trash folder or an `EXPUNGE` command.

---

## Feature Request

**Describe the feature**
Currently, the inbox does not automatically refresh when a new email arrives. Users are required to manually reload the page or re-navigate to the inbox to see incoming emails. This creates a poor user experience, especially for users monitoring time-sensitive emails.

**Expected behavior**
- The inbox should automatically refresh (or show new emails) when a new message is received, without requiring a manual page reload.

**Proposed implementation**
- Implement periodic polling (e.g., every 30–60 seconds) to check for new messages via IMAP or the existing API layer.
- Alternatively, use a push-based approach (e.g., WebSocket or Server-Sent Events) to listen for new mail events in real-time.

**Additional context**
This is especially important for users running OpenShip as their primary webmail interface. Without auto-refresh, the inbox behaves more like a static snapshot rather than a live email client.

Hydralerne:
The deletion bug is fixed in contributor PR #430, adapted and merged into the #892 integration branch (5dbd9b2b1455739791d4533a2921f0de676e8201), awaiting merge into main. The bin action now reaches the existing IMAP delete handler; optimistic, context-menu and keyboard actions preserve the source folder, including permanent deletion from Bin.

[...]

PR #478, opened 2026-08-07
fix(webmail): move deleted messages to Trash mailbox and enable 30s inbox polling

farrasrayhand:
### Description
Fixes issue #429:
1. **Trash Move**: Updates `modifyLabels` in IMAP driver to execute `messageMove(found, 'Trash')` when `TRASH` label is added, ensuring messages move into the Dovecot `Trash` mailbox instead of staying in `INBOX`.
2. **Auto Refresh**: Configures `refetchInterval: 30000` (30s) and `refetchOnWindowFocus: true` in `useThreads` hook for automatic inbox updates when new emails arrive.

### DIFF
--- a/apps/email/client/hooks/use-threads.ts
+++ b/apps/email/client/hooks/use-threads.ts
@@ -35,9 +35,10 @@ export const useThreads = () => {
         // explicit refresh button). Without this, stale-while-revalidate
         // refetches were overwriting a just-marked row with the older
         // server snapshot and visibly "reverting" the action.
-        staleTime: Infinity,
-        refetchOnMount: false,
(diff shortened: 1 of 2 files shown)


Hydralerne:
Thanks for the contribution. The deletion fix is now integrated through #430 into #892, preserving the selected source mailbox for all delete paths and reusing the existing IMAP delete handler. Adding a second deletion path inside modifyLabels would duplicate that logic and still leaves source-folder handling inconsistent.

The polling portion is a separate feature request, and the current inbox cache deliberately avoids background refetches that can overwrite pending optimistic actions. That needs its own design and validation. Closing this overlapping PR; #429 remains open for automatic refresh.

PR #480, opened 2026-08-07
perf(ci): add setup-qemu-action step for multi-arch Docker image builds

farrasrayhand:
### Rationale & Motivation
When publishing multi-architecture Docker images (`linux/amd64` and `linux/arm64`) for OpenShip (`openship-api`, `openship-dashboard`, `openship-edge`, `openship-mail`), GitHub Actions runner builds for `arm64` were taking over 2 hours or hanging indefinitely during `bun run build`.

### Root Cause
1. **Runner Label Mismatch**: The matrix runner label for `arm64` in `.github/workflows/docker-images.yml` was set to `ubuntu-24.04-arm` (missing `64`). GitHub's official native ARM runner label is `ubuntu-24.04-arm64`. Because of this typo, GitHub Actions fell back to x86_64 runners running user-space QEMU emulation.
2. **Missing QEMU Setup**: `docker/setup-qemu-action@v3` was missing prior to `docker/setup-buildx-action@v3`, causing Buildx syscall emulation bottlenecks when cross-compiling Next.js/React bundles under Docker.

### Key Changes
1. Updated `runner: ubuntu-24.04-arm` to `runner: ubuntu-24.04-arm64` in `.github/workflows/docker-images.yml` matrix.
2. Added `docker/setup-qemu-action@v3` step before `docker/setup-buildx-action@v3`.

### Impact
- Multi-arch Docker builds now run natively on GitHub-hosted ARM64 runners instead of 2+ hour QEMU emulation.
- Image build times are reduced from 2+ hours down to ~2 minutes.

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -52,6 +52,9 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
 
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -45,7 +45,7 @@ jobs:
           - platform: amd64
             runner: ubuntu-24.04
           - platform: arm64
-            runner: ubuntu-24.04-arm
+            runner: ubuntu-24.04-arm64
     steps:
       - name: Checkout
         uses: actions/checkout@v7


Hydralerne:
Reviewed against current main as part of the bug-fix integration. This change would regress the Docker release workflow, so I am closing it without merging.

`ubuntu-24.04-arm` is the documented GitHub-hosted native ARM64 label. The proposed `ubuntu-24.04-arm64` is not a supported standard label, and an unmatched runner label does not fall back to an x86 runner. The workflow already assigns each architecture to its native runner and builds only that architecture; adding QEMU does not correct a demonstrated defect here.

Evidence: [GitHub's supported-runner list](https://github.com/github/docs/blob/main/data/reusables/actions/supported-github-runners.md) names `ubuntu-24.04-arm`, and all five ARM64 image jobs in the [v0.7.2 release run](https://github.com/oblien/openship/actions/runs/33961809777) succeeded with the existing workflow. The dashboard ARM64 job completed in about five minutes.

PR #481, opened 2026-08-07
ci(docker): publish each image independently so one stuck build can't block a release

Ar1sC:
## Problem

The `docker-images` workflow builds all four images (`api`, `dashboard`, `edge`, `mail`) in **one** 4×2 matrix job, and every `merge-images` job waits on that whole job. That means one image's stuck/failed build blocks **every** image's version tag.

That's exactly what happened with **v0.6.1** (published 2026-08-07): the run ([#31168013124](https://github.com/oblien/openship/actions/runs/31168013124)) got stuck on `Build dashboard (arm64)` for hours (normally ~5 min — the v0.6.0 run finished the whole pipeline in 7 min). The `edge` amd64+arm64 builds completed in ~1 minute, but because `merge-images` waits for all builds, `ghcr.io/oblien/openship-edge:0.6.1` never appeared — so self-hosted instances trying to update to 0.6.1 got `not found` and stayed on 0.5.0.

## Change

- Split the build job **per image** (`build-api` / `build-dashboard` / `build-edge` / `build-mail`), each with its own amd64+arm64 platform matrix.
- Split the merge job **per image** too; `merge-<image>` only `needs: build-<image>`, so each image publishes as soon as its own two arch builds finish.
- Add `timeout-minutes: 30` to the build jobs so a hung build fails loudly instead of holding the concurrency group (and any queued re-push of the same tag) hostage for hours.

## Impact

A stuck or failed build now delays only its own image. The other images' tags (and `:latest` on tag pushes) publish on schedule.

Note for maintainers: this fixes future releases. For the already-affected **v0.6.1**, the current run is still stuck on dashboard arm64 — canceling it (or letting it hit the timeout) and re-pushing/`workflow_dispatch`-ing the tag will publish the missing `0.6.1` tags, including `openship-edge`.

### DIFF
--- a/.github/workflows/docker-images.yml
+++ b/.github/workflows/docker-images.yml
@@ -32,14 +32,24 @@ permissions:
 
 jobs:
   # Native per-arch builds (no QEMU — reuses the ubuntu-24.04-arm runner). Each
-  # arch is pushed to GHCR BY DIGEST; merge-images assembles the manifest lists.
-  build-images:
-    name: Build ${{ matrix.image }} (${{ matrix.platform }})
+  # arch is pushed to GHCR BY DIGEST; the matching merge job assembles the
+  # manifest list.
+  #
+  # One build job PER IMAGE instead of a single 4×2 matrix, so each image
+  # publishes on its own: the merge for an image only needs that image's two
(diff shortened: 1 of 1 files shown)


Hydralerne:
its design choise

Issue #350, opened 2026-07-31
postgres container start failed


jackychenHY:
follow the Docker Compose installation instructions, and the container "openship-postgres-1" failed to up.
the errors as foolows, so how to solve this problem?:

fixing permissions on existing directory /var/lib/postgresql/data ... ok
creating subdirectories ... ok
selecting dynamic shared memory implementation ... posix
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default time zone ... UTC
creating configuration files ... ok
2026-07-31 08:16:01.170 UTC [34] LOG:  could not write to file "postmaster.pid": Operation not permitted
2026-07-31 08:16:01.178 UTC [34] FATAL:  could not write to file "pg_wal/xlogtemp.34": Operation not permitted
child process exited with exit code 1
initdb: removing contents of data directory "/var/lib/postgresql/data"

Hydralerne:
on it

PR #494, opened 2026-08-07
fix(cli): detect existing pgdata/ subdir when resolving OPENSHIP_PGDATA

### LINKED ISSUE #487: OPENSHIP_PGDATA detection picks the volume root when re-installing over an existing pgdata/ subdirectory
### Summary

When re-installing over an existing Postgres volume, `openship up` detects the pre-existing volume but writes the wrong `OPENSHIP_PGDATA` — the volume root instead of the `pgdata/` subdirectory the previous install actually used. Postgres then refuses to start and the whole stack fails.

### What happens

The compose file defaults to:

```yaml
PGDATA: ${OPENSHIP_PGDATA:-/var/lib/postgresql/data/pgdata}
```

so the existing install had its cluster at `/var/lib/postgresql/data/pgdata`. On re-run, `openship up` detected the pre-existing volume and wrote:

```
OPENSHIP_PGDATA=/var/lib/postgresql/data
```

Postgres treated that as an uninitialised data directory, tried to `initdb`, and failed:

```
initdb: error: directory "/var/lib/postgresql/data" exists but is not empty
```

The container stayed unhealthy, and every dependent service failed to start with it.

### What was expected

Detection should resolve to the path the existing cluster is actually in, i.e. the same value the previous install used.

### Reproduce

1. `openship up --compose` on a fresh host. The cluster is created at `/var/lib/postgresql/data/pgdata` (compose default).
2. Re-run `openship up --compose` against the same volume.
3. The generated `.env` contains `OPENSHIP_PGDATA=/var/lib/postgresql/data`.
4. The postgres container fails with the `initdb` error above and never becomes healthy.

Fixing `OPENSHIP_PGDATA` back to `/var/lib/postgresql/data/pgdata` and restarting resolves it.

### Why this is hard to diagnose

The visible symptom is "the stack won't start" plus an unhealthy postgres container. `initdb: directory exists but is not empty` reads like volume corruption or a leftover from a previous failed install, and the tempting fix — deleting the volume — destroys the data the detection was meant to preserve. The `.env` value is generated, so there is no reason to suspect it.

### Suggestion

Probe for `PG_VERSION` in both the volume root and the `pgdata/` subdirectory and set `OPENSHIP_PGDATA` to whichever contains it. If neither does and the directory is non-empty, fail loudly rather than writing a guess.

### Environment

Ubuntu 22.04, single box, Docker Compose install (`openship up --compose`), openship CLI 0.6.0/0.6.1, self-hosted.

santhiprakash:
## Summary

Fixes `resolvePgData` so re-running `openship up --compose` over an existing Postgres data volume that already has the cluster at `/var/lib/postgresql/data/pgdata` keeps `OPENSHIP_PGDATA` pointed at the `pgdata/` subdirectory, instead of defaulting to the volume root and causing `initdb: error: directory ... exists but is not empty`.

## Motivation

`dbVolumeExists` only tells us that the `<project>_postgres_data` volume exists; it doesn't say where the cluster is inside it. Since the EPERM fix in #350, fresh installs use a `pgdata/` subdirectory, but older volumes may still have the cluster at the volume root. The previous code returned the root whenever the volume existed, which broke reinstalls over volumes created with the new subdirectory.

Closes #487.

## Changes

- `apps/cli/src/lib/compose.ts`
  - Adds `pgDataLocation()` to inspect the volume: `pgdata/PG_VERSION` → `subdir`, root `PG_VERSION` → `root`, empty → `empty`, anything else → `unknown`.
  - `resolvePgData()` uses the probe to pick the correct `OPENSHIP_PGDATA` path, and fails loudly when the volume is non-empty but unrecognizable.
  - Updates the `OPENSHIP_PGDATA` env comment to reflect the new logic.
- `apps/cli/test/unit/compose-pgdata.test.ts` (new)
  - 4 unit tests: fresh install, existing `pgdata/` cluster, legacy root cluster, and an unrecognizable non-empty volume.

## Verification

```bash
$ cd /home/ubuntu/Projects/open-source/openship-pgdata-487
$ python3 /home/ubuntu/Projects/open-source/scripts/preflight_ship.py \
    --repo oblien/openship --local openship-pgdata-487 --branch main --issue 487 \
    --file apps/cli/src/lib/compose.ts \
    --must-contain 'return dbVolumeExists(composeProjectName(prev)) ? PGDATA_ROOT : `${PGDATA_ROOT}/pgdata`;' \
    --search 'OPENSHIP_PGDATA in:body'
PREFLIGHT CLEAR — safe to implement against current upstream.

$ bun run --cwd apps/cli lint
$ tsc --noEmit

$ bun run --cwd apps/cli test
 Test Files  28 passed (28)
      Tests  229 passed (229)
```

## Checklist

- [x] `bun run --cwd apps/cli test` and `bun run --cwd apps/cli lint` pass locally

### DIFF
--- a/apps/cli/src/lib/compose.ts
+++ b/apps/cli/src/lib/compose.ts
@@ -603,18 +603,62 @@ function composeProjectName(prev: Record<string, string>): string {
  *
  * A fresh install uses a subdirectory (`…/data/pgdata`) rather than the bare
  * mount root: initdb against a mount root fails on quirky host filesystems with
- * "Operation not permitted" (WAL preallocation / lost+found — see #350). But a
- * pre-existing install already has its DB at the mount ROOT, and moving PGDATA
- * would make Postgres init a fresh empty DB and orphan the old one. So the
- * decision is made ONCE and then pinned in `.env` (same sticky rule as
- * COMPOSE_PROJECT_NAME): re-runs reuse it; a volume that predates this pin keeps
- * the root. The check uses the resolved project name so it inspects the right
- * `<project>_postgres_data` volume.
+ * "Operation not permitted" (WAL preallocation / lost+found — see #350). A
+ * pre-existing install may have its DB at the mount ROOT or in the `pgdata/`
+ * subdirectory, depending on how it was created; moving PGDATA would make
+ * Postgres init a fresh empty DB and orphan the old one. So the decision is
+ * made ONCE and then pinned in `.env` (same sticky rule as
+ * COMPOSE_PROJECT_NAME): re-runs reuse it; a volume that predates this pin is
+ * probed for PG_VERSION and the existing path is preserved. The check uses the
+ * resolved project name so it inspects the right `<project>_postgres_data`
+ * volume.
  */
 const PGDATA_ROOT = "/var/lib/postgresql/data";
+
+/**
+ * Probe the existing `<project>_postgres_data` volume to find where the
+ * Postgres cluster lives. Returns:
+ *   "subdir"  – `pgdata/PG_VERSION` exists (fresh-install default),
+ *   "root"    – `PG_VERSION` exists at the volume root (legacy pre-subdir data),
+ *   "empty"   – the volume has no entries (safe to initialize in the subdir),
+ *   "unknown" – non-empty but no recognizable PG_VERSION; caller should fail.
+ */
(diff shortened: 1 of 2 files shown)


santhiprakash:
Closing — upstream main now includes its own PGDATA volume probing implementation (probePgDataDir / resolvePgData in apps/cli/src/lib/compose.ts) that addresses #487, making this change redundant.

PR #496, opened 2026-08-07
fix(cli): refuse to regenerate POSTGRES_PASSWORD when a data volume already exists

### LINKED ISSUE #488: up dry-run reports POSTGRES_PASSWORD=<preserved> but writes a newly generated password
### Summary

Re-running `openship up` over an existing install claims in its dry run that secrets are preserved, but the `.env` it writes contains a newly generated `POSTGRES_PASSWORD`. The API cannot then authenticate against the existing database.

### What happens

The dry run for a re-install printed:

```
POSTGRES_PASSWORD=<preserved>
```

and described the operation as one that "re-runs an existing install (secrets preserved)".

The `.env` that was actually written contained a different password from the one the existing Postgres volume had been initialised with. Postgres kept the original password (it is only set at `initdb` time), so the API failed with:

```
error: password authentication failed for user "openship"
```

and the api container never became healthy.

### What was expected

Either the previous password is genuinely read from the existing `.env` and reused, or the dry run does not claim preservation.

### Reproduce

1. `openship up --compose` on a fresh host; let Postgres initialise.
2. Re-run `openship up --compose` over it and read the dry-run output — it reports `POSTGRES_PASSWORD=<preserved>`.
3. Compare `POSTGRES_PASSWORD` in the newly written `~/.openship/compose/.env` with the previous value. They differ.
4. The api container fails with `password authentication failed for user "openship"`.

Recovery required manually restoring the old password into `.env`.

### Why this is hard to diagnose

The dry run explicitly states the opposite of what happens, so the `.env` is the last place anyone would look. `password authentication failed` in a stack where the CLI manages both sides of the credential looks like a database corruption or migration problem. And because the old `.env` has already been overwritten by the time the failure appears, the original password is gone from config unless it was captured beforehand. It is recoverable — start the container with `POSTGRES_HOST_AUTH_METHOD=trust` and reset the role, or edit `pg_hba.conf` — but that is not obvious to someone who has just been told their secrets were preserved.

### Suggestion

- Actually read the prior `.env` (or the volume) and reuse the existing secret.
- If preservation is not implemented for a given key, do not print `<preserved>` for it.
- Consider refusing to write a new `POSTGRES_PASSWORD` when an initialised data volume is present, since the new value cannot possibly be correct.

### Environment

Ubuntu 22.04, single box, Docker Compose install (`openship up --compose`), openship CLI 0.6.0/0.6.1, self-hosted.

santhiprakash:
## Summary

`openship up --compose` no longer silently mints a new `POSTGRES_PASSWORD` when an existing `<project>_postgres_data` volume is present and the current `.env` is missing the key. The dry-run plan now reports this unsafe case, and a real run throws a clear recovery message instead of overwriting the secret.

## Motivation

Fixes #488. Re-running `openship up` over an existing install could show `POSTGRES_PASSWORD=<preserved>` in the dry-run plan while the `.env` it wrote contained a newly generated value. Postgres only applies `POSTGRES_PASSWORD` at `initdb` time, so the volume kept the old password while the API used the new one, failing with `password authentication failed for user "openship"` (28P01).

## Related issue

Closes #488

## Changes

- `apps/cli/src/lib/compose.ts`
  - Add `keepPostgresPassword`: carry the existing `.env` value, generate only on a genuine fresh install, and refuse (real run) or warn (dry run) when an initialized data volume exists but `POSTGRES_PASSWORD` is missing.
  - Thread a `dryRun` flag through `renderEnv` so the preview renders a placeholder while the real run enforces the guard.
  - Add `warnings` to `ComposePlan` and emit the volume-warning message in `composePlan` instead of claiming the secret is preserved.
- `apps/cli/src/lib/up-plan.ts`: print compose plan warnings in the dry-run summary.
- `apps/cli/test/unit/compose-env-preserve.test.ts`: add tests for preservation, refusal, and dry-run warning.
- `apps/cli/test/e2e/up-dry-run.test.ts`: update the `composePlan` mock to include `warnings`.

## Verification

```bash
cd apps/cli
bun run lint            # tsc --noEmit, exit 0
bun run test            # 27 files, 228 tests pass
bun run test -- compose-env-preserve  # 28 tests pass
bun run test -- up-dry-run            # 7 tests pass
```

Note on `bun format`: running the root `bun format` script rewrites pre-existing prettier drift in files I did not change (e.g. `apps/cli/src/lib/compose.ts` has unrelated long-line formatting drift). I left that untouched and hand-formatted the changed lines to the repo's existing style, matching the approach in #385.

## Checklist

- [x] `bun run test` and `bun run --cwd apps/cli lint` pass locally

### DIFF
--- a/apps/cli/src/lib/compose.ts
+++ b/apps/cli/src/lib/compose.ts
@@ -432,7 +432,35 @@ volumes:
 
 /** Persist a stable secret in the compose .env — regenerated only if absent. */
 function keepSecret(existing: Record<string, string>, key: string): string {
-  return existing[key] || randomBytes(32).toString("hex");
+  const value = existing[key];
+  if (value) return value;
+  return randomBytes(32).toString("hex");
+}
+
+/**
+ * Preserve the Postgres password, or generate one on a genuine first install.
+ *
+ * Postgres only applies `POSTGRES_PASSWORD` when it initialises an EMPTY data
+ * dir. If a `<project>_postgres_data` volume already exists, generating a new
+ * password would leave the volume on the old password and the API on the new
+ * one — the api then crash-loops on `password authentication failed for user`
+ * (28P01). So: carry the existing password when we have one. If we don't have
+ * one and a volume exists, refuse rather than silently break the database. On a
+ * true fresh install (no volume), generate a new password.
+ */
+function keepPostgresPassword(existing: Record<string, string>, dryRun = false): string {
+  const value = existing.POSTGRES_PASSWORD;
+  if (value) return value;
+  const project = composeProjectName(existing);
+  if (dbVolumeExists(project)) {
+    const msg =
+      `Existing Postgres data volume ${project}_postgres_data already has a password, but POSTGRES_PASSWORD is missing from ${ENV_FILE}. ` +
+      `A new password cannot be safely generated for this volume. ` +
+      `Restore the previous .env or run \`openship uninstall\` to start fresh (this destroys DB data).`;
+    if (dryRun) return randomBytes(32).toString("hex"); // placeholder for the preview
+    throw new Error(msg);
+  }
+  return randomBytes(32).toString("hex");
 }
 
 /** Parse the existing .env so re-running `up` preserves generated secrets. */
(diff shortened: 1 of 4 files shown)


santhiprakash:
Closing — upstream main now includes its own secret-rotation guard (secretRotationRisk with --reset-secrets in apps/cli/src/lib/compose.ts) covering the #488 scenario, making this change redundant.

PR #499, opened 2026-08-07
fix(deployments): don't let an all-carried compose redeploy take over the active deployment

### LINKED ISSUE #498: Compose deploys: an all-carried-forward redeploy is marked `ready` and takes over the active deployment, desyncing the dashboard from reality (site serves while UI shows "Stopped"; rollback/logs/restart break)
## Summary

On a multi-service (compose) project, a redeploy in which **every service is "carried forward"** (already up to date, nothing rebuilt) is recorded as a successful **`ready`** deployment, `activeDeploymentId` advances onto it, and the parent deployment's `containerId` resolves to the **database** container. Because web routing is by a fixed per-project host port, the site keeps serving — so the dashboard reads **Stopped**, rollback breaks, and logs/restart/stop act on Postgres. Verified on current `main` (v0.6.1).

## What you see

- Dashboard shows the app **Stopped** — `Logs Error: Service has no running container`, logs unreachable — while the site serves 200s with fresh sessions.
- Deploy history shows several consecutive deploys with ~0 ms build, no image, no port check, each marked `ready`.

## Why it's invisible (the key detail)

Routing keys on the project's **fixed host port** (`packages/adapters/src/runtime/host-port.ts` — persisted on the project, reused across redeploys/restarts); the nginx vhost `proxy_pass`es to it with no deployment id. Status and logs key on the **deployment record** (`activeDeploymentId → containerId`). Nothing reconciles the two, so they can silently diverge and each still "looks right" from its own side.

## Repro

1. A compose project with an app that `dependsOn` a database (e.g. app + Postgres).
2. Deploy it; confirm it's Live.
3. Redeploy without changing anything the app rebuilds on (or trigger a smart-deploy whose target subset is empty), so every service is carried forward.
4. Observe: the new deployment is marked `ready` with ~0 ms build / null image ref / no `portCheck`; `activeDeploymentId` moves onto it; the dashboard flips to **Stopped** while the site keeps serving; the deployment's recorded `containerId` is the **Postgres** container.

## Root cause (current `main`)

1. **A carried-forward service counts as success without building or port-checking.** In the compose deploy loop, the carry branch does `successful += 1` and `continue`s (`apps/api/src/modules/deployments/compose/deploy.service.ts:1047`), skipping the real-deploy path; the function returns `status: successful > 0 ? "ready" : "failed"` (`:2256`). An all-carried deploy therefore returns `ready` with `successful = N`, no port checks, ~0 ms build. (The single-app build path, by contrast, refuses to proceed without a built image.)
2. **The port check is skipped exactly when it's needed.** It's collected only in the real-deploy branch and attached only when non-empty (`compose/pipeline.ts:236-237`), so an all-carried deploy has no `portCheck` — the one signal that would catch "nothing new is actually listening for this deployment" is absent precisely when everything was skipped.
3. **The active pointer advances unconditionally on the ready no-op.** `onSuccess` is called for any non-failed compose result (`compose/pipeline.ts:222`), and the success path records `ready` and calls `setActiveDeployment` with no guard. (The failure and reconcile legs are deliberately careful *not* to advance — see the explicit `// Do not add a setActiveDeployment call here` at `deployment-lifecycle.ts:358` and `:478` — but the success leg has no equivalent no-op guard.)
4. **The recorded parent `containerId` is the DB's.** It's `composeResult.services.find((s) => s.containerId)` (`compose/pipeline.ts:212`), but `services` follow `topoSort` (`deploy.service.ts:123`), which emits dependencies before dependents — so for an app that `dependsOn` its DB, the first service with a container is the **database**. (This mis-resolution also affects *healthy* compose deploys, not only no-ops.)
5. **Routing is deployment-agnostic** (fixed host port, above), so the originally-working container keeps serving regardless of which deployment record is "active."
6. **Status/logs read the record, never actual state.** `getRuntimeLogs` / `enableProject` / `disableProject` resolve through `activeDeploymentId → dep.containerId` (`apps/api/src/modules/projects/project-runtime.service.ts`) and never compare against what's actually bound to the project's host port. The only reconciler (`reconcile.service.ts`) runs solely for deployments already in `reconciling` state and only checks recorded-id liveness — it never asks "what is actually serving this project's port?"

**Net:** routing keys on the project host port; status/logs key on the deployment record; nothing reconciles the two.

## Impact

- **Green deploys that ship nothing.** A real change classified as no-change / carry-forward is recorded `ready` without deploying — repeatable, always green. (In our case every deploy carried the same commit so nothing drifted, but a real code change would have silently not shipped, with a green result each time.)
- **Broken rollback.** These deploys carry a null image ref, so there's no artifact to roll back to.
- **Restart / Stop / Logs act on the database** — because the recorded container is the DB (step 4).

## Where it bites hardest

Multi-container framework stacks — app + Postgres + Redis + a queue worker — i.e. the Laravel/Rails multi-role shape being discussed in #231. A single-process Node app rarely triggers it (its build path already has the guard the compose path lacks). Distinct from the edge/route reconciliation in #463 (that reconciles edge vhosts ↔ project domains; this is deployment-record ↔ container drift).

## Fix direction

Two fixes — one to stop it happening, one to heal what it already broke:

**1. Prevent it (write path).** Don't mark a deploy `ready` or move the active pointer onto it when nothing was actually (re)built or deployed and no port check ran — a no-op guard, the same shape the single-app build path already has. And record the *exposed app's* container as the deployment's container, not the topo-first database.

**2. Heal + stay resilient (read path).** Projects already hit by this have the wrong container recorded, and there's no migration to backfill them. So when resolving logs / start / stop, reconcile against what's actually bound to the project's host port instead of trusting the stored container id — which both self-heals the existing bad records and keeps status honest going forward.

Prevention alone leaves already-broken projects broken until their next real deploy; healing alone keeps producing bad records — so both.

I have a fix implementing both, with tests and a clean typecheck, and I'm happy to open a PR. There are a couple of small product decisions in it — what terminal state a no-op redeploy should display, and whether "disable" on a compose project should stop the whole stack or just the app — that are genuinely yours to make, so I'd rather surface them in the PR than presume. Also happy to attach the raw deployment records / `docker inspect` output from the reproduction.

sudanese:
## The bug (#498)

On a multi-service (compose) project, a redeploy where **every service is carried forward** built, created and probed nothing — yet was recorded `ready`, advanced `activeDeploymentId` onto itself, and recorded the **database** as the deployment's container (the result list follows `topoSort`, so the first row with a container is the dependency an app `dependsOn`). Routing keys on a fixed per-project host port, so the site kept serving: the dashboard read **Stopped**, rollback had no image to restore, and logs/info/usage answered for Postgres.

## What changed since the first version of this PR

This is a re-implementation on current `main` rather than a rebase. Reviewing the original against `main` turned up four things worth stating plainly, because two of them made the original fix unsafe:

1. **The no-op verdict was unsound.** `deployed === 0 && failed === 0 && portChecks.length === 0` is also true of the prepare-failure return (which reports `status: "failed"` with `failed === 0`) and of the no-services return. Only branch ordering in `pipeline.ts` kept those from being classified "no changes". It also missed two ways a carried pass mutates real state — the de-listed-service reaper destroys containers, and an app-prepare step persists env — while the counters read zero.
2. **The guard would have broken rollback.** `isExternalUnchanged` excludes `trigger === "update"` but **not** `"rollback"`, and compares the running image against `svc.image` rather than the image this deploy intends (a rollback's pinned refs arrive via `opts.builtImages`, first read *past* the carry `continue`). So a rollback of an image-only compose stack carries everything forward and replays neither its frozen env nor its pinned images. That is already wrong on `main`; with a non-advancing no-op guard on top, `ROLLBACK_ALREADY_ACTIVE` would never trip and the operator could click Rollback forever. Its sibling `newerThanRestoredRelease` guards this exact hazard from the other direction and says so in its docblock. **Fixed at the cause**, which is a prerequisite for the guard.
3. **`primaryContainerId` was never set when the exposed app was itself carried** — it was assigned inside the real-deploy branch — so any partial redeploy that recreated a sidecar while carrying the app fell back to the topo-first row and recorded the database again. It is now derived at return time.
4. **Half the read-path change is now a regression.** `main` changed `enableProject`/`disableProject` to act on `deploymentContainerIds(dep)` — every container — and pins it with `project-toggle.test.ts`. Narrowing them to one exposed container would undo that, and also drop `isEdgeServedStatic`, the `isAbsent`/`isAlreadyInState` classification, and `disableProject`'s `disabledAt` rollback-on-failure. **Dropped**; the read-path change is scoped to logs and container info.

## The fix

**Prevent it (write path).** `summary.deployed` counts only the services actually (re)created or (re)served — derived from a per-service `carried` flag rather than a counter kept alongside `successful`, so a future success branch cannot silently forget it — plus `summary.mutated` for state changed outside that loop. `composeDeployMadeNoChanges` is gated *positively* on `ready` + a non-empty service set, so no reordering can reclassify a failure. An all-carried pass settles through a new `onNoChanges`: it records the row and nothing else — no pointer advance, no container. Deliberately **not** `onFailure`/`onCancelled`, both of which `runtime.destroy()` every service container the deployment lists; here those are the live ones still serving.

**Resolve the primary container once.** New `pickPrimaryServiceId` (beside `pickCanonicalDomainRow`, so the answer agrees with the access URL): the canonical verified domain's service, else the first exposed one, else the first. It feeds the deployment's recorded container on both the success and reconcile legs — the reconciler promotes its row to active without rewriting `containerId`, so it was a second producer of corrupted records.

**Heal what is already broken (read path).** Project logs, deployment logs, container info and usage resolve through `livePrimaryContainerId`, which asks the host via the existing `live-state` matcher and heals the `service_deployment` row — the same thing the per-service path already does. That fixes rows written wrong before this change without a migration. Single-app deployments short-circuit to the recorded id and issue no extra queries.

**Two more sites had the same defect, on public paths.** `project-route.service.ts` guarded on `!containerId`, which the `"compose"` sentinel passes — so a compose project's *project-level* custom domain fell through to whichever service the row named. And the webhook-domain vhost took `svcDeps.find((s) => s.containerId)` over rows returned in insertion (dependency) order. Both could `proxy_pass` at Postgres, which is a worse symptom than the one #498 reports. `COMPOSE_SENTINEL` + `isRealContainerRef` are now one module instead of one constant and three open-coded literals; `deploymentContainerIds` no longer returns the sentinel as a container (it made a pause report success having stopped nothing).

## Your two questions, answered by `main`

**1. What terminal state should a no-op show?** A new **`no_changes`**, not `cancelled`.

`deployment.status` is free text with no check constraint *specifically* so values can be added without a migration — its own docblock says so and names `deployment-status.test.ts` as the ratchet enumerating the guards a new value must reach. `reconciling` is the working precedent. So the cost is five mechanical guard edits and one i18n key, all of which that test names for you.

`cancelled` was the wrong side of the trade: `deployment.cancelled` has **no producer today**, so a successful no-op would have become its only one — a warning-toned audit line reading "System cancelled the deploy of X", a notification asserting someone stopped it, and "Install failed" in both install wizards. It also makes the drift-banner follow-up below inexpressible.

Wired: neutral **"No changes"** chip (both `getStatusConfig` copies), `SETTLED_HEALTHY_STATUSES`, `DeploymentCard` label, both install wizards routed to *done* rather than *failed*, `deployment.no_changes` in the audit taxonomy (`tone: "info"`) and in the notification map under `deploy.succeeded` with an `EVENT_HEADLINES` override so the title is not "Deploy succeeded" over a body saying nothing shipped. The SSE session still closes `ready` — `BuildSessionState` has no third terminal state — so the build page and CLI exit code are unchanged; the reason is persisted where the refresh path reads it. One key × 9 locales, all translated. **No migration.**

**2. Should disable stop the whole stack?** **Yes, and `main` already does it.** `enableProject`/`disableProject` fan out over `deploymentContainerIds(dep)`, with the reason in the comment and `project-toggle.test.ts` pinning it. Nothing to decide, and nothing here touches those paths.

## Notes

- One behaviour to know: a compose project containing a **static** service can never reach `deployed === 0` (a static row carries no container, so it is never carried), so the guard will not fire for those. Safe false-negative, documented in the code.
- The carried-row `ip`/`hostPort` correction (#506) is now mirrored onto the **active** deployment's row. That is the row the next deploy reads, and a non-advancing settle would otherwise strand the fix and let the stale binding come back.
- **Tests:** `compose-noop-settle.test.ts` drives the real `executeComposePipeline` — all-carried settles `no_changes`, does not advance, records no container, and **destroys nothing** (that last assertion is the guard against a future collapse into `onCancelled`); a deployed pass advances and records the app, not the db; a `mutated` pass is not a no-op. `project-logs-container.test.ts` is colocated in `src/` so `tsc` actually covers it (`tsconfig` includes only `src/**`, which is how the previous version's `as never` fixtures went unchecked) and runs the real live matcher. **Every one of them fails on unfixed code — verified by reverting each hunk.** The status ratchet now loops over the settled statuses instead of hardcoding one, so the next value gets seven failing tests naming exactly what to teach.
- **Green:** `tsc --noEmit` clean in `apps/api` and `apps/dashboard`; api 322 files / 3816 passed, dashboard 714, core 575, adapters 2832, db 89, cli 416.

## Deliberately left as follow-ups

1. **Pre-flight no-op detection at the trigger layer** — `webhook-push.ts` already returns `{skipped: true}` without creating a row, so ideally a no-op never mints a deployment at all. Also fixes the success-rate metric being diluted by no-op rows.
2. **The drift banner** reads the *active* row's `commitSha`, so a no-op carrying a new commit leaves "project outdated" on. Reachable only for a git-source project whose services are all image-only externals; the Update button is immune. A real status value is what makes the fix expressible.
3. **Webhook commit dedup is anchored on the active pointer** — extend it to any settled non-failed row for the commit.
4. **A port→container reconciler** for the parent id (`edgeProxy().byPort` exists with no consumer) — catches the *next* divergence (a hand-run `docker compose up`, an adoption) rather than only healing on read.
5. **`ORDER BY` on `listByDeployment`**, and moving the three remaining ad-hoc exposed-first orderings onto `pickPrimaryServiceId`.
6. **A one-shot backfill** of `deployment.container_id` for already-corrupted rows. Feasible as a pure `UPDATE`; deferred because the read-heal covers the surfaces that matter and the multi-exposed tie-break belongs with (5).

Fixes #498

### DIFF
--- a/apps/api/src/lib/container-ref.ts
+++ b/apps/api/src/lib/container-ref.ts
@@ -1,0 +1,20 @@
+/**
+ * The `deployment.container_id` / `image_ref` sentinel, and the one predicate for
+ * "is this a reference I can hand to a runtime".
+ *
+ * A leaf module on purpose: the sentinel is read on the write path (compose
+ * deploy, reconcile), the read path (logs, container info, routing upstreams) and
+ * in rollback planning, and every open-coded `=== "compose"` was a place that
+ * could forget.
+ */
+
+/** Stored in `deployment.container_id` / `image_ref` when a release has no single
+ *  primary container/image. A marker, never a real Docker reference — treating it
+ *  as one is what made compose rollback try `createContainer({ Image: "compose" })`
+ *  and made a project pause report success having stopped nothing. */
+export const COMPOSE_SENTINEL = "compose";
+
+/** True when a stored reference names something a runtime can actually act on. */
+export function isRealContainerRef(ref: string | null | undefined): ref is string {
+  return !!ref && ref !== COMPOSE_SENTINEL;
+}
--- a/apps/api/src/lib/deployment-runtime.ts
+++ b/apps/api/src/lib/deployment-runtime.ts
@@ -22,6 +22,7 @@ import {
   type RuntimeMode,
 } from "@repo/core";
 import { env } from "../config";
+import { isRealContainerRef } from "./container-ref";
 import { cloudClient, getOrgCloudToken } from "./cloud/client";
 import { resolveOrgCloudUserId } from "./cloud/transport";
 import { platform } from "./controller-helpers";
@@ -109,6 +110,13 @@ export interface DeploymentMeta {
    * agree on one path.
    */
   staticServeOutputDir?: string;
+  /** This deploy carried every service forward and shipped nothing, so it never
+   *  became the active release (see `onNoChanges`). Mirrors the row's
+   *  `no_changes` status for clients reading meta rather than status. */
+  noChanges?: boolean;
+  /** Compose roll-up. Loosely typed on purpose — the pipeline writes a wider
+   *  object than any one reader needs. */
+  composeDeployment?: { warningMessage?: string } & Record<string, unknown>;
 }
 
 /**
@@ -791,7 +799,9 @@ export async function deploymentContainerIds(
   const rows = await repos.service.listByDeployment(dep.id);
   const serviceIds = [...new Set(rows.map((r) => r.containerId).filter((id): id is string => !!id))];
   if (serviceIds.length > 0) return serviceIds;
-  return dep.containerId ? [dep.containerId] : [];
+  // The compose sentinel is a marker, not a container: returning it made a pause
+  // report success having stopped nothing (docker 404 → `isAbsent` → swallowed).
+  return isRealContainerRef(dep.containerId) ? [dep.containerId] : [];
 }
 
 /**
--- a/apps/api/src/lib/notification-categories.ts
+++ b/apps/api/src/lib/notification-categories.ts
@@ -247,6 +247,10 @@ const EVENT_TYPE_TO_CATEGORY: Record<string, string> = {
   // Deploy
   "deployment.failed": "deploy.failed",
   "deployment.succeeded": "deploy.succeeded",
+  // A no-op redeploy is a SUCCESSFUL outcome, so it rides the succeeded toggle —
+  // not "cancelled", which tells the operator to try again. Its own headline below
+  // keeps the message honest about nothing having shipped.
+  "deployment.no_changes": "deploy.succeeded",
   "deployment.cancelled": "deploy.cancelled",
   "build.failed": "deploy.failed",
   "build.fatal_error": "deploy.failed",
@@ -317,6 +321,12 @@ export const EVENT_HEADLINES: Record<string, { title: string; description: strin
   "server.reachable": {
     title: "Server reachable",
     description: "A server's Docker daemon is answering again, with how long it was gone.",
+  },
+  // Same reasoning: it subscribes as a success, but "Deploy succeeded" over a body
+  // saying nothing shipped is the exact mismatch this override exists for.
+  "deployment.no_changes": {
+    title: "No changes to deploy",
+    description: "A redeploy found every service already up to date; the live release didn't move.",
   },
 };
 
--- a/apps/api/src/lib/public-endpoints.ts
+++ b/apps/api/src/lib/public-endpoints.ts
@@ -929,6 +929,25 @@ export function pickCanonicalDomainRow<
 }
 
 /**
+ * The service a PROJECT-level question about "the container" means — its logs,
+ * its container info, the upstream a project-level domain proxies to. Keyed off
+ * the canonical domain row first, so the answer agrees with the access URL.
+ *
+ * The thing it replaces is `find((s) => s.containerId)` over a service list in
+ * topoSort order: dependencies come first there, so for an app that `dependsOn`
+ * its database that find answered "postgres" (#498).
+ */
+export function pickPrimaryServiceId<S extends { id: string; exposed?: boolean | null }>(
+  services: S[],
+  domainRows?: Array<Pick<ProjectDomainRow, "verified" | "isPrimary" | "serviceId">> | null,
+): string | null {
+  if (services.length === 0) return null;
+  const canonical = pickCanonicalDomainRow(domainRows)?.serviceId;
+  if (canonical && services.some((svc) => svc.id === canonical)) return canonical;
+  return (services.find((svc) => svc.exposed) ?? services[0]).id;
+}
+
+/**
  * The canonical access URL for a project, computed from ALL its domain rows +
  * the effective deploy target + port. Correct for a project whose only domains
  * are service-scoped — the project-level publicEndpoints resolver excludes those,
--- a/apps/api/src/modules/analytics/analytics.service.ts
+++ b/apps/api/src/modules/analytics/analytics.service.ts
@@ -19,6 +19,7 @@ import { repos } from "@repo/db";
 import { NotFoundError, AppError, safeErrorMessage } from "@repo/core";
 import type { ResourceUsage } from "@repo/adapters";
 import { resolveDeploymentRuntimeForRead } from "../../lib/deployment-runtime";
+import { livePrimaryContainerId } from "../services/service-container";
 import {
   resolveProjectTrafficSources,
   fetchMgmt,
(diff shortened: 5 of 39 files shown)


Hydralerne:
on it

PR #508, opened 2026-08-08
fix(routing): fall back to container IP when loopback port is not published

### LINKED ISSUE #506: [Bug] Same-server migration leaves app unreachable — "Auto" routing mode doesn't publish a loopback port
# Bug report: same-server migration leaves app unreachable — "Auto" routing mode doesn't publish a loopback port

## Summary

After running Openship's built-in **same-server migration** tool to bring an existing Dokploy-managed service into Openship, the resulting project was completely unreachable through its configured custom domain — despite the dashboard showing the project as **Live**, the domain as **Verified**, and SSL as **Active**.

Root cause: the project's edge routing mode defaults to **Auto → Loopback port**, but the migrated container's port was never actually published to `127.0.0.1` on the host. The `openship-edge` container runs in `network_mode: host`, so it has no way to reach the app except via that (missing) loopback binding or an explicit route. Manually switching the project's routing mode to **Container IP** fixed connectivity immediately, with no other change.

## Environment

- OS: Ubuntu 24.04.3 LTS, Docker
- Server previously ran Dokploy; migrated to Openship via the desktop/dashboard migration flow, which also stopped Dokploy's Traefik instance to avoid port clashes
- App: Uptime Kuma (`louislam/uptime-kuma:2.1.0`), migrated **same-server** from Dokploy
- App container: isolated per-project bridge network `openship-uptime-kuma` (subnet `172.19.0.0/16`), container IP `172.19.0.2`, app listening internally on port `3001`
- `openship-edge`: `network_mode: host`, not attached to the app's bridge network

## Steps to reproduce

1. Have an existing service (e.g. Uptime Kuma) running under Dokploy on a target server.
2. In Openship, run a **same-server migration** for that service.
3. Let the migration proceed through: creating the Openship project → moving volume data → deploying on the target server → health-checking the new deployment. It stops at **"target verified healthy — awaiting cutover confirmation."**
4. On the confirmation screen, click **"Stop & remove originals"** (or "Keep originals" — same effect on a same-server migration) to finish the cutover.
5. Note that the project's **Domains & Routes → Advanced → Routing** mode is set to **Auto (Recommended — loopback port)** by default.
6. Visit the project's custom domain. Result: connection fails (for a WebSocket-dependent app like Uptime Kuma, the static bundle may still load and surface the app's own generic error — e.g. "Cannot connect to the backend server" — obscuring the real cause).
7. On the server, confirm nothing is actually published for that container:
   ```
   docker port <container_name>          # empty output
   curl -sv http://localhost:3001        # Connection refused on both ::1 and 127.0.0.1
   ```
   ...despite the dashboard reporting the project as Live and the domain as Verified/SSL Active.
8. Confirm the app itself is healthy and reachable via its Docker bridge IP:
   ```
   docker inspect <container_name> --format '{{json .NetworkSettings.Networks}}'
   # → shows the app on its own bridge network, e.g. 172.19.0.2

   curl -sv http://172.19.0.2:3001
   # → succeeds, valid HTTP 302 to /dashboard
   ```
9. Confirm `openship-edge` has no route to that bridge network:
   ```
   docker inspect openship-edge --format '{{json .NetworkSettings.Networks}}'
   # → {"host": {..., "IPAddress": ""}} — host network mode, not attached to the app's bridge network
   ```
10. Workaround: in the dashboard, go to the project → **Domains & Routes → Advanced → Routing**, switch from **Auto** to **Container IP**, and save. The domain starts routing correctly immediately.

## Expected vs. actual

**Expected:** with the default "Auto (Recommended — loopback port)" mode, Openship should either (a) actually publish the migrated container's port to `127.0.0.1` as part of the migration/cutover, or (b) detect that no loopback port is bound and fall back to Container IP routing automatically — rather than leaving the route silently broken while reporting the project as Live and the domain as Verified.

**Actual:** the loopback port is never published for a same-server-migrated container, the edge has no way to reach it, and none of the dashboard's health indicators (Live / Verified / SSL Active) reflect this.

## Secondary observations (possibly related, possibly separate issues)

- **Misleading frontend error:** for WebSocket-dependent apps (Uptime Kuma), the failure surfaces as the app's own generic in-app message ("Cannot connect to the backend server...") rather than an edge-level error, which sent debugging in the wrong direction initially (looked like a WebSocket/Upgrade-header proxy issue rather than a fully missing route).
- **Inconsistent port shown in UI:** the project's **Overview** page ("Infrastructure" panel) displayed **Port 3000** (Uptime Kuma's default) while **Domains & Routes** correctly showed **"Mapped to: Port 3001"** (the container's actual listening port, since it was originally configured with a custom port under Dokploy). Cosmetic, but confusing while debugging.
- **Route table doesn't seem to update atomically:** immediately after switching routing mode to Container IP and saving, one browser session hit a TLS certificate warning (`NET::ERR_CERT_AUTHORITY_INVALID`) while a fresh incognito session hit Openship's own **"404 · no route"** fallback page. This resolved within a short time without further action, but suggests the route/cert update isn't applied instantly or atomically after a routing-mode change.

## Suggested fix

Either publish the loopback port as part of same-server migration cutover (so "Auto" behaves as advertised), or have "Auto" mode detect a missing loopback binding and transparently use Container IP routing instead. In the meantime, manually setting routing mode to **Container IP** is a reliable workaround for anyone hitting this after a same-server migration.

santhiprakash:
## Summary

Fixes #506.

Same-server migrations that attach the running container in place (reuse) never publish a `127.0.0.1` host port, so the default **Auto → loopback-port** strategy left the app unreachable through its custom domain. The edge had no loopback binding to dial.

## Change

Make the live route-apply paths inspect the container first, then fall back to the container's bridge/host IP when no loopback host port is bound:

- `resolveUpstreamUrl` now falls through from a loopback host port to the container IP when `hostPort` is absent.
- `applyProjectRouting` resolves a live upstream per `service_deployment` row, using `getContainerInfo` for the current host port and `getContainerIp` as the fallback.
- `service.service` `updateService` does the same live resolution when publishing or editing a service route, so migrated domains register with the correct container-IP target instead of a missing loopback port.

## Test plan

- `apps/api/src/lib/upstream-url.test.ts` – unit tests for `buildUpstreamUrl` and `resolveUpstreamUrl`, including the loopback → container-IP fallback.
- `apps/api/src/modules/domains/routing-apply.service.test.ts` – tests that `applyProjectRouting` falls back to the container IP when the migrated container has no loopback host port, uses the live loopback port when present, and still falls back when live inspection fails.
- `bun run lint` (tsc --noEmit) passes.
- Targeted `bun vitest run src/modules/migration` passes.

### DIFF
--- a/apps/api/src/lib/upstream-url.test.ts
+++ b/apps/api/src/lib/upstream-url.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect, vi } from "vitest";
+import { buildUpstreamUrl, resolveRouteStrategy, resolveUpstreamUrl } from "./upstream-url";
+
+describe("buildUpstreamUrl", () => {
+  it("uses the loopback host port when one is provided", () => {
+    expect(
+      buildUpstreamUrl({
+        strategy: "loopback-port",
+        ip: "172.18.0.5",
+        hostPort: 4000,
+        containerPort: 3001,
+      }),
+    ).toBe("http://127.0.0.1:4000");
+  });
+
+  it("falls back to the container IP when no loopback host port is bound", () => {
+    expect(
+      buildUpstreamUrl({
+        strategy: "loopback-port",
+        ip: "172.18.0.5",
+        containerPort: 3001,
+      }),
+    ).toBe("http://172.18.0.5:3001");
+  });
+
+  it("returns null when neither loopback host port nor container IP is known", () => {
+    expect(
+      buildUpstreamUrl({
+        strategy: "loopback-port",
+        containerPort: 3001,
+      }),
+    ).toBeNull();
+  });
+
+  it("always uses the container IP for the explicit container-ip strategy", () => {
+    expect(
+      buildUpstreamUrl({
+        strategy: "container-ip",
+        ip: "172.18.0.5",
+        hostPort: 4000,
+        containerPort: 3001,
+      }),
+    ).toBe("http://172.18.0.5:3001");
+  });
+});
+
+describe("resolveRouteStrategy", () => {
+  it("resolves auto to loopback-port", () => {
+    expect(resolveRouteStrategy("auto")).toBe("loopback-port");
+  });
+
+  it("preserves the explicit container-ip strategy", () => {
+    expect(resolveRouteStrategy("container-ip")).toBe("container-ip");
+  });
+});
+
+describe("resolveUpstreamUrl", () => {
+  it("falls back to the container IP when the container has no loopback host port", async () => {
+    const runtime = {
+      supports: (cap: string) => cap === "containerIp",
+      getContainerIp: vi.fn().mockResolvedValue("172.18.0.5"),
+    };
+
+    const url = await resolveUpstreamUrl({
+      strategy: "loopback-port",
+      runtime,
+      containerId: "abc123",
+      containerPort: 3001,
+    });
+
+    expect(url).toBe("http://172.18.0.5:3001");
+    expect(runtime.getContainerIp).toHaveBeenCalledWith("abc123");
+  });
+
+  it("uses the provided loopback host port when one is bound", async () => {
+    const runtime = {
+      supports: () => false,
+      getContainerIp: vi.fn(),
+    };
+
+    const url = await resolveUpstreamUrl({
+      strategy: "loopback-port",
+      runtime,
+      containerId: "abc123",
+      containerPort: 3001,
+      hostPort: 4000,
+    });
+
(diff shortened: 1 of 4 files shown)


santhiprakash:
Closing — upstream main now resolves route upstreams from the live container (resolveLiveUpstreamUrl in apps/api/src/modules/domains/routing-apply.service.ts) so workloads without a published loopback port route at their container IP, which is what this PR implemented for #506.

PR #517, opened 2026-08-08
fix(deployments): protect carried-forward containers on compose deploy failure

### LINKED ISSUE #500: `onFailure` destroys carried-forward (still-live) containers when a compose deploy fails — a failed redeploy can take down the running app
## Summary

On a multi-service (compose) project, a deploy that **carries some services forward** (keeps their existing live containers) and then **fails** on another step tears down **every** container recorded on that deployment's service rows — including the carried ones, which belong to the **still-active previous release**. So a failed redeploy of one service destroys the currently-running app. The active-release pointer correctly does *not* advance (failure never touches the project row), so the dashboard still shows the previous release as active/ready **while its container is gone and the site is down**. Verified on current `main`.

## Why it matters

This is an **outage**, not a cosmetic drift: a failed deploy takes the live app down. Deletion / rejection / cancellation already protect carried containers with a keep-set (`collectDeploymentManifest` + `protectRetained`); `onFailure` is the one teardown path that doesn't. The most-exposed shape is exactly the carry-forward-heavy one — app-template / image-only stacks where most services are unchanged between deploys.

*(Found while working on #498. It's a separate axis — #498 is record↔container drift on **success**; this is unguarded teardown on **failure** — so filing it on its own.)*

## Repro

1. A compose project with two services (e.g. an app + a sidecar, or app + db).
2. Deploy it; confirm both are Live.
3. Redeploy such that one service is **unchanged** (carried forward — same image/env/ports) and another **fails** to deploy (e.g. a bad image ref, a failing build, or a port conflict on the changed one).
4. Observe: the deploy is marked `failed`, and the **carried, previously-healthy service's container is destroyed** along with the failed one. The project still points at the previous (now container-less) release.

## Root cause (current `main`)

1. **The carry branch records the live container on the failing deployment's rows.** For an unchanged service, the loop upserts a service-deployment row for the *new* deployment with `containerId: carried.containerId` — the container from the previous/active release (`apps/api/src/modules/deployments/compose/deploy.service.ts`, carry-forward branch).
2. **A later failure reaches `onFailure`.** A subsequent service that fails to deploy `throw`s (`deploy.service.ts:1522`, `Failed to deploy service "…"`); the pipeline's outer catch runs `onFailure`.
3. **`onFailure` destroys every service-row container with no keep-set.** It lists all service rows for the deployment and calls `runtime.destroy(serviceDep.containerId)` on each (`apps/api/src/modules/deployments/deployment-lifecycle.ts:311`, the `for (const serviceDep of serviceDeps)` loop ~`:340`) — with no protection for containers owned by the still-active release. Contrast `deleteDeployment` / `rejectDeployment` / `cancelBuildSession`, which pass the manifest through `protectRetained`. (The file already knows this teardown is dangerous — see the `settled`-flag comment at `:74`: re-reporting a working deploy through `onFailure` "*inverts a working deploy and tears down its containers*" — but that guard only prevents *double*-reporting after success; it doesn't protect carried containers on a *genuine* failure.)

## Impact

- **A failed redeploy destroys the running app.** The active pointer correctly stays on the previous release (good), but that release's container has been torn down — so the dashboard reads active/ready while the site is down, and recovery needs a full redeploy.
- Worst for **carry-forward-heavy stacks** (app-template / image-only services), which is the common multi-service shape.

## Fix direction

`onFailure` should not destroy a container owned by the **still-active** deployment. Compute a keep-set from the active release's referenced containers — the same protection `deleteDeployment` / `rejectDeployment` already use (`collectDeploymentManifest` + `protectRetained`) — and skip those in the teardown loop; or only destroy containers this deployment actually **created**, never carried. Either way, a failed deploy leaves the live release untouched.

Happy to open a PR once there's agreement on which of those you'd prefer.

santhiprakash:
## Summary

Fixes #500 — `onFailure` no longer destroys containers carried verbatim from the still-live previous release when a multi-service compose deploy fails partway through.

## Problem

On a compose project, a redeploy that carries unchanged services forward and then fails on a later step runs `onFailure`, which iterated every service row on the failed deployment and called `runtime.destroy` on each `containerId`. Carried-forward rows name the **same** container as the active release, so a failed redeploy took the live app down while `activeDeploymentId` still pointed at the previous release.

## Triage / Root cause

`onFailure` in `deployment-lifecycle.ts` used a naive destroy loop over `repos.service.listByDeployment(dep.id)`. Reject, delete, and build-cancel teardown paths already consult `collectDeploymentManifest(..., { protectRetained: true })` to skip artifacts a live or retained release still references; the failure path was the lone outlier.

## Fix

- Replace the per-row `runtime.destroy` loop with `collectDeploymentManifest` + `executeCleanup` using `protectRetained: true`, matching cancel/reject/delete.
- Add `on-failure-keep-set.test.ts` regression coverage modeled on `reject-keep-set.test.ts`.

## Verification

```bash
cd apps/api && bun run test test/modules/deployments/on-failure-keep-set.test.ts test/modules/deployments/reject-keep-set.test.ts
```

Result: 9 tests passed.

## Notes / Risks

- Scope is limited to the failure teardown path; `onCancelled` still uses the old destroy loop (separate issue if needed).
- `provisioned.imageRef` cleanup at the top of `onFailure` is unchanged; compose pipeline already cleans built service images before calling `onFailure`.

### DIFF
--- a/apps/api/src/modules/deployments/deployment-lifecycle.ts
+++ b/apps/api/src/modules/deployments/deployment-lifecycle.ts
@@ -27,6 +27,11 @@ import { failureStatusFor } from "./blocking-errors";
 import { sanitizeStorableStrings, sliceWithoutSplittingPair } from "./build-log-sanitize";
 import { detectAndStoreFavicon } from "../../lib/favicon-detector";
 import { onWebmailDeployed } from "../mail/webmail/webmail-install.service";
+import {
+  collectDeploymentManifest,
+  executeCleanup,
+  type CleanupManifest,
+} from "../projects/project-cleanup.service";
 
 /**
  * The "your domains didn't route" line for a deploy that otherwise succeeded.
@@ -432,19 +437,17 @@ export async function onFailure(
     }
   }
 
-  if (runtime) {
-    const serviceDeps = await repos.service.listByDeployment(dep.id).catch(() => []);
-    for (const serviceDep of serviceDeps) {
-      if (!serviceDep.containerId) continue;
-      try {
-        await runtime.destroy(serviceDep.containerId);
-      } catch (destroyErr) {
-        console.error(
-          `[DEPLOY] Failed to destroy service container ${serviceDep.containerId} on failure:`,
-          destroyErr,
-        );
-      }
-    }
+  // protectRetained: a failed compose redeploy carries the LIVE release's
+  // containerId/imageRef onto its own service rows for every service it hadn't
+  // replaced yet — same seam as reject/delete/cancel. activeDeploymentId never
+  // advances on failure, so the keep set sees the live release without a hint.
+  const manifest = await collectDeploymentManifest(dep, project, {
+    protectRetained: true,
+  }).catch((): CleanupManifest => ({ projectId: dep.projectId, resources: [] }));
+  if (manifest.resources.length > 0) {
+    await executeCleanup(manifest).catch((err) => {
+      console.error(`[DEPLOY] Cleanup crashed for ${dep.id} on failure:`, err);
+    });
   }
 
   // INVARIANT: failure writes the DEPLOYMENT row only — NEVER the project row.
--- a/apps/api/test/modules/deployments/on-failure-keep-set.test.ts
+++ b/apps/api/test/modules/deployments/on-failure-keep-set.test.ts
@@ -1,0 +1,116 @@
+/**
+ * D3 follow-on: a failed compose redeploy destroyed carried-forward containers.
+ *
+ * `onFailure` tears down a failed release's runtime — but a compose service that
+ * didn't change carries its `containerId` and `imageRef` verbatim onto the failed
+ * release's row (the `carried` seam in compose/deploy.service.ts). The old path
+ * called `runtime.destroy` on every service row, so a deploy that carried `db`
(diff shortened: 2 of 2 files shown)


santhiprakash:
Rebased onto current main and re-ran the tests locally: `on-failure-keep-set.test.ts` (3 passed) and `reject-keep-set.test.ts` (6 passed).

The Typecheck/Test CI failures on this PR are pre-existing on main itself — `apps/api/src/lib/auth.ts` at lines 418–428 contains unresolved merge conflict markers (from the `patch v0.6.6` commit, `a52e2566`), and the same checks fail on upstream `main` right now. This PR doesn't touch that file.

Hydralerne:
Reviewed against current main after #891 merged. The bug in #500 is already fixed by `cleanupOwnedServiceContainers`, shared by failure and cancellation. It protects containers referenced by the live/retained releases and refuses destructive cleanup if the keep set cannot be computed.

This PR’s separate manifest path would duplicate existing protection and change which resources are cleaned during failure. I verified the shared lifecycle and retention paths with 20 relevant tests; the original failed-deployment behavior no longer exists on main.

Thank you for identifying the carried-container hazard and proposing a fix. Closing this PR as superseded by the existing main implementation, with #500 accounted for in the #892 review ledger.

PR #518, opened 2026-08-08
fix: report and repair a missing container→host SSH channel

### LINKED ISSUE #509: Docker app deployment fails because the host channel is not provisioned
### Summary

Deploying a Docker app fails when OpenShip itself runs in a container because the host execution channel has not been provisioned.

### Error

This operation targets the HOST machine, but no host channel is configured (OPENSHIP_HOST_SSH_HOST is unset) and Openship is running in a container — so it would have run inside the container instead, against the wrong filesystem. Re-run `openship up` to provision the host channel.

Image

### Steps to reproduce

1. Install and run OpenShip using Docker.
2. Create a Next.js application.
3. Start its deployment.
4. The deployment fails with the error above.

### Expected behavior

`openship up` should automatically provision and persist a working host channel for containerized installations.

Docker applications should deploy without requiring additional undocumented configuration.

### Actual behavior

`OPENSHIP_HOST_SSH_HOST` is unset, so OpenShip cannot safely execute operations against the host filesystem. The problem is only discovered when deploying an application.

### Suggested improvements

- Provision the host channel automatically during `openship up`.
- Validate the host channel before completing installation.
- Provide a health check for host execution.
- Document how to repair or reprovision the channel on an existing installation.
- Include a verification command in the error message.

Rish-it:
## Summary

A containerized install could come up with no container→host SSH channel, report success, and only fail at the first deploy — `This operation targets the HOST machine, but no host channel is configured (OPENSHIP_HOST_SSH_HOST is unset)`. `openship up` now says why it couldn't provision the channel, `openship doctor` reports whether an install has one, and the raw `docker compose` stack finally has somewhere to mount the key its `.env.example` tells you to configure.

## Motivation

**`openship up` swallowed every reason it couldn't provision.** `provisionHostSshChannel` is one `try { … } catch { return null }`, and `null` is also what the DELIBERATE absences return (`--no-host-control`, a non-Linux box). `renderEnv` writes the `OPENSHIP_HOST_SSH_*` keys only `if (host)`, so any failure produced an `.env` with no channel, an install that printed `✔ Openship is running via Docker Compose`, and a deploy that died later with the error above. A missing `ssh-keygen` was enough to get there.

Worse, the remedy that error names is a no-op for exactly this case: re-running `openship up` on such a box provisions nothing again, exit 0, no output. Nothing else surfaced it either — `openship doctor` reported all-green, and `probeReachable` (`ssh-manager.ts:648`) calls `recordSuccess` when the var is unset, so the box shows Online right up to the deploy.

**The raw compose stack could never have a channel.** `docker/docker-compose.yml` (README: "Self-host with raw Docker Compose (no CLI)") sets `extra_hosts: host.docker.internal:host-gateway` but mounts no key, while `.env.example` tells that install to set `OPENSHIP_HOST_SSH_KEY=/run/secrets/openship_host_key` — a path nothing ever put a file at. That stack fails every host operation regardless of its `.env`.

Deploy is where it surfaces because a "This Server" deploy resolves through `resolveServerExecutor` → `acquireHostChannel`, and the host-port allocation in `build-pipeline.ts:1580` goes through `sshManager.withHostExecutor` — both before any Docker work.

## Related issue

Fixes #509

## Changes

- `apps/cli/src/lib/compose.ts` — `provisionHostSshChannel` returns `{ channel, error }`: `error` is null for the deliberate absences and set for everything else (ENOENT is reported as "ssh-keygen isn't installed (package: openssh-client)", not a generic failure). `composeUp` prints it once the stack is up — after the pull output, not buried in it. The install still succeeds; a box that only manages remote servers is fine without a channel. Adds `composeHostChannel()`: the same question asked of an install that already exists (`.env` + the key on disk), shaped as a `doctor` check.
- `apps/cli/src/lib/repair.ts` — `openship doctor` gains a `Host` row, compose installs only (a bare install is already on the host). Fails when the channel is missing or its key is gone; passes when it's there or host control was deliberately turned off. Lands in the panel, the one-shot report and `--json` through the existing `ComponentCheck` pipeline.
- `docker/docker-compose.yml` — mount the host key from `OPENSHIP_HOST_KEY_PATH`, `/dev/null` when unset, matching the compose file `openship up` generates.
- `.env.example` — name `OPENSHIP_HOST_KEY_PATH` alongside the vars that depend on it.
- `packages/adapters/src/system/executor.ts` — the error points at `openship doctor` before `openship up`, since the run that produced it may already have tried.
- `apps/cli/test/unit/compose-host-channel.test.ts` — new; provisioning reported vs. deliberate absence stays quiet, and each state `doctor` reports.

## Verification

Reproduced end to end first, on a real Linux host rather than a mock: a privileged container running its own `dockerd` and `sshd`, driving the real CLI through `openship up --compose` with real image pulls (postgres, redis, api, dashboard, edge).

**A clean box provisions correctly** — this is not broken in general:

```
OPENSHIP_HOST_SSH_HOST=host.docker.internal   # in .env AND in openship-api-1
host.docker.internal → 172.18.0.1             # inside the from="172.16.0.0/12,…" grant
ssh -i key root@host.docker.internal          → HOST_CHANNEL_OK
```

**Remove `ssh-keygen`, re-run** — `main` reports a healthy install with no channel, and the API then returns the reported error verbatim:

```
✔ Openship is running via Docker Compose.       exit=0
grep -c HOST_SSH ~/.openship/compose/.env   →  0
docker exec openship-api-1 env | grep -c HOST_SSH  →  0

POST /api/system/self-edge/preflight
{"error":"This operation targets the HOST machine, but no host channel is configured
 (OPENSHIP_HOST_SSH_HOST is unset) and Openship is running in a container — …"}
```

**Same box, this branch:**

```
  ! Host operations are NOT available: ssh-keygen isn't installed (package: openssh-client)
    Deploys to this box, the :80/:443 takeover and the host terminal will fail
    with "no host channel is configured".
    Fix the cause, re-run `openship up`, then confirm with `openship doctor`.
✔ Openship is running via Docker Compose.

$ openship doctor   →  ✗ Host   not provisioned — deploys to this box will fail; re-run `openship up`
$ echo $?           →  1
```

Restore `ssh-keygen`, re-run `openship up`: no warning, `✓ Host  SSH to host.docker.internal as root`, and the call that returned the error now returns `{"status":{"classification":"ours",…}}`.

**The raw compose stack**, brought up for real from `docker/docker-compose.yml` (postgres + redis + api) with the key the CLI provisioned:

```
mount:  /root/.openship/compose/host-ssh/id_ed25519 -> /run/secrets/openship_host_key
POST /api/system/self-edge/preflight  →  {"status":{"classification":"free",…}}
```

Negative control on the same stack with `OPENSHIP_HOST_KEY_PATH` unset, so the mount falls back to `/dev/null` (the pre-fix state — key named in `.env`, nothing mounted): the api still boots healthy, and the host operation fails with `SSH requires one of privateKey, sshAgent, or password.` So the mount is load-bearing and its default is harmless.

The regression test fails with only `compose.ts` reverted — 6 of 8, with the two negative controls staying green, so the warning isn't unconditional:

```bash
$ npx vitest run test/unit/compose-host-channel.test.ts   # fix reverted, tests kept
      Tests  6 failed | 2 passed (8)
$ npx vitest run test/unit/compose-host-channel.test.ts   # fix restored
      Tests  8 passed (8)
```

Suites, run twice each and green both times:

```bash
$ bun run test --force
 Tasks:    7 successful, 7 total          # 0 cached, both runs
$ cd apps/cli && npx tsc --noEmit                              # clean
$ cd packages/adapters && npx tsc --noEmit -p tsconfig.json    # clean
```

Every commit on the branch was checked individually — 233 CLI tests passing at each, none red.

Two pre-existing failures on `main` that this branch neither touches nor fixes: `@repo/dashboard#lint` (stale `.next` generated types against the installed Next — fails identically on a clean `main`), and `prettier --check` on `compose.ts`, `repair.ts` and `docker-compose.yml`, all three of which are already unformatted on `main`. Running `bun format` would rewrite unrelated lines in them; the one new file is prettier-clean.

Deliberately out of scope, both worth their own change: `up --dry-run` still previews a channel it may be unable to create (`composePlan` calls the pure planner, never the provisioner), and a local Docker deploy still requires the channel for host-port allocation even though the build and run go over the mounted socket.

## Checklist

- [ ] `bun run test`, `bun run --cwd <workspace> lint`, and `bun format` all pass locally — `bun run test` passes; `@repo/dashboard#lint` and `bun format` are pre-existing failures on `main`, see Verification

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -114,10 +114,13 @@ INTERNAL_TOKEN=change-me-32-byte-random-hex
 # the host over SSH via host.docker.internal (internal bridge, not the public IP).
 # `openship up` provisions the key + these vars automatically; set them by hand
 # only for a raw `docker compose` install that needs host ops.
+# OPENSHIP_HOST_KEY_PATH is the key on the HOST (mounted to OPENSHIP_HOST_SSH_KEY
+# in the container); its public half goes in that user's ~/.ssh/authorized_keys.
 # OPENSHIP_HOST_SSH_HOST=host.docker.internal
 # OPENSHIP_HOST_SSH_USER=root
 # OPENSHIP_HOST_SSH_PORT=22
 # OPENSHIP_HOST_SSH_KEY=/run/secrets/openship_host_key
+# OPENSHIP_HOST_KEY_PATH=/root/.openship/compose/host-ssh/id_ed25519
 
 # ─── OAuth login (optional) ───────────────────────────────
 # GITHUB_CLIENT_ID=
--- a/apps/cli/src/lib/compose.ts
+++ b/apps/cli/src/lib/compose.ts
@@ -543,9 +543,21 @@ function plannedHostChannel(hostControl: boolean): { user: string; keyPath: stri
   return { user, keyPath: join(COMPOSE_DIR, "host-ssh", "id_ed25519") };
 }
 
-function provisionHostSshChannel(hostControl: boolean): { user: string; keyPath: string } | null {
+/**
+ * Provision the channel `plannedHostChannel` names, or say why it couldn't.
+ *
+ * `error` is null for the DELIBERATE absences (`--no-host-control`, a non-Linux
+ * box) and set for everything else. Collapsing the two into a bare `null` is what
+ * let `renderEnv` write an `.env` with no OPENSHIP_HOST_SSH_* keys, report a
+ * healthy install, and leave the first deploy to fail with "no host channel is
+ * configured" (#509).
+ */
+function provisionHostSshChannel(hostControl: boolean): {
+  channel: { user: string; keyPath: string } | null;
+  error: string | null;
+} {
   const target = plannedHostChannel(hostControl);
-  if (!target) return null;
+  if (!target) return { channel: null, error: null };
   const { user, keyPath } = target;
   try {
     mkdirSync(join(COMPOSE_DIR, "host-ssh"), { recursive: true, mode: 0o700 });
@@ -555,10 +567,19 @@ function provisionHostSshChannel(hostControl: boolean): { user: string; keyPath:
         ["-t", "ed25519", "-N", "", "-q", "-f", keyPath, "-C", HOST_KEY_COMMENT],
         { stdio: "ignore" },
       );
-      if (g.status !== 0) return null;
+      // `g.error` is ENOENT — a box with no openssh-client, which is a package
+      // away rather than a broken install.
+      if (g.error || g.status !== 0) {
+        return {
+          channel: null,
+          error: g.error
+            ? "ssh-keygen isn't installed (package: openssh-client)"
+            : `ssh-keygen failed (exit ${g.status})`,
+        };
+      }
     }
     const pub = readFileSync(`${keyPath}.pub`, "utf8").trim();
-    if (!pub) return null;
+    if (!pub) return { channel: null, error: `${keyPath}.pub is empty — delete the key pair and re-run` };
 
     // Authorize the key for the host user the container SSHes in as (see
     // plannedHostChannel for why that user comes from the passwd entry).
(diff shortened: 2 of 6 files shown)


Hydralerne:
Closing after comparison with current main (`4e66349c`): these paths are now covered by the shared host-channel provisioning, preflight and doctor logic. Main also preserves a previous channel on failed reprovisioning, performs a real authentication probe rather than checking only env/key presence, and propagates prefetch environment changes so the running API is recreated with them. Local Docker/edge work uses the socket and shared mounts independently of the host channel.

I verified the related 188 CLI and 47 adapter/API tests. Integrating the older parallel provisioning/doctor implementation would duplicate and weaken those paths; no additional fix from this PR is needed for #509. The contribution and report remain linked here.

PR #519, opened 2026-08-08
fix(backups): bound pg_dump exec and reconcile stale in-flight runs

### LINKED ISSUE #516: Backup runs hang forever in uploading/queued — no timeout or staleness reconciliation; only an API restart clears them
## Summary

Managed service backups (`sourceKind: service`, `payloadKind: auto`) never complete for a `postgres:16` compose service on a self-hosted v0.6.1 install. Runs transition to `uploading` ~1 second after start with `bytesTransferred` left null, then hang indefinitely — no timeout, no error, no reconciliation. The only thing that ever gives them a terminal state is an API restart: the boot sweep marks every in-flight run as `server_error` with `"API restart while backup in flight"`. Once a run is stuck, later runs stop being processed altogether (a fresh manual run stayed in `queued` indefinitely), which suggests the in-process runner's worker is blocked by the hung run.

## Environment

- OpenShip self-hosted, v0.6.1 (`ghcr.io/oblien/openship-api:0.6.1`, `openship-dashboard:0.6.1`, `openship-edge:0.6.1`)
- API reaches the deployment host's Docker daemon via dockerode-over-SSH (remote server)
- Source: `postgres:16` compose service (container running, healthcheck passes)
- Destination: Cloudflare R2 (S3-compatible); preflight (Put+Head+Delete) passes — this is a hang, not a credential/config failure

## Steps to reproduce

1. Create a backup policy (`payloadKind: auto`) targeting a `postgres:16` compose service, destination S3/R2.
2. Trigger a backup (scheduled or manual).
3. The run flips to `uploading` ~1s after start and stays there; `bytesTransferred` stays null; no `errorMessage`.
4. Trigger another backup while the first is stuck — it stays in `queued` and is never picked up.

## Actual behavior

5 runs on the same policy, 0 succeeded:

| run id | trigger | started (UTC) | status |
|---|---|---|---|
| `bkr_23103591-9509-4684-8cb4-c643aeeffce9` | manual | 2026-08-06T13:03:04Z | `server_error` — `"API restart while backup in flight"`, only after API restart |
| `bkr_e27385ca-37eb-4902-bbdb-3667c4914653` | cron | 2026-08-07T03:17:00Z | `server_error` — same |
| `bkr_2f6f8830-6ab4-4f2d-a740-6d3858293347` | manual | 2026-08-07T14:38:13Z | `uploading` — stuck, `bytesTransferred` null |
| `bkr_3b4f4f64-5636-4fc4-bf78-ea0019134670` | cron | 2026-08-08T03:17:00Z | `uploading` — stuck, `bytesTransferred` null |
| `bkr_b0f252f5-8d00-4ad5-82a3-e27b83199f57` | manual | 2026-08-08T17:18:35Z | `queued` — never picked up |

The first two runs only reached a terminal state because the API restarted ~11 hours later; without that restart they would still be in `uploading`.

## Expected behavior

A run that produces no bytes should fail (or be reconciled) within a bounded time with a surfaced error. A stuck run must not block the runner or prevent later runs from being processed. Runs stuck in non-terminal states should be reconciled by a periodic sweep, not only at API boot.

## Root cause (v0.6.1)

- The pg-dump producer streams via `docker.execStream` with no `timeoutMs` (packages/adapters/src/backup/producers/pg-dump.ts), and the docker exec timer only arms when `timeoutMs` is passed (packages/adapters/src/backup/executors/docker.ts). Wrong destination credentials fail fast at preflight (packages/adapters/src/backup/destinations/s3.ts), so this is a hang, not a config error.
- `lastEventAt` is written on transitions but never read; the only reconciliation of in-flight runs is the boot sweep (packages/db/src/repos/backup.repo.ts).
- `retention-prune` only selects `succeeded` runs (apps/api/src/modules/backups/retention-prune.ts), so stuck runs accumulate forever.
- There is no backup-run cancel endpoint.

## Precedent

#434 (restore stuck in `applying`, only cleared on restart) was fixed by #435, which added an idle watchdog (10 min) and a 6-hour ceiling — but only for the volume producer. The pg-dump path did not get the same protection (see apps/api/src/modules/backups/PENDING.md).

## Proposed approach

- Add a timeout to the pg-dump `execStream` and to the destination `put`, and/or an idle watchdog mirroring the volume producer's.
- Add a periodic staleness sweep (not only at boot) that fails runs stuck in non-terminal states past a deadline and surfaces the reason.
- Ensure a stuck run cannot hold the worker so later runs are still processed.

santhiprakash:
Fixes #516.

## Problem

Managed-service `pg_dump` backups transition to `uploading` almost immediately, then hang with `bytesTransferred` null. There is no idle watchdog on `docker.execStream` (unlike `streamPath`/`receiveStream` after #434), and the only reconciliation for in-flight runs is the boot sweep. A wedged run holds a worker slot, so later runs stay `queued` indefinitely until an API restart.

## Triage / Root cause

- `DockerBackupExecutor.execStream` → `attachDemuxed` only armed a wall-clock timer when the caller passed `timeoutMs`; `pg_dump` passes nothing, so a stuck `pg_dump | zstd` pipe never fails.
- `backup_run.lastEventAt` is bumped on FSM transitions but never read between restarts.
- The in-process runner caps concurrency at 2; hung `processRun` calls occupy slots until the exec resolves.

## Fix

- **`attachDemuxed`**: same bounds as capture/restore — 10-minute idle watchdog (fed by mux stream traffic) plus 6-hour absolute ceiling by default. Hooks that pass an explicit `timeoutMs` still honor it as the ceiling.
- **`sweepRunsWithStaleHeartbeat`**: selective DB reconciliation — queued >30m, preparing/snapshotting/verifying idle >10m, uploading with zero bytes idle >10m, any in-flight past 6h.
- **System job** `backups:stale-run-sweep` every 5 minutes so wedged rows clear without an API restart.

## Verification

- `bunx vitest run packages/adapters/test/backup-exec-stream-timeout.test.ts` — 4/4 passing
- `bunx vitest run apps/api/test/modules/backups/backup-stale-sweep.test.ts` — 3/3 passing
- `python3 ../scripts/preflight_ship.py --repo oblien/openship --local . --branch main --file packages/adapters/src/backup/executors/docker.ts --must-contain 'const watchdog = createIdleWatchdog' --must-not-contain 'sweepRunsWithStaleHeartbeat' --issue 516` — clear

## Notes / Risks

- Uploading runs that have already recorded `bytesTransferred > 0` rely on the 6-hour ceiling only (lastEventAt is not bumped per chunk). That matches how long honest large uploads are allowed on the capture path.
- Does not add a backup-run cancel endpoint (called out in #516 as a follow-up).

### DIFF
--- a/apps/api/src/modules/backups/backup-stale-sweep.ts
+++ b/apps/api/src/modules/backups/backup-stale-sweep.ts
@@ -1,0 +1,39 @@
+/**
+ * Periodic reconciliation for backup runs stuck in non-terminal states.
+ *
+ * Boot already calls `sweepStaleRuns` for every in-flight row after a crash,
+ * but a wedged `pg_dump` exec (#516) can hang indefinitely without restarting
+ * the API — `lastEventAt` stops moving, queued siblings never get a worker
+ * slot, and the operator only sees relief after a manual restart.
+ *
+ * Thresholds mirror the docker exec defaults: 10 minutes of silence, 6 hours
+ * absolute, plus a separate queued-or-orphan window so a run that never left
+ * `queued` is surfaced even when concurrency is saturated.
+ */
+
+import { repos } from "@repo/db";
+import type { JobSummary } from "../../lib/system-jobs";
+
+/** Matches `DEFAULT_HELPER_IDLE_MS` in the docker backup executor. */
+export const BACKUP_RUN_IDLE_STALE_MS = 10 * 60 * 1000;
+/** Matches `DEFAULT_HELPER_TIMEOUT_MS` in the docker backup executor. */
+export const BACKUP_RUN_CEILING_STALE_MS = 6 * 60 * 60 * 1000;
+/** A queued row should be picked up within one poll cycle + concurrency wait. */
+export const BACKUP_RUN_QUEUED_STALE_MS = 30 * 60 * 1000;
+
+const STALE_REASON =
+  "Backup run made no progress within the configured timeout and was reconciled.";
+
+export async function runBackupStaleSweep(): Promise<JobSummary> {
+  const now = Date.now();
+  const swept = await repos.backupRun.sweepRunsWithStaleHeartbeat({
+    queuedCutoff: new Date(now - BACKUP_RUN_QUEUED_STALE_MS),
+    idleCutoff: new Date(now - BACKUP_RUN_IDLE_STALE_MS),
+    ceilingCutoff: new Date(now - BACKUP_RUN_CEILING_STALE_MS),
+    reason: STALE_REASON,
+  });
+  if (swept > 0) {
+    console.warn(`[backup-stale-sweep] reconciled ${swept} stale backup run(s)`);
+  }
+  return { swept };
+}
--- a/apps/api/src/modules/jobs/job.registry.ts
+++ b/apps/api/src/modules/jobs/job.registry.ts
@@ -19,6 +19,7 @@ import { platform } from "../../lib/controller-helpers";
 import { renewExpiringCerts } from "../../lib/ssl-scheduler";
 import { runOrphanSweep } from "../projects/orphan-gc-schedule";
 import { runRetentionSweep } from "../backups/retention-prune";
+import { runBackupStaleSweep } from "../backups/backup-stale-sweep";
 import { pruneAuditEvents } from "../audit/audit-prune";
 import { runReconcileSweep } from "../deployments/reconcile-schedule";
 import { runImageGcSweep } from "../deployments/image-gc";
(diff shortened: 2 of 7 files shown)


Hydralerne:
Thanks @santhiprakash — merged as part of #520, your commit `0a737155` landed intact. I added one follow-up on top: the periodic sweep's `uploading` idle branch was reaped honest large single-artifact dumps (bytesTransferred stays NULL and lastEventAt is frozen for the whole upload, since the orchestrator only writes progress at artifact boundaries), which is the exact managed-postgres case #516 is about. That branch is removed; wedged uploads are still reaped in-process by the executor watchdog you added, and the DB sweep backstops `uploading` via the 6h ceiling. Appreciate the fix! 🙏

PR #524, opened 2026-08-09
fix(backups): close exec-stream sinks on EOF so finished dumps don't hang forever

Jcorderop02:
## Summary

Command-based Docker backups (`pg_dump`, `mysqldump`, `mongodump`, custom hooks) could finish writing 100% of their bytes and then hang forever in `uploading`, because the exec attach stream's EOF never reached the destination writer.

## Related issue

None. Self-contained bug fix, no filed issue.

## Problem

An SFTP-destination backup run for a small (~1.7MB) Postgres database sat in `uploading` indefinitely. The `.uploading-*` temp file on the destination stopped growing at a fixed, fully-correct byte count and never got renamed to its final name. No error, no progress, no completion.

## Triage / Root cause

Instrumented the SFTP write stream during a live stuck run to log the raw `ssh2` SFTP channel state (outgoing window, buffered payloads) every 5s. At the moment of the stall the channel window was >2MB free and nothing was buffered, so the SSH/SFTP layer was completely idle. That ruled out a network or `ssh2` backpressure issue.

Reproduced the same `pg_dump | zstd` command directly via the `docker exec` CLI, bypassing Openship entirely: it completed in under half a second, producing a file the same size as where the "stall" happened. So the dump itself was never stalling mid-transfer. It was already fully done, and something downstream just never found out.

`DockerBackupExecutor.execStream` calls `attachDemuxed`, which pipes the exec attach stream through `docker.modem.demuxStream`. `demuxStream` only forwards `data` events; it never calls `.end()` on the destination sinks when the source stream ends. `attachDemuxed` never worked around this, so `stdout` (which becomes the backup artifact stream, piped straight into the destination writer) never received an `end` event once the exec process exited. The destination writer had nothing left to wait for and no signal to stop waiting.

The sibling function a bit lower in the same file, `demuxContainerStream` (used for volume/tar backups), already has this exact fix, with a comment explaining why it's needed. It just never got applied to the exec-based path.

## Fix

- `attachDemuxed`: end `stdout`/`stderrSink` when the source attach stream emits `end` or `close`, mirroring `demuxContainerStream`. Added a backstop that force-closes the sinks once the exec exit code is known, for connections (e.g. Docker reached over an SSH tunnel, as with a remotely-managed server) where `end`/`close` isn't always delivered promptly even though the process has already exited.
- `sftp.ts`: added a 2-minute idle watchdog around the SFTP write itself, so any other future cause of a stall (unrelated to this bug) surfaces as a clear, retryable error instead of hanging the run, and the worker slot behind it, forever.

## Verification

Reproduced against a live SFTP destination before and after the fix, same ~1.7MB Postgres dump:

```
Before: backup run status stuck in "uploading", errorMessage after the new
        watchdog's 2-minute timeout:
        "SFTP upload stalled: no write progress for 120s (wrote 1699620
        bytes so far)."

After:  backup run status "succeeded"
        startedAt  2026-08-09T10:39:05.439Z
        finishedAt 2026-08-09T10:39:08.733Z
        bytesTransferred: 1699619
```

Added a regression test in `backup-exec-stream-timeout.test.ts` asserting `stdout` emits `end` once the fake attach stream finishes (the mock's `demuxStream` reproduces the real bug: it only forwards `data`, same as `docker-modem`). Confirmed this test fails on the pre-fix code and passes with the fix, by toggling `packages/adapters/src/backup/executors/docker.ts` with `git stash` and re-running:

```
$ bun run test -- backup-exec-stream-timeout
 ✓ test/backup-exec-stream-timeout.test.ts (5 tests) 79ms
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

Full `packages/adapters` suite:

```
$ bun run test
 Test Files  106 passed | 1 failed (107)
      Tests  1046 passed | 1 failed (1047)
```

The one remaining failure (`remote-journal.test.ts`) is pre-existing and unrelated: reproduced identically on a clean `upstream/main` checkout with none of this PR's changes applied.

Note: on this machine, `bun run test` only collects correctly after a local, uncommitted `ssr.noExternal: ["zod"]` tweak to `packages/adapters/vitest.config.ts`. Without it, `import { z } from "zod"` resolves to a build missing the `z` export under Vite's SSR transform, unrelated to this PR (reproduced with a trivial standalone test and with this PR's changes fully reverted). Not included in this diff since it's a separate, unrelated environment issue; happy to open it as its own PR if useful.

## Notes / Risks

- Scoped to the exec-based backup path (`attachDemuxed`). The volume/tar path (`demuxContainerStream`) already had this fix and is untouched.
- The idle watchdog in `sftp.ts` is a defense-in-depth addition, not the root-cause fix. It would have converted the original hang into a clear error, but the actual fix is the `attachDemuxed` change above.

## Checklist

- [x] One change per PR: this PR is the exec-stream EOF fix only; an unrelated dashboard i18n fix found during the same investigation is #525
- [x] The diff is scoped: no reformatting or lint fixes on lines I wasn't otherwise changing
- [x] `bun run test` passes (1046 of 1047; the one unrelated pre-existing failure is noted above)

### DIFF
--- a/packages/adapters/src/backup/destinations/sftp.ts
+++ b/packages/adapters/src/backup/destinations/sftp.ts
@@ -212,15 +212,52 @@ class SftpDestinationImpl implements BackupDestination {
 
       await new Promise<void>((resolve, reject) => {
         const ws = sftp.createWriteStream(tmp);
-        ws.on("error", reject);
-        ws.on("close", () => resolve());
+        let lastProgressAt = Date.now();
+        let settled = false;
+
+        // #openship-sftp-stall: ssh2's SFTP write stream has a known
+        // backpressure/drain bug where pipe() can wait forever for a
+        // 'drain' event that never fires once its internal request queue
+        // saturates — the upload silently freezes mid-transfer with no
+        // error and no close. Observed in production: a `.uploading-*`
+        // temp file stuck at a fixed byte size indefinitely. Bound the
+        // whole operation with an idle watchdog (mirrors the idle/ceiling
+        // pattern already used for docker exec streams, see #516) so a
+        // stall surfaces as a clear, retryable error instead of hanging
+        // the run (and the worker slot behind it) forever.
+        const STALL_TIMEOUT_MS = 2 * 60 * 1000;
+        const watchdog = setInterval(() => {
+          if (!settled && Date.now() - lastProgressAt > STALL_TIMEOUT_MS) {
+            finish(
+              new Error(
+                `SFTP upload stalled: no write progress for ${STALL_TIMEOUT_MS / 1000}s ` +
+                  `(wrote ${bytesWritten} bytes so far). The destination's SFTP write stream ` +
+                  `stopped draining — retry, or check the destination server's SFTP subsystem.`,
+              ),
+            );
+          }
+        }, 10_000);
+
+        const finish = (err?: Error) => {
+          if (settled) return;
+          settled = true;
+          clearInterval(watchdog);
+          if (err) {
+            ws.destroy();
+            body.destroy();
+            reject(err);
+          } else {
+            resolve();
+          }
+        };
+
+        ws.on("error", (err) => finish(err));
+        ws.on("close", () => finish());
         body.on("data", (chunk: Buffer) => {
           bytesWritten += chunk.byteLength;
+          lastProgressAt = Date.now();
         });
-        body.on("error", (err) => {
-          ws.destroy();
-          reject(err);
-        });
+        body.on("error", (err) => finish(err));
         body.pipe(ws);
       });
 
--- a/packages/adapters/src/backup/destinations/sftp.ts
+++ b/packages/adapters/src/backup/destinations/sftp.ts
@@ -44,6 +44,21 @@ const CAPS: ReadonlySet<DestinationCapability> = new Set<DestinationCapability>(
   "streamingGet",
 ]);
 
+/**
+ * Idle ceiling for one artifact upload — silence, not wall clock, for the same
+ * reason capture and restore bound on idle: an honest multi-hour artifact must
+ * not be strangled, and a wedged one must not be forever.
+ *
+ * Must never be TIGHTER than the producer's own idle allowance (execStream's
+ * 10-minute default). `put`'s body is the dump stream itself, so from here a
+ * source that has gone quiet — pg_dump waiting on a lock, a cold-cache table
+ * scan, a user's custom command — is indistinguishable from a dead destination.
+ * A shorter bound here would silently override that budget and fail healthy
+ * backups of large databases.
+ */
+const UPLOAD_STALL_TIMEOUT_MS = 10 * 60 * 1000;
+const UPLOAD_STALL_PROBE_MS = 30 * 1000;
+
 interface ConnectionConfig {
   host: string;
   port: number;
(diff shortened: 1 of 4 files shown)


Hydralerne:
Reviewed and pushed the follow-ups directly to the branch (`b37eefd3`) so this can go in — thanks for the triage write-up, the root-cause reasoning is right and the instrumentation work that got there was the hard part.

Confirmed your central claim independently before touching anything: with only the `docker.ts` hunk reverted, the new test doesn't assert-fail, it **hangs to a 30s timeout** — exactly the production symptom. `attachDemuxed`'s only success path is `onEnd`, which resolves inside `stream.on("end")`, so when an exec finishes *cleanly* the #516 watchdog is disposed without ever firing and `stdout` stays open forever. #516 bounded the symptom; this is the cause. Two other things in the tree corroborate it: `demuxContainerStream` documents the same `demuxStream` behavior, and `backup.orchestrator.ts` `runHook` carries a 500ms drain race whose comment ("a killed exec can leave it open forever") only exists because of this bug — hooks stop paying that 500ms now.

What I changed:

**1. `tsc --noEmit` was failing.** `ws.on("error", (err) => finish(err))` is an implicit `any` (TS7006) — the old code passed `reject` directly. That broke `bun run lint` *and* `tsup --dts` for `@repo/adapters` (build exit 1). Worth knowing: CI only typechecks `apps/api` and `apps/dashboard`, and the error doesn't surface through either, so this would have gone **green and broken the release build**. Filing the CI gap separately.

**2. Stall bound 2m → 10m.** `put`'s `body` *is* the producer's stream (via `HashingPassthrough`), so source silence and destination silence are indistinguishable from inside `put()`. `execStream` deliberately allows a dump 10 minutes of silence (`DEFAULT_HELPER_IDLE_MS`, with a long comment defending generous idle bounds). At 2 minutes a healthy `pg_dump` on a lock wait, a cold-cache table scan, or any `custom_command` producer gets killed — and told the destination's SFTP subsystem is at fault. Also softened the message: your own instrumentation ruled ssh2 out (channel window wide open, nothing buffered), so the comment asserting a "known ssh2 backpressure/drain bug" would have sent the next person chasing a bug that isn't there. It's now cause-agnostic, which is all a bound like this can honestly claim.

**3. Removed the exec-exit backstop.** It can't run as a backstop: `awaitExit` only resolves after `stream` emits `end`, so `endSinks` has already fired by then — and it can't help the SSH-tunnel case it describes, because with no `end` there's no resolution either. `createIdleWatchdog.promise` only rejects and `withTimeout` only forwards, so there's no other path in. It was also missing `demuxContainerStream`'s deliberate `setTimeout(endSinks, 3000)` drain delay, which is the load-bearing part there — so if anyone later adds daemon-side exit polling to this path (the natural fix for the tunnel case), they'd have inherited exactly the truncation `docker.ts:924-932` warns about. Replaced it with a comment stating that invariant.

**4. Test for the untested half.** `backup-sftp-upload-stall.test.ts` — a wedged write stream now fails instead of hanging (hangs to timeout without the watchdog), and a source quiet for 9 minutes still completes. That second one fails if anyone tightens the bound back under the producer's budget; both fail at your original 2 minutes, which is how I checked it has teeth.

Also dropped the `#openship-*` markers (zero occurrences of that convention in the repo) and hoisted `finish` above the watchdog that calls it — it worked only because the interval couldn't fire for 10s.

Verification in `packages/adapters`: `tsc --noEmit` clean, `tsup --dts` clean, `vitest` **1049 passed / 1049 across 108 files**. Your `remote-journal.test.ts` failure doesn't reproduce here, nor did the `zod` collection issue — worth re-checking after a fresh `bun install`.

Two known gaps left as follow-ups, deliberately not in this PR: a watchdog abort leaves an orphan `.uploading-*` on the destination (pre-existing on every error path, and prune works off recorded keys), and the bound only exists for SFTP while S3 and local have the same exposure.

Note for anyone reading the description: the "2-minute idle watchdog" and the backstop it mentions are no longer in the diff.

PR #526, opened 2026-08-09
feat(apps): add Shoutrrr to one-click app catalog (#512)

thisisrakshith:
## Summary
Resolves #512

Adds **Shoutrrr** as a one-click template application in the Openship Apps catalog.

[Shoutrrr](https://shoutrrr.com/) is an open-source social media scheduling and publishing platform (alternative to Buffer, Typefully, and Hootsuite) by CoolLabs.

### Changes

- **App Template**: Added `packages/core/src/apps/catalog/shoutrrr.json` configured with:
  - Official image `ghcr.io/coollabsio/shoutrrr:latest`
  - Exposed FrankenPHP Octane port `8080`
  - Auto-generated `APP_KEY` secret config field
  - Persistent SQLite (`shoutrrr_sqlite`) and storage (`shoutrrr_storage`) volumes
  - Octane reverse-proxy headers and secure cookie configuration
  - Curated settings for application name, queue worker count, SSR, and OAuth
- **Merged Catalog**: Regenerated `packages/core/src/apps/catalog.json`
- **Branding**: Registered `shoutrrr` in `AppLogo.tsx` and vendored `apps/dashboard/public/app-logos/shoutrrr.png`

### References

- Issue: #512
- Website: https://shoutrrr.com/
- Repository: https://github.com/coollabsio/shoutrrr
- Docker Compose: https://github.com/coollabsio/shoutrrr/blob/main/docker-compose.production.yaml

PR #522, opened 2026-08-09
feat(catalog): add Kan.bn app

SulimanAbdulrazzaq:
## Summary

Adds Kan.bn as an available one-click app in the Openship catalog.

## Motivation

Issue #513 requests a self-hosted Kan.bn deployment. The template follows Kan.bn's [official repository](https://github.com/kanbn/kan) and [Docker Compose configuration](https://github.com/kanbn/kan/blob/main/docker-compose.yml): PostgreSQL 15, the migration image, and the web image with generated secrets and the documented public endpoint.

## Related issue

Closes #513

## Changes

- Added `packages/core/src/apps/catalog/kanbn.json` with the official Kan.bn services, dependencies, Postgres healthcheck, persistent volume, generated `POSTGRES_PASSWORD` and `BETTER_AUTH_SECRET`, and web endpoint.
- Regenerated `packages/core/src/apps/catalog.json`.
- Added a catalog contract test covering the Kan.bn service wiring and generated secrets.

## Verification

```text
packages/core full test suite (Vitest, --testTimeout 120000): 35 files, 420 tests passed
apps/api/test/lib/app-templates.test.ts: 6 tests passed
apps/api TypeScript: tsc --noEmit -p apps/api/tsconfig.json passed
packages/core TypeScript: tsc --noEmit -p packages/core/tsconfig.json passed
git diff --check passed
```

## Checklist

- [x] One change per PR — one agreed feature, with nothing unrelated bundled in
- [x] The diff is scoped — no reformatting or lint fixes on lines I wasn't otherwise changing
- [x] A test fails without this change and passes with it
- [x] Relevant tests, typechecks, and formatting checks pass locally
- [x] I understand every line of this diff and can explain it in review

Issue #513, opened 2026-08-08
Add Kan.bn to the apps catalog


everonegraham:
Add Kan.bn as a one-click app in the Apps catalog.

Kan.bn is an open-source project management and Kanban board application designed as a direct, self-hostable alternative to Trello.

Website: https://kan.bn/

Github: https://github.com/kanbn/kan

Docker Compose: https://github.com/kanbn/kan/blob/main/docker-compose.yml

Issue #512, opened 2026-08-08
Add Shoutrrr to the apps catalog


everonegraham:
Add Shoutrrr as a one-click app in the Apps catalog.

Shoutrrr is an open-source social media scheduling and publishing platform—essentially an open-source alternative to tools like Buffer, Typefully, or Hootsuite.

Website: https://shoutrrr.com/

Github: https://github.com/coollabsio/shoutrrr

Docker Compose: https://github.com/coollabsio/shoutrrr/blob/main/docker-compose.production.yaml

thisisrakshith:
https://github.com/oblien/openship/pull/526

PR #534, opened 2026-08-10
feat(apps): add Umami app template

### LINKED ISSUE #532: Add Umami to the apps catalog
Add Umami as a one-click app in the Apps catalog.

Umami is a lightweight, privacy-focused alternative to Google Analytics that doesn't use cookies or collect personal data, letting you track website traffic and user behavior without annoying cookie banners.

Website: https://umami.is/

Github: https://github.com/umami-software/umami

Docker Compose: https://github.com/umami-software/umami/blob/master/docker-compose.yml

Jcorderop02:
## Summary

Adds Umami (lightweight, privacy-focused, cookie-free website analytics) as an installable app in the catalog.

## Motivation

Requested in #532: users want a self-hostable analytics option that doesn't require cookie consent banners.

## Related issue

Closes #532

## Changes

- `packages/core/src/apps/catalog/umami.json`: new catalog entry — two-service docker-compose (Postgres + `ghcr.io/umami-software/umami:postgresql-latest`), auto-generated `POSTGRES_PASSWORD` shared into `DATABASE_URL` via `{{config:...}}` templating, auto-generated `APP_SECRET`. Host port pinned to `3009:3000` (not the image's default `3000:3000`) to avoid colliding with other catalog apps that already claim host port 3000.
- `packages/core/src/apps/catalog.json`: regenerated via `bun scripts/gen-catalog.ts`.
- `apps/dashboard/src/components/AppLogo.tsx`: added `umami: { slug: "umami" }` so the dashboard renders Umami's real brand mark instead of the generic fallback icon.

## Verification

```bash
$ bun run test src/apps/catalog.test.ts
 ✓ src/apps/catalog.test.ts (20 tests) 11ms
 Test Files  1 passed (1)
      Tests  20 passed (20)

$ bun run --cwd packages/core lint
$ tsc --noEmit
$ bun run --cwd apps/dashboard lint
$ tsc --noEmit
```

Also deployed live via `docker-compose` on a self-hosted Openship instance: both services (`umami-db`, `umami`) started healthy, ran their Prisma migrations, and the app served `HTTP 200` on the pinned host port with no port collisions.

## Screenshots

N/A (catalog/backend change; app UI is Umami's own, unmodified).

## Checklist

- [x] A test fails without this change and passes with it (or I explained above why there isn't one) — the existing `catalog.test.ts` schema-validation and sync tests cover any new bundled app automatically

### DIFF
--- a/apps/dashboard/src/components/AppLogo.tsx
+++ b/apps/dashboard/src/components/AppLogo.tsx
@@ -38,6 +38,7 @@ export const APP_LOGO: Record<
   freshrss: { slug: "freshrss" },
   excalidraw: { slug: "excalidraw" },
   qdrant: { slug: "qdrant" },
+  umami: { slug: "umami" },
   // Kafka's catalog id is "kafka"; its simpleicons brand slug is "apachekafka".
   // The mark is near-black (brand color #231F20), so it vanishes on the dark/dim
   // tiles — darkInvert flips it to near-white there (dark on light themes as-is).
--- a/packages/core/src/apps/catalog.json
+++ b/packages/core/src/apps/catalog.json
@@ -1829,6 +1829,92 @@
           }
         ]
       }
+    },
+    {
+      "available": true,
+      "verified": true,
+      "id": "umami",
+      "name": "Umami",
+      "description": "Lightweight, privacy-focused website analytics without cookies or personal data collection.",
+      "kind": "template",
+      "logo": "umami",
+      "category": "analytics",
+      "tags": [
+        "analytics",
+        "privacy",
+        "web"
+      ],
+      "framework": "docker-compose",
+      "services": [
+        {
+          "name": "umami-db",
+          "image": "postgres:15-alpine",
+          "environment": {
+            "POSTGRES_DB": "umami",
+            "POSTGRES_USER": "umami"
+          },
+          "secretEnv": [
+            "POSTGRES_PASSWORD"
+          ],
+          "volumes": [
+            "umami_db_data:/var/lib/postgresql/data"
+          ],
+          "restart": "unless-stopped"
+        },
+        {
+          "name": "umami",
+          "image": "ghcr.io/umami-software/umami:postgresql-latest",
+          "ports": [
+            "3009:3000"
+          ],
+          "exposedPort": 3000,
+          "exposed": true,
+          "dependsOn": [
+            "umami-db"
+          ],
+          "environment": {
+            "DATABASE_URL": "postgresql://umami:{{config:POSTGRES_PASSWORD}}@umami-db:5432/umami"
+          },
+          "secretEnv": [
+            "APP_SECRET"
+          ],
+          "restart": "unless-stopped"
+        }
+      ],
+      "configFields": [
+        {
+          "key": "POSTGRES_PASSWORD",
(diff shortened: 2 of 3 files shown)


Hydralerne:
Thanks for this, @Jcorderop02 — and sorry for the collision. Your commit landed at 11:04 on Aug 10; Umami went into `main` about five hours later as part of the 0.6.2 patch (698893ec, merged via #536) before we'd triaged this PR. That's on us for not claiming #532 first, not on you. The "conflict" GitHub is showing is just the two Umami templates landing on the same three files.

Since the shipped template is a superset of this one, I'm going to close this rather than merge it. Concretely, what's on `main` adds:

- **`ghcr.io/umami-software/umami:postgresql-v2.19.0`** instead of `postgresql-latest`. We pin every catalog image — an unpinned tag means a redeploy silently jumps versions, which for Umami means Prisma migrations run against a schema the operator didn't ask to change.
- **Healthchecks on both services** (`/api/heartbeat` for umami, `pg_isready` for the db). Without one, a container that starts and then crashes reads as a successful deploy.
- **`secretEnv: ["DATABASE_URL"]` on the umami service.** This is the one I'd flag as a real bug rather than a difference: `DATABASE_URL` is built from `environment` and interpolates `{{config:POSTGRES_PASSWORD}}`, so without that entry the generated DB password gets written as plaintext compose env. Worth knowing for future templates — `secretEnv` covers keys that come from `environment`, whereas a `configFields` entry with `secret: true` is already encrypted on its own (see the `isSecret` computation in `app-install.service.ts`), so the `secretEnv: ["APP_SECRET"]` in this diff was a no-op.
- **`DATABASE_TYPE: postgresql`**, plus `routes`, `endpoints`, `management`, and `connection.guide` — the routing and connection metadata the dashboard needs to front the app and render its Connect card.
- **`darkInvert: true` on the `AppLogo` entry.** Umami's brand mark is near-black (#000000), so without the flag it disappears on the dark and dim tiles.

On the host port: the reasoning in your description was sound, but `3009:3000` isn't needed here. For a service that's `exposed: true` with `routes`, `withLoopbackPublish` (`apps/api/src/lib/loopback-publish.ts`) replaces the routed port's binding with `127.0.0.1:<pinned>:3000`, so the declared host port is discarded under the default route strategy — which is why 25 of the 28 catalog apps declare a plain `N:N`. Hardcoded publishes only bite on services with **no** routes (`mongodb.json` is the outstanding example).

None of this is wasted effort on your part — the template was structurally right, and the config templating and generated-secret wiring were both correct. If you're up for another one, #512 (Shoutrrr), #513 (Kan.bn), and #415 (OpenTelemetry Collector) are all open catalog requests and unclaimed. Comment on whichever you want and it's yours, and I'll review quickly this time.

Jcorderop02:
Thanks for the pointers. Turns out all three already have an open PR from someone else. #512 has #526 (thisisrakshith). #513 has #522 (SulimanAbdulrazzaq). #415 has #416 (giomartinsdev). All three look complete, not abandoned stubs, just unreviewed.

Happy to pick a different unclaimed catalog request if you have one, or take a pass at reviewing/testing one of those three PRs instead if that's more useful to you than a fourth person building the same thing.

PR #540, opened 2026-08-10
fix(mcp): make the remote MCP OAuth flow work with spec-strict clients (Claude.ai)

kaka-sangi:
## Problem

Connecting a self-hosted Openship instance to Claude.ai as a custom connector fails with `Authorization with the MCP server failed`, even though the OAuth consent completes and the browser is redirected back. Cursor and other MCP clients connect fine, which is what makes this look like a client bug — it isn't. Claude.ai implements the current MCP authorization spec strictly, and the parts it relies on were missing or broken.

Reproduced against a self-hosted instance at `https://<host>/api/mcp` (v0.6.1, `openship up` docker stack with the dashboard same-origin proxy).

## Root causes

**1. The authorize redirect leaked the internal origin — this alone breaks the flow for every remote client.**

`forceMcpConsent` rebuilds the authorize URL from `c.req.url` in order to inject `prompt=consent`. Behind the dashboard's same-origin proxy the API never sees the public host: the proxy sets `host` to the upstream authority (`apps/dashboard/src/app/api/proxy/[...path]/route.ts` → `out.set("host", upstream.host)`), so `c.req.url` is `http://api:4000/api/auth/mcp/authorize?…`. That value went straight into a 302:

```
$ curl -sD- 'https://<host>/api/auth/mcp/authorize?response_type=code&client_id=…'
HTTP/1.1 302 Found
location: http://api:4000/api/auth/mcp/authorize?…&prompt=consent
```

The browser is sent to a host it cannot resolve, mid-OAuth. Fixed by rebuilding on the public origin, the same one `requestPublicOrigin` already supplies to every other advertised URL.

**2. No path-aware Protected Resource Metadata (RFC 9728 §3.1).**

For a resource identifier with a path, the metadata lives at the well-known prefix *followed by that path*. A client configured with `https://<host>/api/mcp` fetches `/.well-known/oauth-protected-resource/api/mcp`; that 404'd, so a strict client falls back to the origin document — whose `resource` is the bare origin and therefore does not match the URL it connected to.

**3. The `resource` parameter (RFC 8707) was ignored.**

better-auth's `mcp()` plugin drops it on both the authorize and the token request and mints an opaque token with no audience, so nothing binds the token to the MCP server it was issued for, and an unknown `resource` was silently accepted instead of rejected.

## Changes

- **Path-aware discovery.** `/.well-known/oauth-protected-resource/api/mcp` (and the `/api/proxy/api/mcp` alias) now return a document whose `resource` is the full MCP URL including its path, plus the matching RFC 8414 authorization-server documents. `apps/dashboard/next.config.mjs` gains the rewrites that make them publicly reachable in proxy builds. The root documents are unchanged.
- **`resource` is honored.** Validated at the authorize endpoint and at the token endpoint against the set of URLs that address this instance's MCP endpoint; an unknown value is rejected with `invalid_target` rather than dropped. Comparison is on canonicalized URLs (lowercase scheme/host, no default port, no trailing slash), so `/api/mcp` and `/api/proxy/api/mcp` are both recognized.
- **Audience-bound tokens.** The MCP token endpoint re-issues the plugin's opaque token as a JWT carrying `aud` (the requested resource), `iss`, `sub` and `jti`, rewriting the stored row so introspection stays an exact-string lookup. `/api/mcp` verifies that audience.
- **401 challenge.** `WWW-Authenticate` now points at the path-aware metadata URL, and an unauthenticated `GET /api/mcp` answers 401 with that pointer instead of a bare 405 (an authenticated GET is still 405 — this server pushes no SSE).

Everything is derived from the instance's configured public URL; no host is hardcoded.

## Backward compatibility

- A client that never sends `resource` gets the canonical MCP URL as its audience and is unaffected.
- Tokens issued before this change are opaque, carry no audience, and are treated as unconstrained, so existing connections keep working.
- The bare origin stays in the accepted-resource set, so a client that echoed back the old root metadata still validates.
- PKCE S256 and dynamic client registration are untouched. The MCP endpoint still never 3xx-redirects (Claude drops `Authorization` across cross-origin redirects), and the token endpoint adds one indexed UPDATE.

## Tests

- Unit coverage for resource canonicalization, the allowed-resource set, audience binding/decoding, and the public-origin rebuild (`apps/api/test/lib/mcp-resource.test.ts`, `apps/api/test/lib/mcp-token.test.ts`).
- `scripts/mcp-oauth-check.ts` drives the whole flow against a running instance — discovery → DCR → PKCE authorize → consent → token exchange → `initialize` — asserting the metadata shape, the 401 challenge, `invalid_target`, the token's `aud`/`iss`, and that the same flow still works with no `resource` parameter:

```bash
bun scripts/mcp-oauth-check.ts --base https://ops.example.com --email you@example.com --password '…'
```

## Verified on a live instance

Against a patched self-hosted deploy (v0.6.1 + this commit):

```
GET /.well-known/oauth-protected-resource/api/mcp
  200 {"resource":"https://<host>/api/mcp","authorization_servers":["https://<host>"],…}
GET /.well-known/oauth-protected-resource/api/proxy/api/mcp
  200 {"resource":"https://<host>/api/proxy/api/mcp",…}
GET  /api/mcp  → 401, WWW-Authenticate: Bearer resource_metadata="https://<host>/.well-known/oauth-protected-resource/api/mcp"
POST /api/mcp  → 401, same pointer, no redirect
authorize?…&resource=https://<host>/api/mcp        → 302 to https://<host>/… (public origin)
authorize?…&resource=https://evil.example.com/…    → 400 {"error":"invalid_target"}
POST /mcp/token with a foreign resource            → 400 {"error":"invalid_target"}
POST /mcp/token with a valid resource, bad code    → 401 from the plugin (wrapper passes through)
```

### DIFF
--- a/apps/api/src/app.ts
+++ b/apps/api/src/app.ts
@@ -17,6 +17,11 @@ import { runWithCallSource } from "./lib/call-source";
 import { authRoutes } from "./modules/auth/auth.routes";
 import { auth } from "./lib/auth";
 import { oAuthDiscoveryMetadata, oAuthProtectedResourceMetadata } from "better-auth/plugins";
+import {
+  MCP_RESOURCE_PATHS,
+  protectedResourceMetadata,
+  publicOriginFor,
+} from "./lib/mcp-resource";
 import { projectRoutes } from "./modules/projects/project.routes";
 import { appRoutes } from "./modules/apps/app.routes";
 import { appSettingsRoutes } from "./modules/apps/app-settings.routes";
@@ -159,6 +164,33 @@ app.route("/api/notices", noticeRoutes);
 // so `Authorization`-less requests to /api/mcp can be discovered end-to-end.
 app.get("/.well-known/oauth-authorization-server", (c) => oauthAuthServerMetadata(c.req.raw));
 app.get("/.well-known/oauth-protected-resource", (c) => oauthProtectedResourceMetadata(c.req.raw));
+
+// RFC 9728 §3.1: metadata for a resource whose identifier has a PATH lives at
+// the well-known prefix FOLLOWED BY that path. A client configured with
+// `https://host/api/mcp` looks there, not at the root — and the root document's
+// `resource` (the bare origin) doesn't match the URL it connected to, so a
+// strict client (Claude.ai) rejects the authorization it just completed.
+// Serve one document per URL that addresses this instance's MCP endpoint.
+const OAUTH_DISCOVERY_HEADERS = {
+  "Content-Type": "application/json",
+  "Access-Control-Allow-Origin": "*",
+  "Access-Control-Allow-Methods": "GET, OPTIONS",
+  "Access-Control-Allow-Headers": "Content-Type, Authorization",
+} as const;
+
+for (const path of MCP_RESOURCE_PATHS) {
+  app.get(`/.well-known/oauth-protected-resource${path}`, (c) => {
+    const origin = publicOriginFor(c.req.raw);
+    const body = protectedResourceMetadata(origin, `${origin}${path}`);
+    return new Response(JSON.stringify(body), { status: 200, headers: OAUTH_DISCOVERY_HEADERS });
+  });
+  // RFC 8414 path-aware authorization-server metadata. Same document as the
+  // root one — served here so a client that only probes the path-aware location
+  // finds it instead of falling back.
+  app.get(`/.well-known/oauth-authorization-server${path}`, (c) =>
+    oauthAuthServerMetadata(c.req.raw),
+  );
+}
 
 /* ---------- OAuth callback landing pages ---------- */
 const authCallbackHtml = `<!DOCTYPE html><html><head><title>Success</title></head><body><script>window.close();</script><p>Authentication successful. You can close this window.</p></body></html>`;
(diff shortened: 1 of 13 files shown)


kaka-sangi:
Pushed a second commit after testing the full flow against Claude.ai on a live instance: `483d801`.

The resource/discovery fixes above got the flow through consent and token exchange, but Claude still aborted immediately after — without ever calling the MCP endpoint. Root cause is in better-auth's `mcp()` plugin itself:

- The discovery metadata advertises `jwks_uri` and `userinfo_endpoint` under `/api/auth/mcp/*`, but the plugin implements **neither** — both fell through to the Better Auth catch-all and 404'd.
- The `id_token` it mints is signed **HS256 with a key generated and discarded inside the request handler**, has no `iss` claim, and stamps `iat`/`auth_time` in milliseconds. The metadata advertises `id_token_signing_alg_values_supported: ["RS256","none"]`, so no verifying client can accept it.

Claude requests `openid` (it's in `scopes_supported`) and validates the id_token it receives: it fetches the jwks, finds nothing that can verify the token, and drops the connection with "the integration rejected the credentials it just issued". Cursor and friends never noticed because they don't validate the id_token.

The commit gives the instance a persistent RS256 keypair (generated once, stored in the Better Auth `verification` table — no migration), serves its public half at the advertised `jwks_uri`, re-issues the id_token in the token wrapper (same user claims, corrected `iss`/`aud`/second-precision timestamps, signed with a `kid` the jwks lists), and implements `userinfo` against the issued-token row.

With both commits, a Claude.ai custom connector completes end-to-end against a self-hosted instance (verified live, `initialize` → `tools/list` working).

Hydralerne:
great, thank you

PR #574, opened 2026-08-13
fix(services): derive commandArgv when an editor sends only the text command

### LINKED ISSUE #573: Editing a compose service's command in the dashboard leaves command_argv stale, so the container keeps running the old command
### Summary

Editing a compose service's `command` from the dashboard does not update `service.command_argv`. Since #332 made `commandArgv` the authoritative thing handed to Docker, the stale argv keeps winning: **the container goes on running the old command while the UI shows the new one**, and clearing the field never hands the image's own `CMD` back.

This surfaced from a user report claiming Openship still forces `sh -c`. The compose-file path is fine — #332 fixed that, and `shellSplitWords` / `commandToArgv` / `resolveComposeCmd` all behave as documented. The gap is the API write path, which never derives the argv from the text `command`.

### Why nothing downstream repairs it

`toComposeSpec` backfills an argv only when the stored one is **null** (`s.commandArgv ?? commandToArgv(s.command)`). A *stale non-null* argv is therefore preserved by the deploy-time `syncFromCompose`, and `reconcileFromCompose` keeps "ours" for an edited row — so the mismatch survives every redeploy.

It also produces phantom drift: the row canonicalizes to `{command: "ak worker", commandArgv: ["ak","server"]}` while the re-parsed file gives `{command: "ak worker", commandArgv: ["ak","worker"]}`, so `composeSpecDiff` reports a `commandArgv` change on a service the operator already edited to match the file.

### Where

- `apps/dashboard/src/app/(dashboard)/projects/[id]/components/services/ServiceSettingsForm.tsx` — the settings payload sends `command: command.trim()` and no `commandArgv`.
- `apps/api/src/modules/services/service.service.ts` → `updateService` — `patch = { ...data }`, `command` gets `trimOrNull`, `commandArgv` is never derived, and `repos.service.update` is a raw `set`, so the column keeps its previous value.
- `packages/adapters/src/runtime/compose-cmd.ts` → `resolveComposeCmd` prefers `commandArgv` and only falls back to `sh -c <command>` when it is null. Correct as written — it's being fed a stale argv.

### Reproduce

1. Import a compose project with a service whose image has a strict `ENTRYPOINT` and whose compose entry sets a `command` (authentik's `server` / `worker` pair on `ghcr.io/goauthentik/server` is the real-world case).
2. Dashboard → that service → Settings → change `command` from `ak server` to `ak worker` → Save.
3. Redeploy.

**Expected:** the container runs `ak worker`.
**Actual:** it runs `ak server`; `command_argv` still holds `["ak","server"]`. Clearing the field instead leaves `["ak","server"]` running rather than restoring the image `CMD`.

### Secondary: `createService` stores a text command with no argv

`createService` persists `commandArgv: data.commandArgv ?? null`, so a caller that sends only the text `command` — the app installer does, forwarding a template's string command at `app-install.service.ts` — leaves a row whose only rescue is the null-argv backfill described above. I traced that backfill and it *does* cover the first deploy (`build-pipeline.ts` runs `syncFromCompose` before the pipeline re-reads the rows), so this is not a live `sh -c` regression today — but it makes correct startup depend on a repair step rather than on the row being written consistently, and 13 catalog services declare a string `command`.

### Proposed fix

Derive `commandArgv` from the text `command` in both editors whenever the caller didn't send one, so the two columns can't disagree — and keep an explicitly-sent argv authoritative (including `[]` to clear the CMD) so argv-aware callers (compose parser, CLI, snapshot replay) are never second-guessed. Two lines in `service.service.ts` plus a regression test.

PR follows.

### Related, not covered here

`entrypoint:` is still in `UNSUPPORTED_SERVICE_KEYS` (`apps/api/src/lib/compose-parser.ts`) — a compose file that overrides the entrypoint gets a warning and is then ignored. Separate gap; happy to open its own issue if that's wanted.

AbdullahM07:
Fixes #573.

### What was broken

#332 made `commandArgv` the argv Openship hands Docker, keeping `["sh","-c",<command>]` only as the fallback for legacy rows that have no argv at all. Neither service editor derives that argv, and the dashboard's settings form sends the one-line `command` field alone — so the two columns drift apart:

- **`updateService`** writes the new `command` and leaves the row's **previous** argv in place (`patch = { ...data }`, `command` gets `trimOrNull`, and `repos.service.update` is a raw `set`). `toComposeSpec` backfills only a **null** argv, so the stale one survives the deploy-time `syncFromCompose` and `reconcileFromCompose` keeps "ours" for an edited row. The container goes on running the old command while the UI shows the new one, clearing the field never restores the image `CMD`, and `composeSpecDiff` reports phantom `commandArgv` drift on a service the operator edited to match its own compose file.
- **`createService`** stores a text command with `commandArgv: null` — the app installer forwards a template's string command through here (13 catalog services declare one). That null *is* covered by the backfill, so it is not a live `sh -c` regression today; I traced `build-pipeline.ts` → `syncFromCompose` → the pipeline's fresh `listByProject` read and confirmed the first deploy repairs it. Included anyway so the row is written consistently instead of depending on a repair step.

The compose-file path was already correct — `shellSplitWords`, `commandToArgv` and `resolveComposeCmd` all behave as documented. This is only the API write path.

### The change

Derive the argv from the text `command` in both editors when the caller didn't send one — in `updateService` alongside the existing trim, so the two columns can't be written out of sync, and in `createService` at the insert. An explicitly-sent `commandArgv` stays authoritative, `[]` (clear the image `CMD`) included, so argv-aware callers — the compose parser, the CLI, a snapshot replay — are never second-guessed and a quoted `sh -c "…"` argv is never re-split.

Two derivation lines plus a regression test; no schema, endpoint or runtime change.

### How I verified it

`apps/api/test/modules/services/service-command-argv.test.ts` drives the real `createService` / `updateService` against mocked repos and asserts the patch that reaches the DB.

On `main`, 3 of the 4 assertions fail:

```
× re-derives the argv when an edit sends only the text command
  → AssertionError: expected undefined to deeply equal [ 'ak', 'worker' ]
× clears the argv when an edit clears the command
  → AssertionError: expected undefined to be null
× derives the argv on create instead of storing none
  → AssertionError: expected null to deeply equal [ 'ak', 'worker' ]
✓ keeps an explicit argv over the text command
```

With the change, all 4 pass. The fourth passes both ways on purpose — it pins the boundary, since the tempting wrong implementation (always re-split the text) would mangle an explicit `["sh","-c","echo hi && ak server"]`.

Also run:

- `bun run --cwd apps/api test` — 328/333 files pass on the branch. The 5 that fail are pre-existing on this machine and unrelated: the same suite on clean `main` fails **6** files (a superset), always the whole-app route/permission builders (`app.routes`, `collection-project-routes`, `env-reveal-gating`, `pending-actions-tools`, `service-exec-isolation`) plus `container-events`, all on 10–20s timeouts, and one surfacing `INTERNAL_TOKEN is required when DEPLOY_MODE="docker"` from my local `.env`. Their membership shifts run to run while the diff stays constant. `test/modules/services/` — the module this change touches, `service-exec-isolation` included — passes 5/5 files, 42 tests, plus the 4 new ones.
- `bun run --cwd apps/api lint` — `tsc --noEmit` clean.
- `bunx prettier --check` on both touched files. The new test file is clean. `service.service.ts` was already Prettier-dirty on `main` (54 hunks — long single-line imports), so per CONTRIBUTING I left that drift alone rather than reformatting lines this change doesn't need; the added lines match the surrounding style.

### Not in this PR

`entrypoint:` is still listed in `UNSUPPORTED_SERVICE_KEYS` — a compose file that overrides the entrypoint gets a warning and is then ignored. Noted in #573 as a separate gap.

### DIFF
--- a/apps/api/src/modules/services/service.service.ts
+++ b/apps/api/src/modules/services/service.service.ts
@@ -3,7 +3,7 @@
  */
 
 import { normalizeRoutingFields, repos, composeSpecDiff, type Project, type Service, type ServicePublicEndpoint } from "@repo/db";
-import { aliasConflictsWithSiblings, getProjectType, mergeAdvanced, normalizeServiceLabel, normalizeAliasStrict, safeErrorMessage, withTimeout, type ComposeAdvanced, type ServiceContainerState, type StackId } from "@repo/core";
+import { aliasConflictsWithSiblings, commandToArgv, getProjectType, mergeAdvanced, normalizeServiceLabel, normalizeAliasStrict, safeErrorMessage, withTimeout, type ComposeAdvanced, type ServiceContainerState, type StackId } from "@repo/core";
 import {
   BuildLogger,
   DockerRuntime,
@@ -484,6 +484,8 @@ export async function createService(
   // service counts as a sibling, which is exactly right for a new row.
   await validateServiceAlias(projectId, "", advanced, project.internalAlias);
 
+  const command = trimOrNull(data.command);
+
   const created = await repos.service.create({
     projectId,
     name,
@@ -495,8 +497,15 @@ export async function createService(
     dependsOn: data.dependsOn ?? [],
     environment: data.environment ?? {},
     volumes: data.volumes ?? [],
-    command: trimOrNull(data.command),
-    commandArgv: data.commandArgv ?? null, // #332
+    command,
+    // #332: derive the argv when the caller sent only the text `command` — the
+    // dashboard's service form and the app installer (app-install.service.ts,
+    // which forwards a template's string command) both do. A row stored with a
+    // text command and a null argv is only rescued by the deploy-time backfill in
+    // `toComposeSpec`; deriving it at the insert keeps the two columns consistent
+    // without depending on that, and out of resolveComposeCmd's legacy
+    // `sh -c <command>` fallback. An explicit argv still wins, `[]` included.
+    commandArgv: data.commandArgv ?? commandToArgv(command), // #332
     restart: data.restart ?? "unless-stopped",
     advanced,
     ...routing,
@@ -589,6 +598,17 @@ export async function updateService(
     if (key in patch) {
       patch[key] = trimOrNull(patch[key]);
     }
+  }
+  // #332: `commandArgv` is what the runtime actually runs — resolveComposeCmd
+  // prefers it and only falls back to `sh -c <command>` for legacy rows that have
+  // no argv at all. The dashboard's service form sends the one-line `command` and
+  // no argv, so a text-only edit left the PREVIOUS argv in the row: the container
+  // went on running the old command while the UI showed the new one, and clearing
+  // the field never handed the image's own CMD back. Re-derive alongside the trim
+  // so the two columns can't disagree. An explicit argv from an argv-aware caller
+  // (the compose parser, the CLI, a snapshot replay) still wins, `[]` included.
+  if ("command" in patch && patch.commandArgv === undefined) {
+    patch.commandArgv = commandToArgv(patch.command);
   }
   // Monorepo sub-app build settings: same trim-or-null treatment so empty
   // strings become null in DB (matches the rest of the service columns).
--- a/apps/api/test/modules/services/service-command-argv.test.ts
+++ b/apps/api/test/modules/services/service-command-argv.test.ts
@@ -1,0 +1,131 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const projectRepo = vi.hoisted(() => ({ findById: vi.fn() }));
+const serviceRepo = vi.hoisted(() => ({
+  findById: vi.fn(),
+  findByName: vi.fn(),
+  listByProject: vi.fn(),
+  listByDeployment: vi.fn(),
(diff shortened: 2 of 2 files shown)


AbdullahM07:
Superseded — `main` already does this, via `resolveCommandArgv` in 74ec3f72. Closing.

Upstream's version is the better one, so there is nothing worth salvaging here:

- **Create** — `resolveCommandArgv({ incomingArgv, incomingCommand })` is what this PR did as `data.commandArgv ?? commandToArgv(command)`.
- **Update** — it also threads `storedCommand` / `storedArgv`, which handles a case this PR did not: the form echoing back a list command's lossy display join, where re-deriving would corrupt an argv that cannot be re-split. This PR re-derived unconditionally on a trim.

Same for the tests. `service-command-argv.test.ts` on `main` has 8 cases and covers all 4 of mine, plus "does not touch argv when the patch never mentions the command", the lossy-display-join case, "forwards an explicit argv verbatim (what the CLI sends)", and "restores masked env without inventing an argv the caller didn't send".

Closing rather than rebasing — a rebase here would only re-litigate a fix that already landed in a stronger form.

PR #582, opened 2026-08-14
fix(runtime): authenticate private registry pulls

### LINKED ISSUE #581: Private registry image pulls omit Docker authentication
## Problem

Local image pulls use dockerode without passing registry credentials. As a result, deployments that reference a private image such as `registry.example.com/org/private-image:immutable` fail even when the host Docker CLI is authenticated.

The source-generated Compose stack also does not make an API-readable Docker configuration available to the API service, so the API cannot resolve credentials at pull time.

## Expected behavior

- Resolve matching inline credentials from Docker CLI `config.json` for local Docker Engine pulls.
- Pass them to dockerode through `pull(ref, { authconfig })`, which sends `X-Registry-Auth` to Docker Engine.
- Keep SSH-executor pulls unchanged and avoid reading or transferring local credentials for remote pulls.
- Mount an existing Docker configuration into only the API service, read-only.
- Fail closed with a redacted actionable error when the configuration relies on `credHelpers` or `credsStore`, because the current API image does not include credential-helper binaries or their backing state.
- Do not expose credentials or Docker error details in surfaced pull errors or logs.

## Proposed regression coverage

- Authenticated local pulls receive `authconfig`.
- Inline identity/basic credentials resolve for the matching registry.
- Helper-backed configurations fail closed without exposing helper names.
- Pull failures do not surface credential material.
- SSH executor pulls do not read local Docker credentials.
- Generated Compose mounts only a usable regular config file into the API with a read-only long bind mount, including safe handling of relative and YAML-significant paths.

I have a focused patch and tests ready and will link the draft PR here.

M-Aref-Shojaei:
## Summary

Authenticate local Docker Engine image pulls from API-readable Docker CLI configuration and mount that configuration into only the API service.

## Motivation

Private image deployments fail because local dockerode pulls omit `authconfig`, even when Docker CLI authentication exists on the host. Pull and credential errors also need to remain redacted.

## Related issue

Fixes #581

## Changes

- resolve matching inline basic or identity-token credentials from Docker CLI `config.json`
- pass resolved credentials to `docker.pull(ref, { authconfig })`
- leave SSH-executor pulls unchanged and isolated from local credentials
- fail closed when matching `credHelpers` or global `credsStore` cannot run in the API image
- redact credential and Docker progress details from surfaced pull errors
- mount an existing regular Docker config into only the API service using read-only Compose long bind syntax
- add focused runtime and Compose regression coverage

## Verification

- `bun run --cwd packages/adapters test src/runtime/docker-pull-auth.test.ts` — 1 file, 6 tests passed
- `bun run --cwd apps/cli test test/unit/compose-source-build.test.ts` — 1 file, 11 tests passed
- `bun run --cwd packages/adapters lint` — passed
- `bun run --cwd apps/cli lint` — passed
- full adapters suite — 132 files, 2862 tests passed
- full API suite — 332 files, 3931 passed, 3 skipped
- API lint/typecheck — passed
- root build — passed
- full CLI suite — 32 files passed and 1 unrelated pre-existing timeout in `test/unit/ports.test.ts`; neither that test nor its implementation is modified

## Checklist

- [x] I read `CONTRIBUTING.md`
- [x] A real regression test was observed failing before the production change
- [x] Focused tests pass
- [x] Changed workspaces typecheck
- [x] No credentials, deployment identifiers, or private environment data are included
- [ ] Live API-image deployment verification complete
- [ ] Full root test gate is green without the unrelated CLI timeout

### DIFF
--- a/.github/workflows/build-patched-api.yml
+++ b/.github/workflows/build-patched-api.yml
@@ -1,6 +1,9 @@
 name: Build patched API image
 
 on:
+  push:
+    branches:
+      - fix/docker-pull-registry-auth
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/build-patched-api.yml
+++ b/.github/workflows/build-patched-api.yml
@@ -1,0 +1,30 @@
+name: Build patched API image
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: docker/setup-buildx-action@v3
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: apps/api/Dockerfile
+          push: true
+          tags: ghcr.io/m-aref-shojaei/openship-api:pull-auth-0eeee135
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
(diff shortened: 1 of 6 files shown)


Hydralerne:
upcoming feature currently out of scope

PR #590, opened 2026-08-16
fix(auth): resolve the merge conflict markers committed in v0.6.6

Farahat612:
apps/api/src/lib/auth.ts carries unresolved conflict markers from the
v0.6.6 patch, so `tsc --noEmit` fails with TS1185 and the API workspace
does not compile. CI runs that typecheck, so main is red and so is every
pull request opened against it.

Kept the v0.6.6 side, which is the change that patch was making and the
behaviour the sibling forget-password branch a few lines below already
implements: sendMail returns a boolean, and a verification code that was
never delivered has to fail loudly rather than tell the user to check an
inbox that will stay empty.

Verified: bun run --cwd apps/api lint is clean again, and the apps/api
lib suite passes (90 files, 1376 tests).

Noticed while rebasing another branch: the API workspace stopped compiling
on main. CI runs `bun run --cwd apps/api lint` (tsc --noEmit), so this is
currently failing on main and on every open pull request against it.

I kept the v0.6.6 side of the conflict because that is the change the patch
was making, and because the sibling forget-password branch a few lines below
already implements exactly that pattern: sendMail returns a boolean, and a
verification code that was never delivered should fail loudly rather than
tell the user to check an inbox that stays empty.

Happy to flip it to the other side instead if the intent was different.

### DIFF
--- a/apps/api/src/lib/auth.ts
+++ b/apps/api/src/lib/auth.ts
@@ -415,9 +415,6 @@ export const auth = betterAuth({
         // types are not enabled, so they fall through and send nothing.
         if (type === "email-verification") {
           const tmpl = verifyOtpEmailTemplate(otp, { expiresMinutes: 10 });
-<<<<<<< HEAD
-          await sendMail({ to: email, ...tmpl });
-=======
           if (!(await sendMail({ to: email, ...tmpl }))) {
             throw new APIError("SERVICE_UNAVAILABLE", {
               message:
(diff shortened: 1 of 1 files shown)


Farahat612:
Closing: this is already fixed on main. b2223b9d resolves the markers in apps/api/src/lib/auth.ts to the same side this PR did, the guarded sendMail that surfaces SERVICE_UNAVAILABLE rather than reporting a success no transport could deliver, and the API workspace compiles again.

Thanks for getting it in.

PR #593, opened 2026-08-16
feat(auth): advertise configured login providers on the public env endpoint

Farahat612:
Nothing told the dashboard which OAuth providers this instance actually
has credentials for, so it guessed from selfHosted: cloud got hardcoded
github+google buttons, self-hosted got none: hiding working buttons
from every operator who had configured a provider, and set to render
dead buttons on any cloud deploy missing a pair. TODO.md names a
server-advertised provider list as the first task for both the SSO item
and the git-provider plan (#75).

The predicate "is this provider configured?" now lives once in
lib/auth-providers.ts: lib/auth.ts reads the credential pairs to
register providers with Better Auth, and GET /health/env advertises the
resulting list as authProviders: ids and a kind discriminator only,
never credentials (asserted in tests; the route is unauthenticated).
Both halves of a pair are required; half-configured counts as absent.
kind is "social" for everything today: the field exists so a future
SSO entry can be advertised without an older dashboard drawing it as a
branded button. Git-provider capabilities are deliberately a separate
future field: a git remote is not a login button.

OAuthButtons renders the intersection of what the server advertises and
what the build can draw (icon + translated label), in a stable order,
and renders nothing: no divider: for an empty list, which makes the
!selfHosted gates on login and register removable. authMode "none"
instances are untouched (the desktop redirect precedes the form). No
new i18n keys: labels stay t.auth.oauth[id], now indexed by id.

Tests: provider predicate (configured/absent/half-configured/no secret
leakage), the endpoint seam both ways, the pure list-to-buttons mapping,
and server-rendered button output for advertised, empty and unknown-kind
lists.

Part of the SSO groundwork: this is the named first task of the SSO item in
TODO.md ("a server-advertised provider list ... is the real first task"). The
same missing primitive is referenced in the git-providers section, but this
change is only about login providers and does not touch git in any way.

Behavioral improvement: a self-hosted instance with OAuth credentials
configured now shows working login buttons (previously hidden outright by the
selfHosted guess). Default self-hosted (no creds) and cloud are unchanged.
The unauthenticated payload carries provider ids and a kind discriminator
only, never credentials (asserted in tests). No new i18n keys; labels are
indexed by provider id.

How verified: tests on the provider predicate (configured, absent,
half-configured, no secret leakage), the endpoint seam both ways, the pure
list-to-buttons mapping, and server-rendered button output for advertised,
empty and unknown-kind lists. apps/api and dashboard suites green,
typechecks clean.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -112,15 +112,25 @@ What's already there:
 
 The prerequisite nobody expects:
 
-- [ ] **The button can't just be added.** Social login is hidden on self-hosted
-      outright today — `{!selfHosted && <OAuthButtons/>}` at
-      `apps/dashboard/src/app/(auth)/login/page.tsx:246` and
-      `register/page.tsx:154` — because an operator with no `GITHUB_CLIENT_ID`
-      would get buttons that fail, and **nothing tells the dashboard which
-      providers are configured**. `OAuthButtons` hardcodes github+google. SSO
-      needs a server-advertised provider list (public, read-only, alongside the
-      `authMode`/`selfHosted` values `useAuthContext` already serves). That
-      endpoint doesn't exist yet and is the real first task.
+- [x] **The button can't just be added** — DONE for the *primitive*, not for
+      SSO. `GET /health/env` now advertises `authProviders`, the social logins
+      the server actually has credentials for
+      (`apps/api/src/lib/auth-providers.ts` owns the predicate; `lib/auth.ts`
+      reads the SAME resolver to decide registration, so "a button is shown" and
+      "the provider is registered" cannot drift). Public, read-only, IDs only —
+      no client id, no secret, asserted by test. The dashboard carries it
+      through the auth layout's context, and `{!selfHosted && <OAuthButtons/>}`
+      is gone from both `login/page.tsx` and `register/page.tsx`:
+      `OAuthButtons` maps the advertised list and renders nothing (not even the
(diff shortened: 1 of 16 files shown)


mohamedadel96e (commented):
Moving the “is this OAuth provider configured?” check to the backend and letting the frontend render the buttons based on that feels like the right approach, especially for self-hosted users who have OAuth credentials configured.

I think this is an important piece for the multi-provider work. Once we start supporting GitLab and other providers, the frontend needs a reliable way to know which providers are actually available instead of relying on things like `!selfHosted`.

I like that `/health/env` only exposes the provider `id` and `kind`. We’re not exposing any credentials, and the frontend doesn’t need to know anything about the actual configuration. The mapping in `auth-providers.ts` also makes sense. If the backend starts returning a provider that an older frontend doesn't know about, it just gets ignored instead of leaving us with a broken button.

The tests cover the important cases, like empty providers, unsupported providers, and non-social providers.

I agree with the point about #75 as well. I think we should get this merged, along with the data-layer fixes around making `gitProvider` a properly checked union, using `gitUrl` for cloning, and fixing the tarball fast-path issue, before merging #386.

Once those are in, rebasing #386 should be cleaner. We can then remove the remaining hardcoded GitHub fallbacks I made and hopefully finish the multi-Git-provider work without having to keep patching around these assumptions.

Nice work BTW I liked it.

PR #595, opened 2026-08-16
fix(auth): resolve the conflict markers left in the verification OTP path

### LINKED ISSUE #594: main does not typecheck — conflict markers committed in apps/api/src/lib/auth.ts
`main` currently does not typecheck: `apps/api/src/lib/auth.ts` has unresolved merge conflict markers committed into it.

## Reproduce

```
git clone https://github.com/oblien/openship && cd openship
bun install
bun run lint
```

```
src/lib/auth.ts(418,1): error TS1185: Merge conflict marker encountered.
src/lib/auth.ts(420,1): error TS1185: Merge conflict marker encountered.
src/lib/auth.ts(428,1): error TS1185: Merge conflict marker encountered.
```

The markers are in the `emailOTP` plugin's `sendVerificationOTP`, on the `email-verification` branch:

```ts
if (type === "email-verification") {
  const tmpl = verifyOtpEmailTemplate(otp, { expiresMinutes: 10 });
<<<<<<< HEAD
  await sendMail({ to: email, ...tmpl });
=======
  if (!(await sendMail({ to: email, ...tmpl }))) {
    throw new APIError("SERVICE_UNAVAILABLE", { ... });
  }
>>>>>>> a52e2566 (patch v0.6.6)
  return;
}
```

## Impact

- `bun run lint` fails (3 errors).
- `bun run test` in `@repo/api`: **48 test files fail to load**. esbuild cannot parse the file, so every test that transitively imports `auth.ts` — via `src/middleware/auth.ts` and up — dies at transform time. With the markers removed the same suite is 354/355 files and 4158 tests passing.
- Anything that bundles the API is broken, so this blocks a build from a clean checkout.

`packages/core` is unaffected.

## Where it came from

`74ec3f72` ("patch v0.6.6") introduced it — its parent `f9a17fc5` is clean, so this is a fresh regression, not long-standing. It reached `main` through #586, and is also still on `mail/562-fixes-and-inbound-foundation`. No tag contains it.

Worth flagging separately: **#586's checks were red when it merged** — `Typecheck` fail and `Test` fail, with only `Test webmail server` green ([run 31834622029](https://github.com/oblien/openship/actions/runs/31834622029)). CI caught this exactly as designed; the merge went in over it. Whatever the fix here, the branch protection gap is the thing that let a non-compiling `main` happen, and it will happen again as-is.

The referenced hash `a52e2566` does not exist in the repository, which fits a local rebase/merge that was resolved-by-committing-the-markers and pushed.

## Fix

Keep the `>>>>>>>` side — fail loudly. That matches the `forget-password` branch immediately below it in the same function, which already checks `sendMail`'s result and throws `SERVICE_UNAVAILABLE`, and the comment there spells out the reasoning: being told to check your inbox for a code that was never sent is worse than an error. The `HEAD` side (fire-and-forget) is the older behavior that `v0.6.6` was in the middle of replacing.

PR follows.

AbdullahM07:
Fixes #594.

`74ec3f72` ("patch v0.6.6") committed unresolved merge conflict markers into `apps/api/src/lib/auth.ts`, in the `emailOTP` plugin's `sendVerificationOTP`. `main` has not typechecked since.

## Before

```
$ bun run lint
src/lib/auth.ts(418,1): error TS1185: Merge conflict marker encountered.
src/lib/auth.ts(420,1): error TS1185: Merge conflict marker encountered.
src/lib/auth.ts(428,1): error TS1185: Merge conflict marker encountered.
```

esbuild also cannot parse the file, so **48 API test files fail at transform time** — everything that reaches `auth.ts` through `middleware/auth.ts`.

## After

```
$ bun run lint          # apps/api
# clean, exit 0

$ bun run test          # apps/api
Test Files  354 passed | 1 failed (355)
     Tests  4158 passed | 1 failed | 3 skipped (4162)
```

The one remaining failure is `service-exec-isolation.test.ts` hitting the 20s test timeout under full-suite load; it passes in isolation in ~8s and is unrelated to this change. `packages/core` is 608/608.

## Which side this keeps

The fail-loudly one:

```ts
if (!(await sendMail({ to: email, ...tmpl }))) {
  throw new APIError("SERVICE_UNAVAILABLE", { ... });
}
```

That is already what the `forget-password` branch does a few lines below in the same function, and its comment states the reasoning: `sendMail` returns silently when no transport is configured, so without the result check the user is told to check their inbox for a code that was never sent. The discarded `HEAD` side was the older fire-and-forget call that v0.6.6 was in the process of replacing.

Four deleted lines, no behavior change beyond removing the markers. No test asserted the silent behavior.

## Not addressed here

#594 also notes that #586 merged with `Typecheck` and `Test` red. This PR only unbreaks `main`; the branch protection gap that allowed it is left for a separate discussion.

### DIFF
--- a/apps/api/src/lib/auth.ts
+++ b/apps/api/src/lib/auth.ts
@@ -415,9 +415,6 @@ export const auth = betterAuth({
         // types are not enabled, so they fall through and send nothing.
         if (type === "email-verification") {
           const tmpl = verifyOtpEmailTemplate(otp, { expiresMinutes: 10 });
-<<<<<<< HEAD
-          await sendMail({ to: email, ...tmpl });
-=======
           if (!(await sendMail({ to: email, ...tmpl }))) {
             throw new APIError("SERVICE_UNAVAILABLE", {
               message:
@@ -425,7 +422,6 @@ export const auth = betterAuth({
                 "email transport. Configure SMTP in Settings → Email.",
             });
           }
->>>>>>> a52e2566 (patch v0.6.6)
           return;
         }
         // Password reset. This replaced the link flow (`sendResetPassword` in the


AbdullahM07:
Superseded by b2223b9d, which landed the same fix — closing.

For the record the two resolutions are byte-identical: both go `e2d84e81..b4394ed4` on this file, keep the guarded side, and cite the same reason (a verification code that could not be sent has to surface as `SERVICE_UNAVAILABLE` rather than a silent success). Nothing here to merge.

The compile break on `main` is fixed as of 77f67dc6.

Issue #92, opened 2026-07-22
[Feature]: Canvas feature


kipavy:
Hi,

I haven't had the time to test openship yet but I love the idea of it. I'd like to know if a 'canvas' feature similar to what https://railway.com/ does is in the plan or could be possible ? I've always wanted an open-source alternative to Railway that has a comparable UX.

Thanks.

Hydralerne:
will consider thank you

ValenFelizia:
> will consider thank you

Hi! First of all, I really like what you’re building with OpenShip!

I’ve been running the project locally and exploring the codebase, and I can
genuinely see myself using it for some of my own projects in the future. I’d
also love to contribute to its development where I can.

I took a closer look at the current project and service data model to explore
what a narrowly scoped first version of this feature could look like.

OpenShip already exposes enough existing data for a read-only topology view:

- one node per existing service;
- explicit edges derived only from `service.dependsOn`;
- existing container status;
- click-through to the current service detail view.

This could be implemented without database or API changes, while deliberately
excluding service creation, edge editing, persisted positions, inferred
environment-variable relationships, domains as nodes, and staged changes.

To keep the work reviewable, I would propose splitting it into two pull requests:

1. A pure, tested topology mapper that converts services into nodes and resolved
   `dependsOn` edges, including safe handling of cycles and dangling dependency
   names.
2. A separate read-only canvas UI in the Services section, complementing rather
   than replacing the current list.

Before writing code, would this limited read-only direction be useful to the
project, and would you be comfortable with that two-PR approach?

If so, I can post a more concrete implementation outline before starting.

ValenFelizia:
Hey @Hydralerne just a quick follow-up when you have a moment, I know the project is currently focusing on bugs and stability, but would you please let me know if you are open to the scope I explained above for this feature? I'd be happy to prepare the implementation plan, but I'll hold off on writing code until you confirm that this direction or this feature all together fits the project. If not, I'll continue helping with the bugs and such.

Hydralerne:
Thanks for the detailed proposal! 
I think this would be a really valuable addition to OpenShip.

I'd definitely be happy to see it. Right now our main focus is stability and getting the core platform rock solid, but if you have the time to work on it, I'd love to review it.

The two-PR approach also sounds good and should make it much easier to review and merge. Feel free to post the implementation outline whenever you're ready.

ValenFelizia:
[...]

PR #600, opened 2026-08-16
feat(dashboard): add deterministic service topology model

ValenFelizia:
## Summary

Adds a pure, deterministic service topology mapper for the first part of the Canvas feature discussed in #92.

The mapper converts existing project services and live container state into service nodes, semantic dependency relationships, and unresolved dependency diagnostics, without introducing any UI or graph-library dependency.

## Motivation

Issue #92 agreed on a two-PR approach for introducing a read-only Canvas view.

This first PR establishes the service-domain topology contract independently from the visual layer, so the follow-up PR can build the reusable canvas on top of it without coupling service dependency semantics to React Flow or layout concerns.

The maintainer also asked that the future canvas remain generic enough to support later features such as load balancing and scaling. Keeping this mapper framework-independent leaves that separation intact.

## Related issue

Related to #92.

This does not close the issue because the visible Canvas UI will be implemented separately in PR 2.

### `apps/dashboard`

* Added `buildServiceTopology()` as a pure service topology mapper.
* Uses existing service IDs as stable node identity.
* Resolves relationships exclusively from exact `service.dependsOn` name matches.
* Represents dependency semantics explicitly as `dependentId` and `dependencyId`, independently from any future canvas `source` / `target` orientation.
* Preserves cycles and self-dependencies.
* Deduplicates repeated dependency declarations.
* Reports missing and ambiguous dependencies without creating synthetic nodes.
* Includes sorted candidate service IDs for ambiguous dependency names.
* Preserves live runtime status when available and keeps missing runtime information as `null`.
* Produces deterministic output regardless of service, dependency, or container input ordering.
* Does not mutate input data.
* Added Vitest coverage for dependency shapes, diagnostics, status behavior, determinism, and input immutability.

No React Flow, UI, layout, API, database, routing, persistence, package, or dependency changes are included in this PR.

## Verification

```bash
bun run --cwd apps/dashboard test
# PASS — 67 test files passed, 866 tests passed

bun run --cwd apps/dashboard lint
# PASS — tsc --noEmit

bunx prettier --check \
  apps/dashboard/src/lib/service-topology.ts \
  apps/dashboard/src/lib/service-topology.test.ts
# PASS — both files use Prettier code style

git diff upstream/main...HEAD --check
# PASS — no output

git status --short
# clean — no output
```
bun run test was also executed. The dashboard suite passed, but the aggregate command stopped in the unchanged apps/cli workspace on Windows with 20 failures across 7 files. The failures were confined to the unchanged apps/cli workspace and appear to involve POSIX path and generated Compose expectations when running on Windows. This PR does not modify apps/cli.

## Checklist

* [x] One change per PR — one bug, or one agreed feature, with nothing unrelated bundled in
* [x] The diff is scoped — no reformatting or lint fixes on lines I wasn't otherwise changing
* [x] A test fails without this change and passes with it (or I explained above why there isn't one)
* [ ] `bun run test`, `bun run --cwd <workspace> lint`, and `bun format` all pass locally
* [x] I understand every line of this diff and can explain it in review

### DIFF
--- a/apps/dashboard/src/lib/service-topology.test.ts
+++ b/apps/dashboard/src/lib/service-topology.test.ts
@@ -1,0 +1,372 @@
+import { describe, expect, it } from "vitest";
+
+import {
+  buildServiceTopology,
+  type ServiceTopologyContainerInput,
+  type ServiceTopologyServiceInput,
+} from "./service-topology";
+
+function service(
+  id: string,
+  name: string,
+  overrides: Partial<ServiceTopologyServiceInput> = {},
+): ServiceTopologyServiceInput {
+  return { id, name, ...overrides };
+}
+
+describe("buildServiceTopology", () => {
+  it("returns an empty topology for an empty service list", () => {
+    expect(buildServiceTopology([])).toEqual({
+      nodes: [],
+      edges: [],
+      unresolvedDependencies: [],
+    });
+  });
+
+  it("creates an isolated node using the service id as identity", () => {
+    expect(buildServiceTopology([service("svc_api", "api")])).toEqual({
+      nodes: [
+        {
+          id: "svc_api",
+          serviceId: "svc_api",
+          name: "api",
+          kind: null,
+          status: null,
+        },
+      ],
+      edges: [],
+      unresolvedDependencies: [],
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


Hydralerne:
looking forward to the rest of the feature

PR #617, opened 2026-08-18
feat(services): support recreate option on service restart

### LINKED ISSUE #615: [Bug]: Service restart does not apply updated environment variables (container is not recreated)
### Summary
Updating a service's configuration or environment variables (e.g. via `PATCH /projects/:id/services/:serviceId` or `PUT /projects/:id/services/:serviceId/env`) updates the service definition in Openship. However, invoking the service restart action (`POST /projects/:id/services/:serviceId/restart`) performs a standard container restart (`docker restart`).

Because Docker container environment variables are immutable after container creation, the restarted container continues running with its previous environment variables until a full project redeploy is triggered.

---

### Steps to Reproduce
1. Deploy a project with one or more services.
2. Update a service-level environment variable (e.g. `SERVICE_ENV_VAR=new_value`).
3. Trigger a restart for that specific service.
4. Check the environment variables inside the container (`printenv SERVICE_ENV_VAR`).

---

### Expected Behavior
The service container should be recreated so the new environment variables and updated configuration take effect.

---

### Actual Behavior
The existing container is merely restarted, so changes to environment variables or config are not applied until a new deployment cycle occurs.

---

### Suggested Resolution
- Have the service restart action recreate the container if configuration or environment variables have changed, or
- Add an explicit `recreate` parameter / action on the service endpoint.

chbndrhnns:
### Problem
Updating service environment variables (or other service configuration) in Openship updates the database records, but calling `POST /api/projects/:id/services/:serviceId/restart` performed a raw `docker restart`. Because container environment variables are fixed at container creation time, the restarted container continued running with its stale configuration until a full project redeployment occurred.

Closes #615

---

### Solution
- Updated `restartServiceContainer` to support `opts?: { recreate?: boolean }`. When `recreate: true` is passed (or when no active container exists), it provisions and recreates the service container with fresh environment and configuration via `provisionServiceContainer`.
- Updated the route `POST /api/projects/:id/services/:serviceId/restart` and controller to accept `recreate?: boolean` via query parameter (`?recreate=true`) or JSON body (`{ "recreate": true }`).
- By default (when `recreate` is omitted/false), fast container restart (`docker restart`) is preserved.
- Added comprehensive unit tests in `apps/api/test/modules/services/service-restart-recreate.test.ts`.

### DIFF
--- a/apps/api/src/modules/services/service.controller.ts
+++ b/apps/api/src/modules/services/service.controller.ts
@@ -306,9 +306,17 @@ export async function restartContainer(c: Context) {
   const ctx = getRequestContext(c);
   const projectId = param(c, "id");
   const serviceId = param(c, "serviceId");
+  const recreateParam = c.req.query("recreate");
+  let recreate = recreateParam === "true" || recreateParam === "1";
+  if (!recreate && c.req.header("content-type")?.includes("application/json")) {
+    const body = await c.req.json<{ recreate?: boolean }>().catch(() => null);
+    if (body?.recreate) recreate = true;
+  }
   try {
-    await serviceService.restartServiceContainer(ctx, projectId, serviceId);
-    return c.json({ success: true });
+    const result = await serviceService.restartServiceContainer(ctx, projectId, serviceId, {
+      recreate,
+    });
+    return c.json({ success: true, ...result });
   } catch (err) {
     const message = err instanceof Error ? err.message : "Failed to restart container";
     return c.json({ success: false, error: message }, 400);
--- a/apps/api/src/modules/services/service.routes.ts
+++ b/apps/api/src/modules/services/service.routes.ts
@@ -24,6 +24,8 @@ import * as ctrl from "./service.controller";
 import { AgentExecBody } from "../../lib/agent-exec.schema";
 import {
   CreateServiceBody,
+  RestartServiceBody,
+  RestartServiceQuery,
   SetServiceEnvVarsBody,
   SyncServicesBody,
   UpdateServiceBody,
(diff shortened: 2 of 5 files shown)


Hydralerne:
Thanks for both the issue and the PR — #615 is a real bug and you diagnosed the cause exactly right: a container's environment is fixed at creation, so `docker restart` cannot apply it.

I'm closing this in favour of a different fix, but not because the idea was wrong. Three things came up in review:

**1. It doesn't compile.** `query` isn't a field on `RouteSpec` (`apps/api/src/lib/route-permission.ts:416` — it's `PermissionSpec | PublicSpec`, and neither declares one), so `service.routes.ts` fails `tsc --noEmit`, which CI gates on via `bun run --cwd apps/api lint`. It never surfaced here because the fork's CI run is still sitting at `action_required`, so no check has ever executed on the branch. Related: `secureRouter` only auto-wires `body`, never `query`, so `RestartServiceQuery` would reach no consumer even once it compiled — the controller's own `c.req.query("recreate")` is what actually makes `?recreate=true` work.

**2. `body: RestartServiceBody` breaks the CLI.** Declaring a body makes `secureRouter` mount `tbValidator("json", …)` ahead of the handlers. Hono's json validator only skips parsing when there is no JSON content-type — with one present it calls `c.req.json()` and throws `400 Malformed JSON in request body` on an empty body. `openship service restart <svc>` POSTs with no body (`apps/cli/src/commands/service.ts`), and the CLI's api-client sets `Content-Type: application/json` unconditionally (`apps/cli/src/lib/api-client.ts:37`), so every CLI restart would start 400ing. The dashboard is unaffected — it only sets that header when there is a body — which is why this wouldn't show up in manual testing.

**3. `recreate: true` would not have applied project-level env.** `mergeServiceDeployEnv` layers `{...project, ...frozen, ...inline, ...service}` (`apps/api/src/modules/deployments/compose/deploy.service.ts`), so the active deployment's frozen `envVars` snapshot shadows the live project rows. Service-scoped env does land, so your exact repro works — but a project-level variable would have kept its old value while the endpoint returned success. And when the active deployment came from a rollback, `frozenWins` moves the frozen layer last and shadows the service rows too.

**Separately**, the platform already has this path: `POST /api/deployments {"refresh": true, "serviceIds": ["<id>"]}` — re-apply current env to the active deployment, no git pull, no rebuild, recreating env-changed services from their existing images. It is what the dashboard's env editor has always used (the **Apply** button after you save env vars), it is exposed as `openship deploy --refresh --service-ids <id>`, and it auto-detects which services are env-dirty. So the capability existed; `restart` just never mentioned it.

**What we did instead:** kept `restart` a cheap bounce and made it stop lying. With pending env changes it now answers `409 SERVICE_CONFIG_STALE`, naming the drifted keys and pointing at the refresh deploy; `?force=true` bounces anyway, because a crash-looping container has to stay bounceable. The dirtiness check reuses the same resolver the smart-redeploy router uses (extracted to one module), so the refusal and the refresh it recommends can never disagree about what is stale.

We went with refusing rather than your option 1 (auto-recreate) for the reason your own PR description hints at: recreating turns a restart into a container replacement — brief downtime, new private IP — and a caller asking for a bounce shouldn't get that implicitly.

Really useful report, and the diagnosis saved us the hard part. Thank you.

PR #625, opened 2026-08-18
fix(adapters): log that HTTP is live while SSL is still provisioning

Devenio:
## Summary

When a custom-domain route is registered, the deploy log now says the site is live on HTTP while the certificate is still being issued (`HTTPS in ~1 min`), instead of a silent ~1 minute HTTPS gap that operators filed as broken SSL.

## Motivation

A new project's route is registered with `tls: true`, but nginx only emits the 443 block once the cert exists (`route.tls && certsExist(domain)`). The pipeline is correct — cert lands, vhost re-renders, HTTPS works — but during that ~1 minute window the site answers HTTP and nothing (or a bootstrap self-signed cert) on HTTPS. Issuance is best-effort by design (domains never fail a deploy), which is why the *silence* was the bug: two people reported it as broken SSL. Documented in `TODO.md` under “SSL provisioning is invisible in the deploy log.”

## Related issue

None

## Changes

- **packages/adapters** — After `registerRoute`, if `provisionSsl && ssl`, log `Route live on HTTP for ${hostname} — provisioning the certificate, HTTPS in ~1 min`, then call `provisionCert` as before. No log when `provisionSsl` is off.
- **apps/api** — Same wording on the compose static-service path that duplicated the SSL step outside `registerResolvedRoutes`. Issuance stays best-effort: a failed cert never fails the deploy.

## Verification

```bash
$ bunx vitest run src/runtime/route-registration.test.ts
# cwd: packages/adapters

 RUN  v4.0.18

 ✓ src/runtime/route-registration.test.ts (13 tests) 11ms

 Test Files  1 passed (1)
      Tests  13 passed (13)

$ bun run --cwd packages/adapters lint
$ tsc --noEmit
# exit 0
```

The new describe `registerResolvedRoutes — SSL provisioning is visible in the deploy log`:

- fails without the log line (asserts the HTTP-live message after `registerRoute`, then `provisionCert`)
- does not log `HTTPS in ~1 min` and does not call `provisionCert` when `provisionSsl` is off

Before: deploy log showed `Checking SSL for example.com...` (or nothing on first compose deploys) while HTTPS was still dead. After: the log states HTTP is live and HTTPS is coming in ~1 min; existing `SSL certificate active` / Action Required lines still fire when the cert lands or fails.

## Checklist

- [x] `bun run --cwd packages/adapters lint` and the route-registration tests pass locally (full `bun run test` not re-run; this PR does not touch the dashboard)

### DIFF
--- a/apps/api/src/modules/deployments/compose/deploy.service.ts
+++ b/apps/api/src/modules/deployments/compose/deploy.service.ts
@@ -1651,9 +1651,13 @@ export async function deployComposeServices(
           // Kept OUT of the try above so a cert failure can never be reported as a
           // route-registration failure.
           if (route.provisionSsl) {
-            logger.log(`Checking SSL for ${route.hostname}...\n`, "info", {
-              serviceName: svc.name,
-            });
+            logger.log(
+              `Route live on HTTP for ${route.hostname} — provisioning the certificate, HTTPS in ~1 min\n`,
+              "info",
+              {
+                serviceName: svc.name,
+              },
+            );
             await routeContext.trackedSsl.provisionCert(route.hostname).catch((err) => {
               logger.log(
                 `SSL provisioning failed for ${route.hostname} (route is up on HTTP, retry from ` +
--- a/packages/adapters/src/runtime/route-registration.test.ts
+++ b/packages/adapters/src/runtime/route-registration.test.ts
@@ -4,7 +4,11 @@ import { explainEdgeDown } from "../system/edge-exec-error";
 
 const logger = { log: vi.fn(), step: vi.fn() } as any;
 const routeTarget = { targetUrl: "http://127.0.0.1:12345" };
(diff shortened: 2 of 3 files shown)


Hydralerne:
Thanks for improving certificate-provisioning feedback. I compared this with #627, #580 and current main during the fix review.

The updated #580 is now merged into the integration branch for #928. It places the message once in the shared tracked SSL provider used by both deployment pipelines, confirms that the certificate is actually missing before describing the HTTP interval, and avoids an unsupported one-minute ETA. It retains main's newer certificate recovery and failure-state handling. All 137 focused routing/SSL cases and the full PR CI passed.

Closing this overlapping version to avoid duplicate messages and an HTTP-only claim on deployments that already have a certificate. The consolidated fix is awaiting merge to main in #928.

PR #653, opened 2026-08-20
fix(dashboard): allow scrolling in CustomSelect and project dropdown menus

### LINKED ISSUE #652: [Bug]: Dropdown to connect a Supabase DB to a project does not scroll
### Problem Description
When connecting a Supabase database (or any database/source app) to a project via the **"Use in a project"** modal, the **Target project** dropdown list cannot be scrolled when there are multiple projects in the workspace.

### Root Cause
In `CustomSelect.tsx`, the portal menu element has `className="... overflow-hidden ..."` and an inline `maxHeight` style. The inner options container has `max-h-full overflow-y-auto`.
Because the parent container has `height: auto` and lacks `flex flex-col`, the inner element's `max-h-full` (`max-height: 100%`) resolves against an indefinite height and expands to fit all items. Consequently, the inner container never overflows itself (so `overflow-y-auto` never triggers scrolling), while the outer container's `overflow-hidden` and `maxHeight` clips the list, preventing users from seeing or scrolling to options past the initial visible height.

Additionally, in `ProjectFilter.tsx`, `overflow-hidden` was declared after `overflow-y-auto` on the menu element, overriding the scroll behavior.

### Steps to Reproduce
1. Deploy a Supabase instance (or have a source database app with connection outputs).
2. Have multiple target projects in the workspace (e.g. > 5 projects).
3. Open the Supabase app, navigate to the connection details, and click **"Use in a project"**.
4. Click the **"Project"** / **"Select a project…"** dropdown.
5. Attempt to scroll down to view and select projects further down the list.
6. The list is clipped and does not scroll.

### Solution
1. Add `flex flex-col` to the outer portal container in `CustomSelect.tsx`.
2. Change the inner container in `CustomSelect.tsx` to `min-h-0 flex-1 overflow-y-auto` so flexbox properly constrains the inner container's height to the available space within `maxHeight`, allowing `overflow-y-auto` to activate and scroll.
3. Remove overriding `overflow-hidden` from `ProjectFilter.tsx`.

chbndrhnns:
Fixes #652

### Problem
When connecting a Supabase database to a project via the "Use in a project" modal (`UseInProjectModal`), or when using `CustomSelect` / `ProjectFilter` with a long list of items, the dropdown list cannot be scrolled and gets clipped.

### Cause
In `CustomSelect.tsx`, the portal menu container had `overflow-hidden` and `maxHeight`, but was a standard block container without flex layout. The child options container used `max-h-full` (`max-height: 100%`), which evaluates to `none` against an indefinite parent `height: auto`. The child expanded to the full content height of all options and never overflowed its own box, while the parent's `overflow-hidden` and `maxHeight` clipped the menu without enabling scrollbars.

In `ProjectFilter.tsx`, trailing `overflow-hidden` was overriding `overflow-y-auto`.

### Fix
- Added `flex flex-col` to the outer container in `CustomSelect.tsx`.
- Changed the inner options list to `min-h-0 flex-1 overflow-y-auto` so flexbox properly constrains the options container to the available `maxHeight` space, enabling scrolling.
- Removed conflicting `overflow-hidden` from `ProjectFilter.tsx`.
- Added render regression test for `CustomSelect`.

### DIFF
--- a/apps/dashboard/src/app/(dashboard)/deployments/components/ProjectFilter.tsx
+++ b/apps/dashboard/src/app/(dashboard)/deployments/components/ProjectFilter.tsx
@@ -51,7 +51,7 @@ export const ProjectFilter: React.FC<ProjectFilterProps> = ({
       </button>
 
       {isOpen && (
-        <div className="absolute top-full start-0 z-50 mt-2 max-h-96 w-72 overflow-y-auto rounded-xl border border-border/50 bg-popover overflow-hidden">
+        <div className="absolute top-full start-0 z-50 mt-2 max-h-96 w-72 overflow-y-auto rounded-xl border border-border/50 bg-popover">
           {/* All Projects Option */}
           <button
             onClick={() => {
--- a/apps/dashboard/src/components/ui/CustomSelect.tsx
+++ b/apps/dashboard/src/components/ui/CustomSelect.tsx
@@ -161,7 +161,7 @@ export function CustomSelect<T extends string>({
         <div
           ref={menuRef}
           role="listbox"
-          className="fixed z-[10050] overflow-hidden rounded-2xl border border-border/50 bg-popover shadow-xl shadow-black/[0.08]"
+          className="fixed z-[10050] flex flex-col overflow-hidden rounded-2xl border border-border/50 bg-popover shadow-xl shadow-black/[0.08]"
           style={{
             left: menuPosition.left,
             width: menuPosition.width,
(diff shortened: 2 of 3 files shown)


Hydralerne:
I really appreciate your amazing help, but please don't abuse the tests, its simple component fix no need for 3 files of tests, please cleanup the tests

chbndrhnns:
I kept a single regression test.

PR #681, opened 2026-08-21
feat(services): fast env apply after save — restart without rebuild

### LINKED ISSUE #669: feature: add publish env button after update env / updated env without need to redeploy it, as fast env update in services / project

Rish-it:
Fixes #669

## Summary

Services panel now offers a fast **Apply** after saving environment variables on a service with a live deployment. Apply triggers a refresh deployment that recreates only that service from its existing image — no rebuild, same as the project-level fast apply.

## Changes (2 atomic commits)

1. **`feat(dashboard): add per-service env apply trigger payload helper`** — pure helper (`service-env-apply.ts`) returning `{ projectId, refresh: true, serviceIds: [serviceId] }`, the contract the API's refresh path already honors. Unit tests pin the payload shape and that it never pulls source.
2. **`feat(services): offer restart-only env apply after env save`** — `ServiceDetailPanel` swaps Save → Apply once env is saved and a deployment is live. Clicking routes to `/build/<id>` for progress, matching the panel's existing redeploy handler (same response guard, same navigation). Pending offer resets on service switch so it can't leak across services. Editing again reverts to Save.

Reuses existing pieces only: backend `refresh + serviceIds` path, panel's `deployApi` import and `RotateCw` icon, i18n keys from `projectSettings.json` (`envVars.applyButton`, `envVars.toast.applyFailed*`). No new dependencies, no new translation entries.

## Testing

- New unit tests: `service-env-apply.test.ts`
- Full dashboard suite: 1048 passed
- `tsc --noEmit`: clean


Hydralerne:
Reviewed against current main: the fast service environment apply flow is already implemented in the shared engine and Environment panel, including the environment ownership fixes in #926.

The Environment panel compares saved values with the running service and calls the existing service `apply-env` operation. The engine recreates only that service from its current image under the project runtime lock, preserves settings and concurrent saves, and records the replacement container. It does not clone, build, or queue a deployment. Unsupported runtimes explicitly direct the user to Redeploy.

The 22 rendered environment/apply tests and 12 engine apply tests pass. Reintroducing this PR's separate `refresh` deployment and transient pending flag would duplicate the newer flow and weaken its state handling. Closing as superseded by main.

PR #683, opened 2026-08-21
fix(deploy): avoid host ports pinned to other projects in compose deploys

### LINKED ISSUE #682: [Bug]: Compose deploys collide with host ports pinned to offline containers of other projects
### Summary
When deploying a Compose stack (`deploy.service.ts`), host port allocation (`allocateHostPort`) only populated `usedHostPorts` with services from the current stack's previous deployment. Unlike single-service deployments in `build-pipeline.ts`, it did not query other projects in the organization for their pinned host ports.

Because `allocateHostPort` falls back to scanning live listening ports on the host via TCP (`scanPorts`), any container from another project that is temporarily stopped, restarting, or crashed is not listening on TCP and appears free. The Compose deployment then allocates that port. When the offline container restarts or is started by Docker's restart policy, Docker fails to bind the port:
```
driver failed programming external connectivity on endpoint openship-xxx-yyy: Bind for 127.0.0.1:20007 failed: port is already allocated
```
and the offline service enters a permanent restart crash loop.

### Fix
1. Extract `listOrgPinnedHostPorts` to collect active host ports pinned across all other projects in the organization (both single-service `project.hostPort` and compose `service_deployment.hostPort` rows from active deployments).
2. Seed `usedHostPorts` in `deploy.service.ts` with these organization-pinned ports while preserving each service's own `carried` port as preferred.
3. Update `build-pipeline.ts` to use `listOrgPinnedHostPorts` so single-service deploys also avoid compose stack ports.

chbndrhnns:
Closes #682

### Problem
When deploying a Compose stack (`deploy.service.ts`), loopback host port allocation only avoided ports belonging to sibling services in the current project. If a container belonging to another project was stopped, restarting, or crashed during the deployment, host port scanning did not detect it as listening on TCP. The deploy then allocated that port to the Compose service. When the offline container attempted to start, Docker rejected the bind (`Bind for 127.0.0.1:<port> failed: port is already allocated`), causing a permanent restart failure.

Additionally, `build-pipeline.ts` previously only inspected `project.hostPort` when computing `avoid`, missing active host ports belonging to Compose stacks.

### Solution
- Created `listOrgPinnedHostPorts` in `apps/api/src/modules/deployments/pinned-host-ports.ts` to aggregate all active host ports pinned across the organization (both single-service `project.hostPort` and Compose stack active `service_deployment.hostPort` rows).
- Seeded `usedHostPorts` in `deploy.service.ts` with organization-pinned ports and ensured a service's own `carried` port is not in its own `avoid` set so it can be reused as preferred.
- Updated `build-pipeline.ts` to use `listOrgPinnedHostPorts`.
- Added unit tests in `pinned-host-ports.test.ts` and updated `carried-host-port.test.ts`.

### DIFF
--- a/apps/api/src/modules/deployments/build-pipeline.ts
+++ b/apps/api/src/modules/deployments/build-pipeline.ts
@@ -85,10 +85,23 @@ import {
 import { firePreDeployBackups } from "../backups/triggers/pre-deploy";
 import { buildBackgroundContext } from "../../lib/request-context";
 import * as sessionManager from "./session-manager";
-import { onFailure, onSuccess, onCancelled, reportPipelineError, setDeploymentStatus, routeIssuesWarning, type LifecycleContext } from "./deployment-lifecycle";
+import {
+  onFailure,
+  onSuccess,
+  onCancelled,
+  reportPipelineError,
+  setDeploymentStatus,
+  routeIssuesWarning,
+  type LifecycleContext,
+} from "./deployment-lifecycle";
 import { auditPorts } from "./port-audit.service";
+import { listOrgPinnedHostPorts } from "./pinned-host-ports";
 import { verifyDeployedContainers } from "./stability-audit.service";
-import { resolveReadinessGate, runReadinessGate, type ResolvedReadinessGate } from "./readiness-gate";
+import {
+  resolveReadinessGate,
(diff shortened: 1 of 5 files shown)


Hydralerne:
fixed already sorry

PR #684, opened 2026-08-21
fix(dashboard): keep service rows during refetch (#666)

### LINKED ISSUE #666: service page had poor state update, showing the skelton for every action which is annoying not smoth

Rish-it:
Fixes #666

## Problem
The services tab collapsed into its skeleton pulse-cards on **every** action (create service, launch, drift fix) and every tab switch, then popped back when the response landed. Reported as "poor state update… annoying not smooth".

## Root cause
Two loading flags gated the entire tab (`ServicesTab.tsx`):

```ts
const loading = servicesData.isLoading || containersLoading;
```

- `refreshServices()` set `isLoading: true` unconditionally on every refetch — and its error path wiped the list to `[]`, blanking the tab on transient failures.
- `containersLoading` flipped to `true` at the top of every `fetchData()` call **and on every remount of the tab** (the tab is conditionally rendered per `activeTab`, so switching back to Services re-skeletoned even with cached data).

The codebase already solved this exact class of bug for project info — `begin-fetch-state.ts` documents how "every invalidation unmounted the page's whole subtree" and fixed it with stale-while-revalidate. Services/containers never got that treatment.

## Fix
Port the same semantics:

- `refreshServices` reports `isLoading` only when the list is empty (nothing on screen to keep), and keeps the previous list when a refetch fails.
- The tab's skeleton gate requires emptiness: `servicesData.isLoading || (containersLoading && services.length === 0)`. Rows render fine without container data — status falls back per service (`ct?.status ?? …`) — so container chips simply populate when the read lands (~20ms locally).

Consumers of `servicesData.isLoading` (`LogsSettings`, `OverviewTab`, `DomainSettings`) inherit the calmer semantics automatically: stale data stays visible during revalidation instead of flashing placeholders.

## Verification
- `tsc --noEmit` clean
- dashboard suite: 87 files / 1046 tests passing
- Manual: services list persists through create/launch/drift actions and tab switches; first cold load still shows the skeleton; failed refetch keeps rows + shows error state

### DIFF
--- a/apps/dashboard/src/app/(dashboard)/projects/[id]/components/ServicesTab.tsx
+++ b/apps/dashboard/src/app/(dashboard)/projects/[id]/components/ServicesTab.tsx
@@ -59,7 +59,12 @@ export const ServicesTab = () => {
     () => sortServicesByPublicFirst(servicesData.services),
     [servicesData.services],
   );
-  const loading = servicesData.isLoading || containersLoading;
+  // Skeleton only when there is nothing to show. containersLoading flips on
+  // every refetch (and every remount of this tab), so OR-ing it raw flashed
+  // the full-tab skeleton on every action and tab switch (#666) — rows render
+  // fine without container data (status falls back per service).
+  const loading =
+    servicesData.isLoading || (containersLoading && services.length === 0);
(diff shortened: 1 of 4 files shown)


Hydralerne:
thank you bro

PR #690, opened 2026-08-22
fix(dashboard): invalidate project caches after environment mutations (#657)

### LINKED ISSUE #657: Adding new environment not update the ui state and need refresh, and deleting project from environment does not update the state too
issue with client state managemnt of new environments

Rish-it:
Closes #657

## Problem

Adding a new environment didn't update the UI until a hard refresh, and a deleted environment lingered in every switcher the same way.

## Root cause

Two writers into one `environments` state:

- `ProjectSettingsContext.tsx` re-seeds the list from the bundled `/info` payload whenever `projectInfo` changes
- `refreshEnvironments()` patches it from the dedicated list endpoint

`createEnvironment` only patched the second, so the module-level `infoCache` kept the stale bundle — the next `projectInfo` emit wrote the old list back over the fresh patch. Delete went through `handleDeleteProject` (which doubles as the environment delete — the server tears down THIS environment and only soft-deletes the app row once its last one is gone) with no invalidation at all, and it navigates home on success, so the dead entry survived in every *sibling's* cached bundle.

The list ships inside EVERY environment's `/info` payload, so sibling caches go stale on any mutation — invalidating only the current id would leave the bug one click away.

## Fix

Two commits, each self-consistent green:

1. **Create** (`400b52c2`) — `createEnvironment` calls `invalidateProjectCaches(id)` before returning, then also invalidates known sibling ids. `refreshEnvironments()` stays as the immediate correct patch.
2. **Delete** (`67671ef2`) — all three success exits of `handleDeleteProject` invalidate: full ok, unrecoverable-partial, and the cross-tab 404 in catch, where another tab already deleted the row so this tab's bundles are guaranteed stale. Real failures (409 active-work, deletion-in-progress, teardown-failed) keep caches as-is.

Reuses the existing tool for this exact situation (`useProjectEndpoints.ts` — "Drop ALL caches for an id AND notify subscribed hooks to refetch"), already used by AdvancedSettings, RoutingUnsyncedCallout and useDeploymentBuild.

No flash risk: same-id refetches are stale-while-revalidate (`beginFetchState` keeps what's on screen), and eviction of unmounted siblings is pure cache-drop — no request storm.

## Test plan

- New `project-settings-environments.test.ts` pins the wiring in source (same no-jsdom idiom as `advanced-migration-session.test.ts`): create-path invalidation + ordering, sibling coverage, helper definition, ≥3 success exits, 404 ordering
- `vitest run`: **1058/1058 passed** (×2 clean runs)
- `tsc --noEmit`: clean (×2)

### DIFF
--- a/apps/dashboard/src/app/(dashboard)/projects/[id]/[[...slug]]/page.tsx
+++ b/apps/dashboard/src/app/(dashboard)/projects/[id]/[[...slug]]/page.tsx
@@ -32,7 +32,11 @@ import { DraftProjectView } from "../components/DraftProjectView";
 import { environmentErrorMessage, environmentWizardHref } from "../components/environment-next";
 import { getProjectStatus } from "@/utils/project-status";
 import { useProjectSettings } from "@/context/ProjectSettingsContext";
-import { useProjectInfo, PROJECT_INFO_NOT_FOUND } from "@/hooks/useProjectEndpoints";
+import {
+  invalidateProjectCaches,
+  useProjectInfo,
+  PROJECT_INFO_NOT_FOUND,
+} from "@/hooks/useProjectEndpoints";
 import Link from "next/link";
 import { useRouter, useSearchParams } from "next/navigation";
 import { useToast } from "@/context/ToastContext";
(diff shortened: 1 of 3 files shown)


Hydralerne:
already fixed sorry

Rish-it:
my bad Thanks !!

PR #703, opened 2026-08-24
fix(ssh-transport): ensure consistent timeouts for streamlocal probe …

### LINKED ISSUE #698: Desktop app: self-hosted SSH deploy can't reach remote Docker daemon ("Cannot reach Docker daemon: No such file") — works fine under Node, likely #10 regression
## Summary

Deploying a Next.js project to a self-hosted server over SSH fails immediately at the Building stage from the **desktop app**, with:

```
Docker build failed: Cannot reach Docker daemon: No such file. Check that SSH credentials are correct, Docker is running on the server, and the SSH user can access the Docker daemon (in the `docker` group or root). The transport uses SSH socket forwarding where available and falls back to `docker system dial-stdio` over an exec channel otherwise.
```

This reproduced consistently (4/4 attempts) from the desktop app, and disappeared completely when the exact same project/server/deploy config was redeployed by running the app's own binary directly under Node instead (`/Applications/Openship.app/Contents/MacOS/openship`, bypassing the Electron/Bun wrapper) — the deploy then proceeded past this point (successful git clone, Docker build context packaging), confirming the transport and the target server are fine.

This looks like the same root cause as #10 (Bun's HTTP client not honoring the SSH-tunneled transport for `this.docker.*` / Docker-socket calls) resurfacing on the newer transport path described in the current error message (SSH socket forwarding + `docker system dial-stdio` fallback) — possibly a partial fix or regression, since #10 was closed as fixed.

## Environment

- Openship desktop app on macOS (Apple Silicon)
- Target: freshly provisioned Ubuntu 22.04 server, added under Servers, auto-configured by the app (installed Docker 29.7.2 + Git + rsync successfully)
- SSH: dedicated ed25519 key, `openship` user in the `docker` group (added manually after first failure, no change in outcome)
- Framework: Next.js, `runtime: sandboxed (Docker container)`, build strategy `docker (remote Docker daemon via SSH)`

## What I verified server-side (all pass)

- `docker version` / `docker context ls` over a plain SSH exec as the `openship` user: works
- SSH stream-local port forwarding of `/var/run/docker.sock` from my machine to the server (`ssh -L localsock:/var/run/docker.sock`): works, socket created successfully
- `docker system dial-stdio` over a plain SSH exec channel: no error returned
- `sshd -T`: `allowstreamlocalforwarding yes`, `disableforwarding no`
- Docker socket: `srw-rw---- root:docker`, group-accessible

## Steps to reproduce

1. Add a fresh Ubuntu server via SSH (key auth), let auto-configuration install Docker.
2. Create a project from a GitHub repo (Next.js), deploy target = that server.
3. Click Deploy from the **desktop app**. Fails at Building with the error above, in ~5-7s (before any clone/build output).
4. Quit the desktop app, run `/Applications/Openship.app/Contents/MacOS/openship` directly (plain Node process, same local data dir), open the local dashboard it serves, redeploy the same project/config from there.
5. This time it proceeds normally (git clone succeeds, build context packaged) — fails later at a separate, unrelated step (rsync key transfer, reported in a separate issue).

## Suggested next step

Given #10's own suggested fix #1 ("run/ship `openship-api` under Node instead of Bun-compiling it") appears to fully resolve this when done manually, it seems like the Bun-compiled desktop binary may not actually be running the API under Node yet, or a Bun-specific code path was reintroduced after #10. Happy to provide more logs/environment details if useful.

AbdullahM07:
Fixes GitHub issue #698: 'Desktop app: self-hosted SSH deploy can't reach remote Docker daemon'

## Summary
This PR fixes an inconsistency in SSH transport timeout values that caused intermittent 'can't reach remote Docker daemon' errors in self-hosted deployments. The connection probe used an 8-second timeout while actual requests used a 3-second timeout, leading to scenarios where the probe would succeed but actual requests would fail.

## Motivation
The SSH transport had mismatched timeout values between connection validation and actual usage:
- Connection probe (during initialization): 8-second timeout (STREAMLOCAL_PROBE_TIMEOUT_MS)
- Actual requests: 3-second timeout (STREAMLOCAL_DATA_VERIFY_TIMEOUT_MS)

This inconsistency could cause:
1. Probe to succeed (waiting up to 8 seconds for streamlocal to work)
2. Actual requests to fail (timing out after only 3 seconds)
3. Result: Deployments fail with 'can't reach remote Docker daemon' despite the probe indicating the connection was usable

This was particularly problematic in environments where streamlocal channel establishment took between 3-8 seconds.

## Changes
- Removed unused 8-second probe timeout constant (STREAMLOCAL_PROBE_TIMEOUT_MS)
- Updated comment for STREAMLOCAL_DATA_VERIFY_TIMEOUT_MS (3 seconds) to clarify it's used for both probe and real requests
- Fixed decideMode function to use consistent 3-second timeout for:
  * Streamlocal channel open operation
  * Streamlocal data flow verification
  * Updated timeout messages to reflect the 3-second value

## Verification
- All existing tests pass: `npx vitest run packages/adapters/src/runtime/docker-ssh-agent.test.ts` (11/11 tests passed)
- No breaking changes to existing functionality
- Fix is minimal and focused, addressing only the timeout inconsistency
- Verified the changes resolve the core issue by ensuring consistency between probe and actual request timeouts

## Related issue
Fixes #698

### DIFF
--- a/packages/adapters/src/runtime/docker-ssh-agent.ts
+++ b/packages/adapters/src/runtime/docker-ssh-agent.ts
@@ -18,17 +18,13 @@ import { safeErrorMessage, withTimeout } from "@repo/core";
 const DEFAULT_REMOTE_DOCKER_SOCKET_PATH = "/var/run/docker.sock";
 const resolvedDockerSocketPathCache = new WeakMap<DockerConnectionOptions, Promise<string>>();
 
-// One-time streamlocal capability probe per bridge. Short so the Bun-compiled
-// desktop (where streamlocal hangs) falls through to dial-stdio quickly instead
-// of stalling behind the 25s reachability cap on every connection.
-const STREAMLOCAL_PROBE_TIMEOUT_MS = 8_000;
 // Per-CHANNEL data-flow verification for a real bridge request (see
-// `bridgeClient`). The one-time probe above only proves the FIRST channel on a
-// connection can move data; some sshd/ssh2 combinations then open every later
-// channel "dead" (open resolves, no byte ever crosses). Bound both the open and
-// the first-byte wait with this — deliberately short so most of the caller's
-// reachability budget (e.g. the 25s migration-scan cap) is left for the
-// dial-stdio fallback to actually complete when a channel proves dead.
+// `bridgeClient`). Also used for the initial streamlocal probe to ensure
+// consistency between probe and real requests — some sshd/ssh2 combinations
+// then open every later channel "dead" (open resolves, no byte ever crosses).
+// Deliberately short so most of the caller's reachability budget (e.g. the
+// 25s migration-scan cap) is left for the dial-stdio fallback to actually
+// complete when a channel proves dead.
 const STREAMLOCAL_DATA_VERIFY_TIMEOUT_MS = 3_000;
 // How long a dial-stdio channel may say nothing before we commit the socket to it anyway.
 // NOT a verification gate — there is no transport left to fall back to. It is a REPORTING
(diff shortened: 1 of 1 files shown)


Hydralerne:
Reviewed this against current main. The exact timeout correction is already in `8f696630`: both the initial streamlocal probe and each real channel use `STREAMLOCAL_DATA_VERIFY_TIMEOUT_MS = 3000` for channel opening and data verification. The old eight-second constant is gone.

The current 13-test Docker SSH bridge suite passes, including a regression that refuses a probe exceeding the real-channel verification window. Issue #698 is already closed as fixed. Closing this PR as superseded by main; merging it would only duplicate the existing change.

PR #761, opened 2026-08-30
Add PocketBase to the app catalog

### LINKED ISSUE #762: Add PocketBase to the apps catalog
[PocketBase](https://pocketbase.io) is an MIT-licensed backend built with Go and SQLite. One executable provides a database, realtime subscriptions, user management, file storage, an API, and an admin dashboard. The [GitHub repository](https://github.com/pocketbase/pocketbase) has 60,887 stars.

It gives OpenShip users a small self-hosted backend when they do not need a multi-service database and authentication stack. A typical instance uses around 50–100 MB of RAM under light load. PocketBase also maintains an [official benchmark suite](https://github.com/pocketbase/benchmarks) covering API workloads on low-cost VPS instances with a roughly 180 MB test database.

Please add PocketBase to the Install App catalog. The implementation is in #761 and has been validated on an OpenShip host.

fcakyon:
OpenShip has no small, single-container auth and database option. [PocketBase](https://pocketbase.io) is an MIT-licensed backend with [60k+ GitHub stars](https://github.com/pocketbase/pocketbase) that combines auth, realtime SQLite data, file storage, REST APIs, and an admin dashboard in one Go service. This catalog app gives users a lighter alternative to multi-service Supabase or Neon stacks, typically using about 50–100 MB of RAM depending on workload.

- Builds PocketBase 0.40.1 for AMD64 or ARM64 from the official archive and verifies its published SHA-256 checksum
- Uses a two-stage Alpine build so download tools stay out of the runtime image
- Fixes Docker named-volume ownership at the responsible root entrypoint, then immediately runs PocketBase as the unprivileged pocketbase user
- Uses a volume-persistent marker with one-time superuser upsert, so clean installs and preexisting volumes both bootstrap while later restarts preserve password changes
- Persists /pb/pb_data and exposes the API, admin dashboard, generated connection outputs, health check, and 30-second shutdown grace period
- Adds the PocketBase Simple Icons mark to the dashboard catalog
- Closes #762

Validated against OpenShip 0.6.9: the checksum build completed, generated credentials authenticated through the PocketBase API, the admin route resolved to login rather than first-run setup, and the same volume, marker, and credentials survived a container restart. Catalog validation passes 27/27, core and dashboard checks pass, and the generated catalog is in sync.


fcakyon:
@Hydralerne are you okay with new app template contributions?

Hydralerne:
of course

PR #768, opened 2026-08-30
fix(server): require Docker for remote setup

### LINKED ISSUE #767: fix(server setup): require Docker before installing Edge
## Problem

The new-server setup can omit Docker from its prerequisite results when the OpenShip control plane runs with `DEPLOY_MODE=bare`. The UI then shows other components and can attempt Edge installation even though the managed Edge is an `openship-edge` container and remote application deployments use Docker.

`resolveRequiredComponents()` in `server-check.controller.ts` currently derives requirements from the control plane's deployment mode:

- `docker` control plane: Docker + Git
- `bare` control plane: Git only

That environment describes how OpenShip itself runs, not what a remote deployment target needs.

## Proposed scope

- Treat Docker and Git as required components for managed remote deployment servers regardless of the control plane deployment mode.
- Keep using the existing Docker health check and installer; no new privileged installation mechanism is needed.
- Ensure automatic setup installs Docker before Edge.
- Add regression coverage for a bare-mode control plane checking a remote server without Docker.
- Keep the manual setup UI explicit: Docker should appear as missing with the existing install action rather than allowing Edge setup to fail later.

## Expected result

Creating or checking a deployment server always reports Docker as a prerequisite. Automatic setup can install it through the existing installer, and Edge installation is not attempted before Docker is available.

## Implementation note

The smallest fix appears to be removing the control-plane `DEPLOY_MODE` dependency from the remote-server required-component resolver, plus enforcing dependency order/guarding Edge installation if necessary.

luckycrm:
## Summary

Makes Docker an explicit required prerequisite for every managed remote deployment server, including when the OpenShip control plane itself runs in bare mode.

## Motivation

The server health-check endpoint derived remote requirements from the control plane's `DEPLOY_MODE`. A bare control plane therefore required only Git and omitted Docker from the new-server setup results, even though remote application deployments and the managed Edge container depend on Docker. This allowed setup to appear ready and Edge installation to fail later.

## Related issue

Closes #767

## Changes

- Removed the control-plane deployment-mode dependency from remote server requirements.
- Always require Docker and Git on managed remote deployment targets.
- Extracted the requirement policy into a small pure module.
- Added regression coverage proving Docker remains required for a bare-mode control plane.
- Reused the existing Docker check, installer, UI row, and automatic installation order.

## Verification

```text
bun run test -- src/modules/system/server-check.requirements.test.ts
# 1 passed, 0 failed

bun run build
# API build passed

bun run test
# 5,155 passed; two environment-restricted tests failed inside the sandbox:
# - ssh-key-path fixture could not write beneath the user home directory
# - safe-fetch could not bind 127.0.0.1

bun run test -- test/lib/ssh-key-path.test.ts test/lib/safe-fetch.test.ts src/modules/system/server-check.requirements.test.ts
# Rerun with normal filesystem/socket access: 34 passed, 0 failed

bunx prettier --check src/modules/system/server-check.requirements.ts src/modules/system/server-check.requirements.test.ts
# passed
```

The existing controller file does not pass a whole-file Prettier check on upstream. It was not reformatted to avoid unrelated formatting changes.

## Screenshots

Not included. The existing Docker component row is now returned by the backend when missing; this change does not introduce a new UI component.

## Checklist

- [x] This PR contains one change (no grab-bag).
- [x] The diff is scoped and contains no unrelated formatting or refactors.
- [x] I added a regression test that fails before and passes after the fix.
- [x] Relevant tests, build, and formatting checks pass locally; sandbox-only failures were rerun successfully with the required access.
- [x] I understand every line in the diff.

### DIFF
--- a/apps/api/src/modules/system/server-check.controller.ts
+++ b/apps/api/src/modules/system/server-check.controller.ts
@@ -60,6 +60,7 @@ import {
   setupPromptState,
 } from "./setup-session";
 import type { PromptUserFn } from "@repo/adapters";
+import { resolveRequiredComponents } from "./server-check.requirements";
 
 // ─── Allowlisted components ──────────────────────────────────────────────────
 
@@ -97,16 +98,9 @@ async function withCapabilities<T extends { name: string; installed?: boolean }>
 }
 
 /**
- * Core components required for the current deployment mode.
+ * Core components required on every remote deployment server.
  * These are always shown in System Health regardless of install state.
  */
-function resolveRequiredComponents(): string[] {
-  const mode = env.DEPLOY_MODE;
-  if (mode === "docker") return ["docker", "git"];
-  if (mode === "bare") return ["git"];
-  return ["git"];
-}
-
 /**
  * Infrastructure components - optional but important for app deployment.
  * Shown in System Health only when detected (installed) on the server.
--- a/apps/api/src/modules/system/server-check.requirements.test.ts
+++ b/apps/api/src/modules/system/server-check.requirements.test.ts
@@ -1,0 +1,8 @@
+import { describe, expect, it } from "vitest";
+import { resolveRequiredComponents } from "./server-check.requirements";
+
+describe("remote server requirements", () => {
+  it("requires Docker even when the control plane runs in bare mode", () => {
+    expect(resolveRequiredComponents()).toEqual(["docker", "git"]);
+  });
+});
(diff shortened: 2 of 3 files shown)


Hydralerne:
Reviewed against main `d6fb0acb`: the shared adapter policy now requires Docker and Git for every managed remote server independently of the control plane runtime. Engine server checks and setup both use that policy; the shared dependency planner also installs Docker before Edge, including Edge-only installation requests.

All 9 current API dependency tests pass, and the prerequisite/setup tests passed in the complete adapter suite. #767 is already closed.

Closing as superseded: adding another API-local requirements helper would duplicate the current shared policy. Thank you for identifying the remote/control-plane mismatch and providing a focused fix.

PR #771, opened 2026-08-30
fix(apps): publish port-only install endpoints

### LINKED ISSUE #770: App installer port-only mode does not publish a host port
## Summary

Choosing **Port only (no domain)** while installing a catalog app does not persist a Docker host-port mapping. Services that depend on `{{publicUrl:<service>}}` then deploy with those environment variables omitted and the advertised `http://<server>:<port>` URL refuses connections.

## Reproduction

1. Install the bundled Supabase app on a server.
2. Choose **Port only (no domain)** for the `kong:8000` Studio & API endpoint.
3. Deploy.
4. Open `http://<server-ip>:8000`.

Observed warning:

```text
Domain routing needs attention

Some domains aren't routed yet — the app is deployed and running; fix DNS/routing and Retry from the Domains tab: Service "auth": no public URL is known for GOTRUE_SITE_URL={{publicUrl:kong}}, API_EXTERNAL_URL={{publicUrl:kong}} — those variables are left UNSET rather than blank. Assign a domain to the referenced service, or publish a fixed host port for it, then redeploy.; Service "studio": no public URL is known for SUPABASE_PUBLIC_URL={{publicUrl:kong}} — that variable is left UNSET rather than blank. Assign a domain to the referenced service, or publish a fixed host port for it, then redeploy.
```

On the deployment host, `openship-supabase-kong` is healthy but has `PortBindings={}` and Docker only reports `8000-8001/tcp`. Nothing listens on host port 8000, so the connection is refused. Supabase Auth repeatedly restarts because `API_EXTERNAL_URL` is absent.

## Root cause

`routeChoices()` sends `{ service, port, mode: "port" }`, but `planInstallRouting()` intentionally produces no hostname and the service creation path copies only the template's existing `ports`. Supabase's Kong service declares `exposedPort`/`routes` but no compose `ports`, so port-only mode publishes nothing. The deploy-time public URL map cannot resolve `{{publicUrl:kong}}`.

## Proposed fix

When an HTTP endpoint is explicitly selected as `mode: "port"`, persist a fixed, deliberately public Docker mapping (`0.0.0.0:<hostPort>:<containerPort>`) unless the template already declares a mapping for that container port. Keep domain modes unchanged.

Add a regression test proving Supabase `kong:8000` receives a public fixed mapping in port-only mode and does not receive one in domain mode.

## Scope

Shared catalog app installer behavior, not a Supabase-specific hard-coded workaround.

luckycrm:
## Summary

- persist fixed all-interface host bindings for catalog endpoints explicitly selected as port-only
- preserve catalog-authored host remaps and protocols while leaving domain-mode endpoints unpublished
- repair the binding when retrying an existing failed install draft
- add regression coverage for Supabase Kong and generic remapped ports

Fixes #770.

## Why

The app installer stored the routing choice but did not add a Docker host-port mapping. For Supabase, Kong therefore had no public URL, leaving `GOTRUE_SITE_URL`, `API_EXTERNAL_URL`, and `SUPABASE_PUBLIC_URL` unset and causing the deployment warning.

## Verification

- `bunx vitest run test/modules/apps/app-install-routing.test.ts` (15 passed)
- `bun run lint` in `apps/api`
- full API test run: 5,157 passed, 3 skipped; the two sandbox-blocked test files were rerun with host permissions (33 passed)

### DIFF
--- a/apps/api/src/modules/apps/app-install.service.ts
+++ b/apps/api/src/modules/apps/app-install.service.ts
@@ -84,47 +84,49 @@ function signHs256Jwt(secret: string, role: string): string {
  */
 export async function getAppCatalog(ctx: RequestContext) {
   const custom = await listOrgCustomApps(ctx.organizationId);
-  return [...getRuntimeCatalog(), ...custom].filter((t) => !t.unlisted).map((t) => ({
-    id: t.id,
-    name: t.name,
-    description: t.description,
-    kind: t.kind,
-    logo: t.logo,
-    category: t.category,
-    tags: t.tags ?? [],
-    flowHref: t.flowHref,
-    // How the installed app is managed (schema settings / custom href / none).
-    management: getAppManagement(t),
-    // Verified trust mark (official open-source image + reviewed pipeline).
-    verified: !!t.verified,
-    // Hosting model for the catalog badge + wizard notice (self-hosted default).
-    hosting: t.hosting ?? "self-hosted",
-    // What the app needs from the machine — the wizard shows it against the
-    // chosen destination's real capacity, and deploy preflight enforces it.
-    minResources: t.minResources,
-    // A per-org user-uploaded app — always unverified; dashboard shows the warning.
-    custom: !!t.custom,
-    // Not installable this version → dashboard dims it + blocks the click.
-    comingSoon: !t.available,
-    // Needs a newer Openship than this instance → dashboard shows a guided
-    // "Requires Openship ≥ X" state; install is refused server-side too.
-    requiresUpdate: t.requiresUpdate,
-    updateAvailable: t.updateAvailable,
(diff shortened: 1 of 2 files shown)


Hydralerne:
Reviewed against main `d6fb0acb`: this is already fixed by `76ff1cfc` and retained in the shared engine app installer. The existing `installServicePorts` implementation publishes fixed public bindings for port-only choices, preserves authored remaps/protocols, and reverses its generated binding when retrying with domain routing.

All 20 current install-routing tests pass, including the real Supabase Kong template and its public URL resolution, failed-draft reconciliation, and idempotence. #770 is already closed.

Closing this as superseded rather than adding a second implementation or losing the newer reversal behavior. Thank you for the scoped fix and regression cases.

PR #786, opened 2026-09-02
fix(mail): write the relay maps as root on a non-root SSH login (#756)

### LINKED ISSUE #756: Enable outbound relay fails with "Permission denied" when the server's SSH user is not root
# Enable outbound relay fails with "Permission denied" when the server's SSH user is not root

**Version:** 0.6.8 (`ghcr.io/oblien/openship-api:0.6.8`, `openship-mail:0.6.8`)
**Host:** Ubuntu 25.x (Oracle Cloud ARM64, kernel 6.17.0-1020-oracle), Docker, mail engine flavor = `container`
**Server record:** `is_local = true`, `ssh_user = ubuntu` (the distro's default sudo user, no root SSH login)

## Summary

`POST /api/mail/admin/:serverId/relay` ("Enable relay" in Mail, Sending) always returns **500 "Permission denied"** on a host whose registered SSH user is not `root`.

The relay config is written to the host with **SFTP** (`CommandExecutor.writeFile`), which has no privilege escalation, while the bind-mount directory it targets is `root:root 0755`. Everything else in the same code path works, because every other step goes through `exec.exec("docker exec openship-mail ...")`, which runs as root *inside* the container. So the operation is split across two privilege contexts and only the SFTP half fails.

This is not an SES or config problem; it fails before any SMTP connection is attempted.

## Error

API log:

```
<-- POST /api/mail/admin/b426bbaa-.../relay
[MAIL ADMIN ERROR] POST /api/mail/admin/b426bbaa-.../relay
error: Permission denied
 code: 3,
      at 101 (/app/node_modules/ssh2/lib/protocol/SFTP.js:2854:23)
      at push (/app/node_modules/ssh2/lib/protocol/SFTP.js:278:11)
      at CHANNEL_DATA (/app/node_modules/ssh2/lib/client.js:585:23)
      ...
--> POST /api/mail/admin/b426bbaa-.../relay 500 37ms
```

SFTP status code 3 = `SSH_FX_PERMISSION_DENIED`. The UI surfaces it as **Outbound sending, Permission denied** with no remediation.

## Root cause

`apps/api/src/modules/mail/admin/outbound-relay.service.ts`:

```ts
await exec.writeFile(saslMap.write, `${nexthop} ${input.username}:${input.password}\n`);
await exec.exec(`chmod 600 ${saslMap.write}`);
await exec.exec(engine(`postmap ${saslMap.engine}`));
```

- `saslMap.write` comes from `mailConfigFile(flavor, "saslPasswd")`; for `flavor === "container"` that resolves to `MAIL_HOST_PATHS.saslPasswd` = `/var/lib/openship/mail/config/postfix/sasl_passwd` (`packages/adapters/src/infra/mail-container.ts`).
- `SshExecutor.writeFile` (`packages/adapters/src/system/ssh-executor.ts:399`) does `mkdir -p` best-effort, then `sftp.writeFile(...)`. There is **no `sudo` anywhere in that executor**; the whole file has zero occurrences.
- Non-root shell commands still succeed for the container flavor only because the SSH user is in the `docker` group, so `docker exec openship-mail ...` gives root *inside* the container. The SFTP write has no such escape hatch.

On this host:

```
drwxr-xr-x 7 root root /var/lib/openship/mail/config/postfix
drwxr-xr-x 2 root root /var/lib/openship/mail/config/amavis
drwxr-xr-x 4 root root /var/lib/openship/mail/config/dovecot
drwxr-xr-x 2 systemd-timesync systemd-timesync /var/lib/openship/mail/dkim

$ id
uid=1001(ubuntu) ... groups=...,988(docker)

$ touch /var/lib/openship/mail/config/postfix/.probe
touch: cannot touch '...': Permission denied
```

The same failure should hit every `writeFile` to `MAIL_HOST_PATHS` under a non-root SSH user, not just the relay: `saslPasswd`, `senderRelayhost`, `relayTlsPolicy` (selected-scope TLS policy) and `amavisUserConf` (the DKIM `50-user` write). The relay is just the first one a user reaches from the UI.

Note the `mkdir -p` before the SFTP write is wrapped in `try/catch` "best effort", so a permission problem there is swallowed and only resurfaces as the opaque SFTP error.

## Reproduce

1. Install Openship on a host where the default login is a non-root sudo user (Ubuntu `ubuntu`, Debian `admin`, AL2023 `ec2-user`, etc.) and register that host as the local server, so `ssh_user = 'ubuntu'`, `is_local = true`.
2. Run mail setup; the container engine comes up fine (Postfix/Dovecot/amavis all RUNNING).
3. Mail, Sending, Enable relay, with valid SES SMTP credentials (`email-smtp.<region>.amazonaws.com:587`).
4. Result: 500, "Permission denied". `postconf -h relayhost` inside `openship-mail` stays empty and `/etc/postfix/sasl_passwd` is never created.

## Expected

Either the write succeeds regardless of the SSH user's uid, or the API returns a typed error naming the actual problem ("cannot write `<path>` as `<user>`; the mail state directory is root-owned") the way `requireMailEngine` does for a stopped engine, instead of a raw ssh2 SFTP string.

## Suggested fixes

The security invariant in the code comment ("SFTP, no shell; the SASL password must never reach a command line") is worth keeping, so plain `sudo tee` on an argv is not the answer. Options that preserve it:

1. **Write through the engine over stdin.** `docker exec -i openship-mail tee /etc/postfix/sasl_passwd` with the secret piped on **stdin**. Root inside the container, secret never in argv or in the process table, and it targets the container path directly so the host-side ownership stops mattering. This matches how every other step in the function already reaches Postfix.
2. **Make the mail state dirs writable by the SSH user at setup time.** `chown` or ACL `${MAIL_HOST_STATE_DIR}/config/*` to the connecting user when the mail engine is provisioned, so `MAIL_HOST_PATHS` writes are consistent with the executor's actual privileges.
3. **Give `SshExecutor` a privileged write path** (`sudo -n install -m 600 /dev/stdin <path>` fed over stdin) and use it for all `MAIL_HOST_PATHS` targets.

Whichever route you pick, a preflight check ("mail state dir not writable by `<ssh_user>`") plus a real remediation message in the UI would turn this from a dead end into a one-click fix.

## Workaround

Grant the SSH user write access to the mail config dirs on the host:

```sh
sudo setfacl -m u:$USER:rwx /var/lib/openship/mail/config/postfix \
                            /var/lib/openship/mail/config/amavis
```

then retry Enable relay. (Or register the server with `ssh_user = root`.)

## Related issues

Searched open and closed issues; this does not appear to be reported.

- **#563** (closed, fixed) "Docker mail deployment runs doveadm and chown vmail on the host, where neither exists". Same *family* of defect: an operation aimed at the wrong execution context on the container flavor. That fix routed shell commands through the engine (`runMailCommand`), which is why every `exec.exec` call in the relay path works today. The SFTP `writeFile` half was not covered, and it fails for a different reason: the SSH user's uid, not missing host tooling. The fix suggested above (write through the engine over stdin) is the same remedy #563 applied to its commands.

AbdullahM07:
Fixes #756.

## Problem

`POST /api/mail/admin/:serverId/relay` always returned **500 "Permission denied"** on a host registered with a non-root SSH login (Ubuntu's `ubuntu`, Debian's `admin`, AL2023's `ec2-user` — most cloud images).

The relay code writes the SASL password map with `exec.writeFile` — SFTP, as the login user — into `/var/lib/openship/mail/config/postfix`, a directory Docker materialises as `root:root 0755`. Every other step in the same function reaches Postfix as root through `docker exec` (the login is in the `docker` group), so the SFTP half was the only step that could fail, and it did — as a raw ssh2 `SSH_FX_PERMISSION_DENIED` with no remediation. `packages/adapters/src/system/ssh-executor.ts` has no privilege escalation at all (zero occurrences of `sudo`).

The same defect sat on the TLS policy map write, on the `chmod 600` / `rm -f` of both maps, and on the DKIM `50-user` amavis conf write in `provisionDomainDkim`.

## Fix

The security invariant in the file header — the SASL password must never reach a command line — is kept. The repo already has the right mechanism: `elevatedExecutor.writeFile` stages the content **over SFTP** into a private `0700` dir and publishes it with `sudo mv`, so the secret still never crosses argv or the process table.

- New `mailConfigWriter(exec)` in `mail-engine.ts` (next to `mailConfigFile`, whose `write` paths it exists for) asks the existing privilege gate (`privilegedExecutor`, `onRefusedHost: "proceed"`):
  - root login → the caller's own executor, byte-for-byte the previous behaviour;
  - non-root + passwordless sudo → `elevatedExecutor`;
  - supported host with neither → typed **409 `MAIL_HOST_NOT_WRITABLE`** (`MailHostNotWritableError`) naming the fix, thrown **before** anything touches the box — same shape as the `requireMailEngine` gate;
  - unmeasurable host (banner / forced command) → proceeds as before, so a chatty sshd can't turn a working root login into a refusal.
- `relayTransport` now carries `files`, the executor for the **host end** of the maps: the SASL and TLS-policy writes, their `chmod 600`, and their `rm -f` go through it. Engine commands (`postmap`, `postconf`, `postfix reload`) deliberately keep `exec` — the `docker` group already makes them root where it counts, and elevating them would change their transport for nothing.
- `provisionDomainDkim` writes the amavis conf through the same gate.

## Testing

`outbound-relay.service.test.ts` now answers the host-privilege probe (`login: root | sudo | none`), so the real `privilegedExecutor` gate decides the executor rather than a stub. New tests, for both flavors where relevant:

- **sudo**: the secret lands via SFTP in `/tmp/.openship-elev-*/payload`, is published by `sudo … chown 0:0 … mv -f` into the real path, `chmod 600` runs elevated, and no shell command contains the password or username;
- **sudo**: only the host-side file half is elevated — Postfix commands keep their transport;
- **sudo**: disabling removes the root-owned map as root;
- **none**: typed 409 with `code: MAIL_HOST_NOT_WRITABLE`, zero writes and zero commands issued;
- **root**: the caller's own executor, exactly as before.

All 43 pre-existing relay tests pass unchanged on the `root` default. `test/modules/mail` + `src/modules/mail`: 388/388; `host-command-ownership` rules: pass; `tsc --noEmit` clean.

## Follow-up: the map is 0600 from the moment it exists

A security pass on the first commit found a window it introduced: the elevated write staged the map at the login user's umask and the mode-preserving `sudo mv` published it **0644** into the root-owned `0755` directory, with the `chmod 600` landing in a *separate* SSH round trip — so on every save the credentials were briefly readable by any local account. (The root-login path had the same window on a first-ever write, since SFTP creates at 0644 too.) The repo's own precedent for secret-bearing elevated writes — `nginx.ts _writeFile` — tightens the staged copy *before* the `mv` for exactly this reason.

The second commit closes it at the transport layer rather than with another chmod:

- `CommandExecutor.writeFile` gains an optional `{ mode }`. `elevatedExecutor` applies it to the staged copy **inside the same `sudo` as the publish** (`mkdir -p … && chmod 600 <staged> && chown 0:0 <staged> && mv -f …`), so `mv` carries it and the published inode never exists more open. `SshExecutor` and the local executor put it on the open, so even a first-ever write is never 0644. Executors that can't honour it ignore it — documented on the interface — which is why the relay keeps its `chmod 600` as a backstop (a no-op on the transports that matter, and still the fix for a map an older version left 0644).
- The relay passes `{ mode: 0o600 }` for the SASL map only; the TLS policy map and the amavis conf carry nothing sensitive.

Tests: the elevated-executor suite pins the exact publish command with the `chmod` ahead of the `mv`; the relay's sudo test asserts the tightening precedes the `mv` inside the publish command and names the staged path, and the root test asserts the mode rides the write. Adapters `src/system` + `src/infra`: 1513/1513; API mail suites: 386/386; `tsc` clean in adapters, api and cli.


AbdullahM07:
Superseded by the 0.6.10 patch (76ff1cfc), which landed its own fix for #756 in the same files: sudo elevation for host-side config writes, a typed 409 when there is no route to root, the file mode applied before any secret bytes land in every executor, plus the ACL-granted bind-mount case. After merging main, this branch is identical to it, so there is nothing left to merge. Closing.

Issue #779, opened 2026-09-01
Enhance self-hosted image retention with age-based cleanup and runtime image visibility


fuzhyperblue:
## Summary

OpenShip already includes a project-scoped built-image garbage collector and rollback-aware retention logic. However, self-hosted operators currently have limited visibility into whether image GC has run, what it protected, what it removed, and why older images remain on disk.

I would like to request better observability and operator controls for image retention and cleanup on self-hosted instances.

## Environment

- OpenShip version: `0.6.9`
- Installation: self-hosted Docker Compose
- Host OS: Ubuntu `22.04.5`
- Docker storage: `/var/lib/docker`
- Root filesystem: approximately `57 GB`

## Observed disk usage

On a self-hosted OpenShip instance, the root filesystem reached approximately 83% usage:

```text
Filesystem                         Size  Used  Avail  Use%
/dev/mapper/ubuntu--vg-ubuntu--lv   57G   45G   9.3G   83%
```

Docker reported:

```text
Images:
  Total:       35
  Active:      11
  Size:       21.74GB
  Reclaimable: 12.28GB (56%)

Containers:
  Total:       11
  Active:      11
  Reclaimable: 0B

Local Volumes:
  Total:        8
  Active:       6
  Size:       2.18GB
  Reclaimable: 0B

Build Cache:
  0B
```

The largest host-level consumers were:

```text
/var/lib/docker: approximately 27 GB
/var/lib/docker/overlay2: approximately 25 GB
/var/log/journal: approximately 4 GB
```

The journal is a separate host-level concern and is not expected to be handled by the OpenShip image garbage collector.

## Current OpenShip cleanup behavior

The current codebase includes:

```text
apps/api/src/modules/deployments/image-gc.ts
```

The image GC implementation appears to:

- identify OpenShip-owned build images using OpenShip labels;
- preserve the active deployment image;
- preserve pinned deployment images;
- preserve images within the configured rollback window;
- remove superseded OpenShip-owned images outside that keep-set;
- avoid removing third-party images such as PostgreSQL and Redis;
- skip images that are still in use;
- run as the built-in `images:gc` system job;
- run after successful deployments and as a scheduled daily sweep.

The job registry defines:

```text
Job: images:gc
Label: Built-image garbage collection
Default schedule: 23 4 * * *
```

## Current instance state

On the inspected self-hosted instance:

```text
images:gc
enabled: true
schedule: 23 4 * * *
```

[...]

PR #794, opened 2026-09-02
feat(system): make the built-image GC inspectable and runnable by hand (#779)

AbdullahM07:
## Summary

Makes the built-image garbage collector (`images:gc`) inspectable and runnable by hand: a read-only dry run that lists every project's build images with the decision a sweep would take **and why**, a manual run recorded in job history, an optional age filter that can only make a run more conservative, and a host-local command for the previous `openship-api` / `-dashboard` / `-edge` images an update leaves behind.

## Motivation

#779: on a self-hosted box at 83% disk, the operator could see that `images:gc` was enabled and scheduled but not whether it had run, what it removed, or why older `openship/<slug>:bld_…` images and three generations of runtime images were still on the host. The only visible lever was `docker system prune -a`, which is exactly what a rollback-capable platform must not push people towards. The GC's decisions were already sound; they just weren't visible.

## Related issue

#779. This is a proposal for the issue's "useful first version" list; scope and approach are open to whatever the maintainers agree on there. Deliberately **not** in this PR: a dashboard view, and any change to what the scheduled sweep removes by default.

## Changes

**apps/api**
- `deployments/image-gc.ts`: `computeKeepSetDetail` keeps *why* each ref is protected (`active` / `pinned` / `rollback-window`, strongest wins; `computeKeepSet` is now derived from it). `classifyImage` is the single classifier — adds `in-use`, `foreign-tag`, `min-age` and the two removal reasons (`superseded`, `dangling`) — and `reapProjectImages` now runs on it, so a plan says exactly what a sweep does. `planImageGc` / `planProjectImages` are the read-only dry run (also lists containers on the host to name in-use images up front instead of leaving it to the daemon's 409). `parseMinAge` (`30d`, `12h`, `2w`, `90m`) rejects zero/junk. An age-limited run skips the dangling prune, which has no age filter of its own. `imageGcJobSummary` is the one summary shape, shared with the registry.
- `system/image-gc.controller.ts` + `system.routes.ts`: `GET /api/system/image-gc/plan?olderThan=30d` and `POST /api/system/image-gc/run`. `job:read` / `job:write` tags **plus `requireInstanceAdmin()`** — both walk every org's projects and hosts, same reasoning as `/edge/untracked`. A manual run goes through `recordJobRun(..., { trigger: "manual" })` so it shows in job history next to the scheduled ticks.
- `jobs/job.registry.ts`: uses `imageGcJobSummary`.

**packages/adapters**
- `runtime/docker.ts`: `listProjectImages` reports `createdAt` (Docker's seconds → ms, `null` when unreported).

**apps/cli**
- `openship system image-gc [--dry-run] [--older-than <age>] [--yes]`: status table (enabled, schedule, last run + its summary, age filter), per-image table (project, image, size, age, rollback window, decision, reason), totals; without `--dry-run` it confirms and runs. `--json` supported.
- `openship system runtime-images [--prune] [--keep <n>] [--yes]`: host-local (no API), lists the control-plane images; `--prune` removes versions beyond the one a container references plus `--keep` previous. Pure selection in `lib/runtime-images.ts`: anything referenced by a container (running **or stopped**) is kept, non-version tags (`latest`) are never candidates, removal is plain `docker rmi` with no `--force`.

**docs / changelog**: `cli/self-host.mdx`, `api/system.mdx`, `CHANGELOG.md` (Unreleased).

## Safety

Nothing changes in what the scheduled sweep removes by default. The label-scoped selection, the keep-set, untag-only-our-tags, and the in-use 409 backstop are untouched — `selectImageRemovalRefs` and its tests still pass unchanged. Age can only narrow a run (an image of unknown age counts as young). The plan never writes: no prune, no capacity re-measure.

## Verification

```bash
# adapters / api / cli typecheck
$ npx tsc --noEmit            # in each of packages/adapters, apps/api, apps/cli → exit 0

# api: deployments + projects + jobs suites (includes the new image-gc tests)
$ npx vitest run src/modules/deployments src/modules/projects src/modules/jobs
 ✓ src/modules/deployments/image-gc.test.ts (15 tests)
 Test Files  54 passed (54)   Tests  560 passed (560)

# adapters runtime suite
$ npx vitest run src/runtime
 Test Files  42 passed (42)   Tests  610 passed (610)

# cli
$ npx vitest run test/unit/runtime-images.test.ts
 ✓ test/unit/runtime-images.test.ts (3 tests)
$ npx vitest run
 Test Files  2 failed | 37 passed   Tests  4 failed | 494 passed
```

The 4 CLI failures (`test/e2e/completion.test.ts` ×3 — a Node `DEP0205` deprecation warning on stderr — and a 5 s timeout in `test/unit/ports.test.ts`) fail identically on a clean `upstream/main` checkout in this environment; they're unrelated to this change.

New tests that fail without the change: keep-reason priority (`active` outranks `pinned` for a shared image), `classifyImage` naming `in-use` only when containers are listed, own-tags-only / foreign-tag / dangling, the age filter keeping young and age-unknown images while still removing old ones and never overriding the keep-set, `parseMinAge` rejecting `0d` and junk; runtime-image prune keeping referenced + N previous per repository and never ordering `latest`/`main`.

Not exercised against a live Docker host from this machine; the daemon-facing surface is one added field (`createdAt`) plus the existing `listAllContainers()`.

## Checklist

- [x] One change per PR — one agreed feature, with nothing unrelated bundled in
- [x] `bun run test` (relevant workspaces), `bun run --cwd <workspace> lint`, and `bun format` (on the touched files) pass locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,22 @@ the in-app updater surfaces critical advisories from `release-advisories.json`.
 
 ## Unreleased
 
+### Maintenance
+
+- **The built-image garbage collector is inspectable and runnable by hand** —
+  `openship system image-gc --dry-run` lists every project's build images on its
+  deploy host with the decision the `images:gc` sweep would take and why (active,
+  pinned, inside the rollback window, in use, operator-retagged, or superseded),
+  the rollback window that decided it, the job's schedule and last run, and the
+  space a run would reclaim. Without `--dry-run` it confirms and runs the sweep,
+  recorded as a manual `images:gc` job run. `--older-than 30d` narrows a run to
+  superseded images at least that old; age never overrides the keep-set. Backed
+  by `GET /api/system/image-gc/plan` and `POST /api/system/image-gc/run`
+  (instance admin). `openship system runtime-images` lists the previous
+  `openship-api` / `openship-dashboard` / `openship-edge` images an update leaves
+  on the host and, with `--prune`, removes all but the referenced version and
+  `--keep` previous ones (#779).
+
 ## 0.6.10
 
 This release hardens the full deployment lifecycle, makes Compose reconciliation
--- a/apps/api/src/modules/deployments/image-gc.test.ts
+++ b/apps/api/src/modules/deployments/image-gc.test.ts
@@ -1,6 +1,13 @@
 import { describe, it, expect } from "vitest";
 import type { Deployment } from "@repo/db";
-import { computeKeepSet, selectImageRemovalRefs } from "./image-gc";
+import {
+  classifyImage,
+  computeKeepSet,
+  computeKeepSetDetail,
+  parseMinAge,
+  selectImageRemovalRefs,
+  type KeepReason,
+} from "./image-gc";
 
 // Minimal deployment shape for the pure keep-set logic (loaders are injected, so
 // no DB). Casts keep the fixtures terse — computeKeepSet only reads id/imageRef/pinned.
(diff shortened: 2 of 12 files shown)


Hydralerne:
Closing this PR as outside the current integration scope, which the maintainer has limited to bugs and general improvements. This is new feature work. Feature request #779 remains open for a separate feature effort.

Deferred: new image-GC inspection, dry-run and retention controls; PR #794 is not included.

PR #796, opened 2026-09-03
fix(docker): stop build immediately on memory starvation and add actionable OOM diagnostics

lagmandu:
### Summary

When a project build (e.g. Next.js/Turbopack, Vite, Rust) exceeds its allocated RAM limit inside Docker:
1. The container hits 95–99% memory usage and enters an intense Linux kernel swap-thrashing loop (in production benchmarks, this produced >270 GB of block I/O).
2. The compiler produces zero stdout/stderr output while thrashing.
3. OpenShip previously waited for a hardcoded **30 minutes** (`DOCKER_BUILD_IDLE_TIMEOUT_MS = 30 * 60 * 1000`) before timing out, printing generic `Still building... (no output for Xm)` messages.
4. On timeout, the error was misleading: *"Docker build produced no output for 30 minutes. This usually means the remote server cannot reach the package registry, has broken DNS, or the Docker daemon stalled during the build."*
5. When processes were killed by the Linux kernel OOM killer (`exit code 137`), OpenShip outputted raw `returned a non-zero code: 137` without explaining that code 137 is Out Of Memory.

### Changes

1. **Active Memory Starvation Watchdog**:
   - In `streamDockerodeBuild`, a watchdog checks intermediate build container stats during silent periods.
   - If an active build container is pinned at >= 95% memory usage with zero log output for >= 30 seconds, OpenShip **immediately kills the container and halts the deployment**, preventing disk thrashing and providing clear diagnostics:
     > `Docker build terminated: container is memory-starved (XMB / YMB, 99.5% limit reached). The compiler or build process ran out of RAM and stalled in swap thrashing. Increase the project's Build Memory limit in OpenShip Settings -> Resources.`
2. **Actionable Exit Code 137 (OOM) Diagnostics**:
   - In `extractBuildFailureHint`, translates exit code `137` into an explicit Out-Of-Memory explanation with immediate resolution steps.
   - Detects `JavaScript heap out of memory`, `Fatal process out of memory`, and npm `ENOMEM`.
   - In `buildImageOnRemote`, decodes exit code 137 across remote SSH builds.
3. **Configurable & Reasonable Idle Timeout**:
   - Lowered default idle timeout from 30 minutes to **10 minutes** (`OPENSHIP_BUILD_IDLE_TIMEOUT_MS`).
   - Included configured RAM limit in the timeout error message.

### DIFF
--- a/packages/adapters/src/runtime/docker-build-oom.test.ts
+++ b/packages/adapters/src/runtime/docker-build-oom.test.ts
@@ -1,0 +1,196 @@
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import { Readable } from "node:stream";
+import { DockerRuntime, getDockerBuildIdleTimeoutMs } from "./docker";
+import { BuildLogger } from "./build-pipeline";
+
+describe("Docker build OOM diagnostics and starvation detection", () => {
+  const originalEnv = process.env.OPENSHIP_BUILD_IDLE_TIMEOUT_MS;
+
+  afterEach(() => {
+    if (originalEnv !== undefined) {
+      process.env.OPENSHIP_BUILD_IDLE_TIMEOUT_MS = originalEnv;
+    } else {
+      delete process.env.OPENSHIP_BUILD_IDLE_TIMEOUT_MS;
+    }
+  });
+
+  describe("getDockerBuildIdleTimeoutMs", () => {
+    it("defaults to 10 minutes (600_000 ms)", () => {
+      delete process.env.OPENSHIP_BUILD_IDLE_TIMEOUT_MS;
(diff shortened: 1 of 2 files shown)


Hydralerne:
Superseded by #827, now merged into main as 1292fbe6.

Thank you @lagmandu for identifying the stalled/OOM-thrashing build behavior and contributing the original diagnostics. We kept the bounded configurable idle timeout and actionable OOM/SIGKILL reporting, but replaced the host-wide recent-container watchdog: recency plus memory usage cannot prove build ownership and could terminate an unrelated workload.

The merged implementation only cleans up the exact classic-builder RUN container after matching the build stream ID, a random per-build transient marker, parent image, command, and memory configuration. It covers Docker Engine and native SSH build paths, and all CI checks passed.

PR #803, opened 2026-09-03
fix(dashboard): preserve secret env values on config save

### LINKED ISSUE #800: Dashboard "Save" on project env vars reports success but silently drops secret values (manual single-field edit, not bulk import)
## Environment
- Openship version: 0.6.9 (self-hosted, `openship up`, docker compose)
- Services: api, dashboard, edge, postgres, redis — all `0.6.9`, all healthy
- Project: single Next.js app, one env var environment in use (`production`)

## Summary
Manually editing a single secret-typed env var in the project's Environment Variables panel and clicking Save shows a "Saved / Configuration saved" success toast, and the project's dashboard card shows "updated Xm ago", but the value is **not persisted**.  Reloading the panel shows the field back to its empty placeholder state, and a build immediately after fails with the app's own "missing required env var" error, confirming the value never made it into storage.

This is **not** the same issue as #587. That issue is about the bulk-paste import wizard defaulting everything to `isSecret: true`, and its own
conclusion was "the underlying values are correctly persisted, it's a display-only problem."  This report is the opposite: I confirmed directly
against the database that the value itself is lost on save, for a single manually-typed secret field, no import wizard involved.

## Repro steps
1. Open a project > Environment Variables (production environment).
2. Edit an existing secret-typed variable (e.g. a DB password field, already marked Secret from a previous set) and type a new value directly into the input.
3. Click Save. Toast confirms "Configuration saved."
4. Navigate away and back to the Environment Variables panel.
5. The field shows its empty/unset placeholder state again, indistinguishable from a field that was never set.
6. Trigger a build. The build (_or app start, depending on where the var is consumed_) fails with a "missing env var" error, proving the value was not actually stored, not just hidden from the UI.

## Evidence (checked directly against the self-hosted Postgres instance)
All env var values are encrypted at rest, but ciphertext length still scales with plaintext length, so it's usable as an is-this-empty signal without decrypting anything:

- Before the edit: the target secret var's stored ciphertext was the **minimum possible length**, the same length produced by encrypting a genuinely empty string (_verified against other, non-secret single/short digit plain vars in the same project, which land at the same minimum length_).
- After typing a real ~17-character value in the dashboard and getting the "Saved" toast: **the stored ciphertext length was unchanged** still the empty-string minimum, timestamped (`updated_at`) to the same second as the save action.  So a write did happen at that moment (the row's `updated_at` moved), but whatever got written was empty, not the typed value.
- API access logs for that exact time window show no `PUT`/`PATCH` request to any env-var-shaped route at all.  The only `POST`s in that window are to the project's "ensure" endpoint and the deployment/build-trigger endpoint, suggesting the dashboard's save bundles the env var write into one of those calls rather than a dedicated env-write endpoint, and something in that path drops secret field values (_plausibly: the secret input is never seeded with the real value on load, since secrets can't be read back correct, but the save payload then appears to send the empty display value for that field instead of "leave unchanged," clobbering it_).

## Isolation test: CLI vs dashboard for the identical write
To rule out a backend/encryption-layer bug, I set the same variable via the CLI instead of the dashboard, same project, same key:

```
openship project env set <project-id> --environment production --secret \
  --set SOME_SECRET_KEY=<value>
```

Output: `Updated env (production): 1 upserted, 0 deleted`

Re-checked the database immediately after: the stored ciphertext length for that key **jumped from the empty-string minimum to a length matching the real value's length**, timestamped to the CLI call.

So the API/encryption/persistence layer works correctly — this isolates the bug specifically to the **dashboard's save flow** for manually-edited secret fields, not the backend.

## Expected
Typing a new value into a secret field and clicking Save persists that value, the same way the CLI's `project env set --secret` does.

## Actual
The dashboard reports success but the secret field's value is silently dropped, no error shown to the user, build/deploy later fails downstream with no indication the cause was a lost env var save.

## Impact
This blocks anyone using the dashboard-only workflow (_i.e. not using the CLI_) from ever setting secret env vars by hand-editing an existing field, the save appears to succeed every time, so there's no obvious signal something went wrong short of the next build failing and someone tracing it back to the env var store.

## Suggested angle for a fix
Given the CLI path works, the likely fix is in how the dashboard's save handler builds its request payload for secret-typed fields specifically, it should either omit fields the user didn't touch (_so the backend keeps the existing encrypted value_) rather than resending the display value (which is always blank for secrets), or the save endpoint itself needs to distinguish "field submitted blank because user cleared it" from "field submitted blank because that's just what unreadable secrets render as."

AbdullahM07:
## Summary

- keep persisted secret values masked and track their server-side identity in the configuration editor
- save existing-project env changes through the partial merge API before reporting success or deploying
- omit existing-project env values from build requests while preserving the initial-deploy flow for new projects
- render unrevealable masked values as placeholders so typing replaces the secret instead of appending to the mask

## Tests

- `bun --cwd apps/dashboard test` (110 files, 1138 tests)
- `bun run --cwd apps/dashboard lint`

Fixes #800

### DIFF
--- a/apps/dashboard/src/components/import-project/EnvironmentVariables.tsx
+++ b/apps/dashboard/src/components/import-project/EnvironmentVariables.tsx
@@ -818,7 +818,9 @@ const EnvironmentVariables: React.FC<EnvironmentVariablesPropsOptional> = ({
           const displayValue =
             masked && Object.hasOwn(revealedValues, env.key)
               ? revealedValues[env.key]
-              : env.value;
+              : masked && !onReveal
+                ? ""
+                : env.value;
           const canToggleValue = !masked || Boolean(onReveal);
           return (
             <div key={index} data-env-index={index} className="space-y-1.5">
@@ -844,7 +846,7 @@ const EnvironmentVariables: React.FC<EnvironmentVariablesPropsOptional> = ({
                     type={showAsText ? "text" : "password"}
                     value={displayValue}
                     onChange={(e) => handleValueChange(index, e.target.value)}
-                    placeholder={ev.valuePlaceholder}
+                    placeholder={masked && !onReveal ? ENV_MASK : ev.valuePlaceholder}
                     readOnly={!isEditingMode}
                     className={`w-full px-3.5 py-2.5 pe-9 border border-border/50 rounded-lg text-sm font-mono text-foreground placeholder:text-muted-foreground/40 focus:outline-none focus:ring-2 focus:ring-primary/20 transition-all ${
                       !isEditingMode ? 'cursor-default bg-muted/20' : 'bg-muted/30'
--- a/apps/dashboard/src/components/import-project/environment-variables.render.test.tsx
+++ b/apps/dashboard/src/components/import-project/environment-variables.render.test.tsx
@@ -84,7 +84,12 @@ describe("EnvironmentVariables reveal affordance", () => {
 
   it("omits it when nothing can resolve the sentinel", () => {
     // Deliberate: the toggle could only ever display the dots themselves as text.
-    expect(eyeCount(render({ envVars: MASKED }))).toBe(0);
+    const html = render({ envVars: MASKED });
+    expect(eyeCount(html)).toBe(0);
+    // Keep the sentinel in state for safe diffing, but render it as a placeholder
+    // so typing starts a replacement value instead of appending to the dots.
+    expect(html).toContain('value=""');
+    expect(html).toContain('placeholder="••••••••"');
   });
 
   it("gives a plaintext row an eye with or without a source", () => {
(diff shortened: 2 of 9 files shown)


Hydralerne:
Reviewed this against current main (`d6fb0acb`). The bug reported in #800 has already been fixed by `062bb6d8` and the subsequent environment hardening on main.

The current wizard uses the shared `createProjectEnvEditState` / `computeProjectEnvDiff` path, saves explicit secret edits before reporting success, and omits unchanged secrets from both the partial merge and redeploy payload. It also preserves source-owned values and rejects saves when the saved environment failed to load. I verified the current behavior with all 48 environment diff, persistence, wiring, and rendering cases passing.

Closing this as superseded to avoid adding a second environment persistence planner. Thank you for the contribution and regression coverage; the original report remains fixed on main.

PR #814, opened 2026-09-04
fix(email): load attachment bytes only for downloads and forwards

cesarfrancots:
## Summary

- return base64 attachment bytes for explicit download, draft, and forward reads
- keep normal thread reads metadata-only so opening a message does not transfer every attachment
- preserve the original body and attachments when forwarding
- retain user-added forward attachments and avoid repeated `Fwd:` prefixes
- add regression coverage for the attachment-byte gate

## Context

This builds on the diagnosis and implementation in #355 by @DiogoDuart3. It incorporates the maintainer feedback there: attachment bytes are gated behind `includeAttachmentBytes` instead of being included in every `getThread` response.

The current `main` branch and v0.7.0 still set every parsed attachment's `body` to an empty string, so downloads produce an empty file. The download, draft, and forward paths now opt into bytes explicitly; the ordinary reading-pane path keeps the documented empty-body behavior.

## Verification

- `bun run --cwd apps/email/server test` — 56 passed
- `bun run --cwd apps/email/server build`
- `bun run --cwd apps/email/client build`
- `git diff --check`

### DIFF
--- a/apps/email/client/components/mail/mail-display.tsx
+++ b/apps/email/client/components/mail/mail-display.tsx
@@ -252,33 +252,14 @@ const cleanNameDisplay = (name?: string) => {
   return name.trim();
 };
 
-const ThreadAttachments = ({ attachments }: { attachments: Attachment[] }) => {
+const ThreadAttachments = ({
+  attachments,
+  onDownload,
+}: {
+  attachments: Attachment[];
+  onDownload: (attachment: Attachment) => void;
+}) => {
   if (!attachments || attachments.length === 0) return null;
-
-  const handleDownload = async (attachment: Attachment) => {
-    try {
-      // Convert base64 to blob
-      const byteCharacters = atob(attachment.body);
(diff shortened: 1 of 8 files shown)


Hydralerne:
Maintainer follow-up pushed in commit 66242833.\n\n- Attachment bytes now load only after an explicit open/download action.\n- Download and forward requests carry the active mailbox instead of hard-coding INBOX.\n- Forwarding fails closed if the source message cannot be loaded.\n- Thread-level and per-message attachment actions hydrate bytes before use.\n- User-added and valid zero-byte attachments are preserved.\n- Added route-level regression coverage for non-INBOX reads and forwarding.\n\nVerification on the rebased head:\n- bun run --cwd apps/email/server test — 60 passed\n- bun run --cwd apps/email/server build — passed\n- bun run --cwd apps/email/client build — passed\n- git diff --check — passed

PR #829, opened 2026-09-04
fix(mail): avoid PostgreSQL port conflicts during mail setup

### LINKED ISSUE #828: [Improvement]: Make mail server sql port configurable
### Existing issue

### Current behavior

Currently if there is a running instance on postgresql's  default port 5432 this will conflict

### Proposed improvement

 It would useful to be able to point the mail server instance to another port. 

### Benefits

This would allow the server's main postgres instance to maintain its default behaviour

### Additional context

_No response_

lagmandu:
Mail setup fails when another PostgreSQL instance already occupies port 5432. New installations now choose a free loopback port in 5432–5460, or use a valid explicit `OPENSHIP_MAIL_DB_PORT`. The selected port is retained across engine recreation, and start-only repairs read the existing sidecar binding. PostgreSQL continues to use port 5432 inside its container.

The mail entrypoint reconciles the SQL connection fields in Postfix, Dovecot, Amavis, and iRedAPD while preserving listener settings, credentials, file ownership, and permissions. Setup preserves existing env-file password values exactly. Port selection only inspects the host; container lifecycle changes remain in the existing start operations.

Addresses #828. This PR targets `feat/bugfix-integration-2026-09-15` for integration before the eventual main merge. Original contributor commits are preserved.

Validation:

- 3,709 adapter tests passed, including port selection, retained state, custom container names, credential preservation, and start-only repairs.
- Nine config reconciliation tests passed against the repository’s real iRedMail configuration templates.
- Seven Docker tests passed against real PostgreSQL, including full bootstrap on a non-default port.
- API and adapter TypeScript checks passed. Fifteen regression cases fail on the original PR implementation and pass with these adjustments.

### DIFF
--- a/apps/email/docker/entrypoint.sh
+++ b/apps/email/docker/entrypoint.sh
@@ -35,9 +35,9 @@ set -euo pipefail
 
 log() { echo "[openship-mail] $*"; }
 
-# No DB_HOST/DB_PORT here on purpose: db-bootstrap.sh reads the same two env vars and
-# owns every conversation with the sidecar, so duplicating them invites the two files
-# to disagree about where the database is.
+# db-bootstrap.sh reads OPENSHIP_MAIL_DB_HOST and OPENSHIP_MAIL_DB_PORT to wait for
+# and bootstrap the schema; step 3c below reconciles OPENSHIP_MAIL_DB_PORT into daemon
+# configs when a non-default port is configured.
 FIRST_DOMAIN="${FIRST_DOMAIN:-}"
 SEED_DIR="/opt/openship-mail/seed"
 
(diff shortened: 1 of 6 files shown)


Hydralerne:
Merged into `feat/bugfix-integration-2026-09-15` after updating this original PR branch for the current architecture, preserving the original contributor commits. Awaiting merge to main.

The fix supports an explicit mail database host port and a bounded automatic fallback when the default is occupied. It preserves existing database bindings and literal credentials on repair, keeps PostgreSQL bound to loopback, and changes only mail SQL connection settings rather than replacing port-like text in passwords or listener settings.

Validation: all GitHub CI checks passed at `00f2b982`; 3,709 adapter tests passed locally, along with 9 real mail-config rewrite tests and 7 Docker/PostgreSQL bootstrap tests, including a non-default port. API and adapter typechecks passed.

Issue #825, opened 2026-09-04
[Bug]: GitHub connection fails during authentication - process gets stuck indefinitely


fredyteheranto:
### Existing issue

- [x] I have searched the existing issues and found no duplicate.

### Problem description

<img width="1813" height="870" alt="Image" src="https://github.com/user-attachments/assets/0f29a14c-51f3-4795-b47d-682e398e9f3a" />

When trying to connect my GitHub account from OpenShip to deploy projects, the authentication process gets stuck in a "Connecting..." state without completing. This prevents me from exploring repositories and setting up automatic deployments.

The error occurs consistently when accessing the "New Project" section and selecting the GitHub connection option. The interface shows the "Connecting..." message indefinitely, never finalizing, forcing the user to reload the page to exit the loading state.

Impact: Completely blocks the main workflow, as GitHub integration is essential for project deployment.

### Steps to reproduce

1. Log in to OpenShip with valid credentials
2. Navigate to the "New Project" section from the main menu
3. Select the "GitHub" option among the available choices (Folder, GitHub, Git URL, Template)
4. Click on the button or link to connect the GitHub account
5. Observe that the system shows the "Connecting..." status indefinitely
6. The process never completes the authentication
7. The page remains in this state without showing error or progress messages

### Environment

- OpenShip version/commit:
- De

ployment method:
- Operating system:
- Browser (if applicable):

### Logs and screenshots

```shell

```

Hydralerne:
Verified already fixed on main; no additional PR merge is needed for this issue.

The relative OAuth navigation issue is already fixed on main: GitHub redirects are resolved against the configured API origin before opening the authentication window. The integration branch includes that fix.

Fix: [34b3d0e6](https://github.com/oblien/openship/commit/34b3d0e6) on [`feat/bugfix-integration-2026-09-15`](https://github.com/oblien/openship/tree/feat/bugfix-integration-2026-09-15).

Validation: URL resolution coverage passes for proxy mode and split API origins. The separate device-code visibility bug (#851) is also fixed on the integration branch.

PR #831, opened 2026-09-04
fix: use API origin for GitHub OAuth redirect

krsnna-05:
Fix GitHub OAuth redirects for hosted split-origin deployments by returning an absolute API URL instead of a dashboard-relative path.

Motivation

#825 

When hosted on separate origins, the dashboard received:

 /api/github/connect/redirect?install_state=... 

The browser resolved this against  https://app.openship.io , resulting in a 404. The redirect must target the API origin at  https://api.openship.io .

Related issue

None — bug fix.

Changes

• apps/api
• Use  resolveApiPublicUrl()  for GitHub OAuth redirect URLs.
• Apply the fix to both App/OAuth connection paths.
• Add a regression test covering the absolute API origin and encoded install state.

Verification

DEPLOY_MODE=desktop bunx vitest run apps/api/test/modules/github/github.controller.test.ts

✓ apps/api/test/modules/github/github.controller.test.ts (4 tests) 106ms
Test Files  1 passed (1)
Tests  4 passed (4)

bun run --cwd apps/api lint

$ tsc --noEmit

The formatter was initially run repository-wide and generated unrelated changes across many files. Those changes were reverted, and only the two changed files were formatted individually:

bunx prettier --write \
  apps/api/src/modules/github/github.controller.ts \
  apps/api/test/modules/github/github.controller.test.ts

Checklist

✓ One change per PR — one bug, or one agreed feature, with nothing unrelated bundled in
✓ The diff is scoped — no reformatting or lint fixes on lines I wasn't otherwise changing
✓ A test fails without this change and passes with it
•  bun run test ,  bun run --cwd <workspace> lint , and  bun format  all pass locally — targeted API tests and lint pass; repository-wide  bun format  was not retained because it modified unrelated files
✓ I understand every line of this diff and can explain it in review

### DIFF
--- a/apps/api/src/modules/github/github.controller.ts
+++ b/apps/api/src/modules/github/github.controller.ts
@@ -22,6 +22,7 @@ import { resolveProjectInfo, projectInfoToScanResponse } from "../deployments/pr
 import { paginateRepoList, type RepoListParams } from "./repo-list";
 import { getRequestContext } from "../../lib/request-context";
 import { hasConfiguredGitHubSource } from "./github-source.service";
+import { resolveApiPublicUrl } from "../../lib/public-url";
 
 /** Map a MappedRepository to the owner/repo key the access filter needs.
  *  `full_name` is canonically "owner/repo"; fall back to the discrete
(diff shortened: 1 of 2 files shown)


Hydralerne:
The reported bug in #825 is already fixed on main and verified in the integration branch.

The linked redirect-origin fix is already handled by current main; PR #831 is superseded.

Main commit 34b3d0e6 resolves API-root redirects through resolveApiNavigationUrl before navigating. It also bounds the popup lifecycle. Reapplying PR #831 would duplicate the origin handling at a different layer. The separate device-code display bug is covered by #851.

Closing this redundant PR as part of the issue/PR reconciliation.

PR #838, opened 2026-09-06
fix(mail): reconcile mounted Let's Encrypt certificates into daemon paths on boot

### LINKED ISSUE #837: [Bug] openship-mail: Postfix and Dovecot fall back to self-signed certificate when Let's Encrypt cert is mounted
### Summary

When \openship-mail\ runs, \/etc/letsencrypt\ is bind-mounted read-only from the host, but Postfix and Dovecot are configured to load \/etc/ssl/certs/iRedMail.crt\ and \/etc/ssl/private/iRedMail.key\.

Because \/etc/ssl\ is inside the container layer (not a persistent volume), any temporary symlink created during initial setup is wiped whenever the container is recreated or upgraded (e.g. \docker pull\ + recreate). Furthermore, \entrypoint.sh\ does not reconcile SSL certificates on boot, causing mail daemons to fall back to the self-signed iRedMail certificate. External TLS clients (such as Node.js, Python, or mail clients) connecting over STARTTLS on port 587 or SSL on 465 reject the connection with:
\\\
Error: self-signed certificate (code: ESOCKET, command: CONN)
\\\

### Steps to reproduce

1. Deploy \openship-mail\ with a custom mail domain (e.g. \mail.example.com\).
2. Provision a Let's Encrypt certificate for the mail domain.
3. Restart or recreate the \openship-mail\ container (e.g. \docker restart openship-mail\ or image update).
4. Connect via Node.js \
odemailer\ or \openssl s_client -connect mail.example.com:587 -starttls smtp\.
5. Notice the mail daemons serve the fallback self-signed certificate (\iRedMail.crt\).

### Expected behavior

\entrypoint.sh\ in \openship-mail\ should reconcile SSL certificates on every boot:
- Check if a valid certificate for \mail.\\ (or \\\) exists in the mounted \/etc/letsencrypt/live/\ directory.
- If present, link it into \/etc/ssl/certs/iRedMail.crt\ and \/etc/ssl/private/iRedMail.key\ so Postfix and Dovecot reliably use the valid Let's Encrypt certificate across container recreations and certificate renewals.

lagmandu:
### Summary

Postfix and Dovecot inside the \openship-mail\ container load certificates from \/etc/ssl/certs/iRedMail.crt\ and \/etc/ssl/private/iRedMail.key\.

Because \/etc/ssl\ is inside the ephemeral container layer (not a persistent volume), recreating the container or updating the image reverts \/etc/ssl\ to the baked self-signed certificates — even when valid certificates are mounted at \/etc/letsencrypt\.

As a result, applications and clients connecting via STARTTLS on port 587 or SSL on 465 (like Node.js / nodemailer, Python, or mail clients) fail with:
\\\
Error: self-signed certificate (code: ESOCKET, command: CONN)
\\\

This PR adds a reconciliation step (\3c\) to \pps/email/docker/entrypoint.sh\ that checks for \mail.\\ (or \\\) under \/etc/letsencrypt/live/\ on every boot and symlinks \ullchain.pem\ and \privkey.pem\ into the default daemon certificate paths.

Fixes #837

### Verification

- [x] Added and ran unit tests: \pps/api/test/lib/mail-ssl-reconcile.test.ts\
- [x] Formatted with Prettier
- [x] Tested certificate reconciliation and TLS handshake with \openssl s_client -connect 127.0.0.1:587 -starttls smtp\ verifying \Verify return code: 0 (ok)\
- [x] Verified Node.js client (\
odemailer.createTransport().verify()\) succeeds without certificate rejection


Hydralerne:
Updated this PR branch for the current architecture and retargeted it to `feat/bugfix-integration-2026-09-15`. The original contribution remains in the commit history.

The startup repair now validates the certificate hostname, validity period, server usage, and matching private key before switching iRedMail to the mounted certificate. It repairs the certificate and key independently, follows stable Let’s Encrypt live paths for renewal, and retains the fallback files when no valid mounted pair exists.

Validation: 10 Docker tests passed with real OpenSSL certificates and a TLS handshake, covering stale key links, repeat starts, renewal, container recreation, invalid certificates, and apex SAN/wildcard certificates. API typechecking passed. CI also exposed two older integration test fixture problems; those were corrected on the integration branch and merged into this PR. All CI jobs passed in run 35008541433. This original PR is now merged into the bug-fix branch through 1a4f535e, awaiting the eventual merge to main.

PR #855, opened 2026-09-08
fix(dashboard): rescan deployment configuration when branches change

### LINKED ISSUE #852: [Bug]: Switching a project's branch in the deploy config UI never re-scans the repo, stale framework/compose detection from the previous branch is kept and used to deploy
### Existing issue

### Problem description

A project's repository has Docker Compose config (_a `docker-compose.yml` and related files_) on its `main` branch, and zero references to Docker anywhere on a different branch (`openship`), a plain non-containerized deploy target on that branch.  Importing/configuring the project against `main` correctly detects "Docker Compose" with 3 services.  Switching the branch selector to `openship` should re-scan the repo at that ref and detect there's no compose config there at all.   Instead, the Framework panel keeps showing "Docker Compose · Detected" completely unchanged, and attempting to deploy fails because the backend (_correctly_) can't find any compose services on that branch.

## Evidence
Watched the browser's network tab across the branch switch (_step 3 above_): **zero requests fired.** No call to any branches/scan/detect-framework endpoint, nothing.  The dropdown selection changes client-side state only (_the label updates from `main` to `openship`_), with no corresponding re-fetch of the repository at the new ref. Confirmed this both directions (_`openship` → `main` and back_), neither switch produces any network activity at all, and the Framework panel is byte-for-byte identical regardless of which branch is selected.

## Expected
Switching branches re-scans the selected branch's tree and updates the Framework panel (_and any dependent config, like the compose file path_) to match what's actually present there, including correctly showing "no framework detected" / prompting manual configuration when a branch has no recognizable deploy config at all, rather than silently keeping whatever was detected for a previously-selected branch.

## Actual
The branch dropdown is purely cosmetic with respect to framework detection, it changes which branch will be *deployed*, but the detected framework, service count, and defaults shown above it are permanently whatever was scanned for the branch that was selected at project-creation/ first-scan time, regardless of how many times the branch selector is changed afterward.

## Impact
Any project that deploys different branches with genuinely different structures (_a common pattern e.g. a "no infra changes on this branch" convention, environment-specific branches, or, as here, a branch that predates a Docker migration_) gets a confusing, misleading config screen that looks correct and lets the user proceed to a deploy that's guaranteed to fail, with no indication beforehand that the framework info on screen doesn't apply to the branch actually selected.

### Steps to reproduce

1. Import/configure a project whose default branch has Docker Compose config and at least one other branch that doesn't.
2. On the project's deploy config page, note the Framework panel: "Docker Compose · Detected."
3. Open the branch dropdown (_top-right, next to the repo name_) and switch to the branch with no Docker files.
4. Observe: the Framework panel is unchanged, still "Docker Compose · Detected," still "docker-compose defaults applied" in the Deploy Configuration section below it.
5. Deploy. Get:
   ```
   Error
   Could not initialize compose services from "repository root": The
   configured compose path "repository root" contains no services.
   ```
which is the backend correctly reporting reality for the selected branch, while the UI never reflected that reality before letting the deploy proceed.

### Environment

- OpenShip version/commit: Openship version: 0.7.2 (self-hosted)
- Operating system: Ubuntu 26.04.1 LTS
- Project source: GitHub, Docker Compose framework, private repo
- Reproduced on `/deploy/<slug>?projectId=<id>&mode=config`

### Logs and screenshots

```shell

```

## Suggested angle for a fix
The branch dropdown's `onChange` (_or equivalent_) needs to trigger the same repo-scan/framework-detection call that runs on initial project import, scoped to the newly-selected branch/ref, and update the Framework panel and any branch-dependent defaults (_compose file path, detected services, etc._) from that fresh result, including correctly clearing/flagging when the newly-selected branch has no detectable framework at all.

AbdullahM07:
## Summary

Changing the branch in the deploy wizard now scans the selected ref and applies the branch and its detected configuration together. A branch without a recognizable framework opens the manual picker instead of retaining the previous branch's Compose detection.

## Motivation

The selector only updated `config.branch`. Repository initialization also preferred the saved branch and build defaults over an explicit branch override, so deployments could use a different tree from the one described by the UI.

## Related issue

Fixes #852

## Changes

- Rebuild framework, Compose services/path, monorepo apps, and mode defaults from the selected branch; preserve project env edits, destination settings, and single-app domains/redirects.
- Serialize branch and Compose-path scans, disable Save/Deploy while scanning, and retain the last valid configuration with an inline error if scanning fails.
- Persist the chosen branch on config save and send the single-app deployment mode explicitly so retained service rows cannot override the new branch's detected topology.
- Add nine interaction regressions using the actual selector and deployment hooks, with Happy DOM as a test-only dependency. Keep the manual framework picker within mobile widths.

## Verification

```text
bun run test -- --filter=!openship -- --maxWorkers=2
  6 workspace tasks successful; dashboard: 114 files, 1190 tests passed

bun run --cwd apps/cli test --maxWorkers=2 --testTimeout=15000
  38 files, 496 tests passed

bun run --cwd apps/dashboard lint
  passed

bun format
  passed in an isolated checkout; unrelated existing formatting excluded from the diff
```

The initial full run hit an existing CLI port-test timeout: this machine already uses port 3001, so the resolver waits six seconds against Vitest's five-second default. The entire CLI suite passes with the timeout shown above; no CLI code or tests were changed.

Regression proof: all eight original branch-switch cases failed against `upstream/main`; the final suite passes all nine cases, including domain preservation. Headless Chromium checks of loading, error, and manual-selection states passed at 1280px and 390px with no horizontal overflow.

## Checklist

- [x] One bug fix, with no unrelated changes bundled in
- [x] Scoped diff with unrelated formatting excluded
- [x] Regression cases fail before the fix and pass afterward
- [x] Workspace tests, dashboard typecheck, and formatting verified as described above
- [x] Reviewed the complete diff


Hydralerne:
Incorporated this PR into [`feat/bugfix-integration-2026-09-15`](https://github.com/oblien/openship/tree/feat/bugfix-integration-2026-09-15) via [c45bcf5d](https://github.com/oblien/openship/commit/c45bcf5d), preserving the original commits and adapting the fix to the current architecture.

Integrate PR #855: rescan branch changes, discard stale Compose defaults, and save the new branch with its settings in one platform update.

Validation:

- 14 dashboard branch-rescan/picker tests and 10 actual native/HTTP GitHub/project parity tests pass. API and dashboard type checks pass.
- The branch dropdown regression fails without the updated event wiring (zero prepare requests), and passes with the fix.

PR #862, opened 2026-09-08
fix(dashboard): keep GitHub device sign-in visible until completion

### LINKED ISSUE #851: [Bug]: "Sign in with GitHub" (device flow) silently does nothing, backend returns a valid device code but the UI never displays it
### Existing issue

## Summary
Clicking "Sign in with GitHub" under Settings → Git → GitHub does not surface a device code anywhere in the UI, no modal, no toast, nothing.  The button appears to do nothing at all, so there's no way to complete browser-based device-flow sign-in through the dashboard right now.

This is not a backend problem: the API call the button makes still returns a completely valid device code, confirmed both through the browser's own network tab and independently via direct API/CLI calls.  The bug is entirely in the frontend failing to do anything with a successful response.

## Evidence

**Network tab, captured live on the click:**
```
POST /api/proxy/api/github/connect  → 200
GET  /api/proxy/api/github/status   → 200
```
No failed requests, no 4xx/5xx, no console errors or exceptions logged at all during or after the click.

**Confirmed the backend response itself is valid**, by calling the same underlying endpoint directly (_bypassing the dashboard entirely_):
```
$ openship api /github/connect -X POST
{
  "connected": false,
  "flow": "device_code",
  "userCode": "3DD7-7B49",
  "verificationUri": "https://github.com/login/device",
  "expiresIn": 899,
  "interval": 5
}
```
That's a completely normal, usable device-code response visiting the verification URL and entering the code works fine and completes the connection (_confirmed via `/github/status` afterward showing `available: true`_).  So the backend half of this flow is entirely healthy.

**DOM inspection immediately after clicking the button** (_accessibility tree, full page, `filter: all`_) shows no dialog/modal element anywhere on the page, hidden or visible, nothing was ever rendered for the returned device code.  The only visible change after clicking is the "Last checked" timestamp on the existing status banner ticking forward, because the button's click handler also re-triggers the status check, which (_unsurprisingly_) still reports the old credential as rejected since nothing new was ever authorized.

## Expected
Clicking "Sign in with GitHub" shows the returned `userCode` and a link/instructions to `verificationUri`, ideally polling `/github/status` in the background until the device is authorized, then updating the banner to reflect the new connection.

## Actual
The device code is fetched successfully and then silently discarded by the frontend.  There is currently no way to complete device-flow sign-in through the dashboard UI at all.

## Impact
Anyone whose stored GitHub credential expires or gets revoked (_which itself happens periodically, see the related, separate report about credentials needing periodic re-auth_) has no way to reconnect through the dashboard.  The only working path we found is calling the API directly (_`openship api /github/connect -X POST`, or an equivalent raw HTTP call_) to get the code, then manually visiting `github.com/login/device` not a reasonable expectation for a normal user, and not documented anywhere in the UI as a fallback.

### Steps to reproduce

1. Get into a state where the stored GitHub connection is rejected (_or just have no connection configured_) so "Sign in with GitHub" is the primary call to action.
2. Go to Settings → Git.
3. Click "Sign in with GitHub."
4. Observe: nothing happens. No modal appears, no code is shown, no verification link, nothing.  The page just quietly re-runs its status check and re-shows the same "rejected"/disconnected banner as before, since obviously nothing new was ever completed.

### Environment

- OpenShip version/commit: `0.7.2` (_self-hosted, `openship up`_), regression present since at least `0.7.0`
- Deployment method: Self-hosted, docker compose
- Operating system: Ubuntu 26.04.1 LTS
- Browser (if applicable): Google Chrome.

### Logs and screenshots

```shell

```

## Suggested angle for a fix
Given the API call and its response are both fine, this is very likely a regression introduced in the Settings → Git page rewrite that shipped in `0.7.0` (_new "Self-hosted GitHub Apps" section, new "Other methods" dropdown with GitHub App / SSH key / Access token options_) plausibly whatever component used to render the device-code modal got disconnected or dropped during that rework, while the button's click handler still correctly fires the API calls underneath it.

AbdullahM07:
## Summary

Keep the GitHub device code and verification link visible until that grant completes, then refresh Settings to show the newly connected identity.

## Motivation

Settings hid the pending action whenever any GitHub source was connected. The shared provider also discarded it based on an existing App connection or a stale CLI identity. A valid device-code response could therefore disappear immediately, and successful polling did not update Settings' separate status snapshot.

## Related issue

Fixes #851

## Changes

- Give the pending login panel precedence over connection badges and background status loading.
- Complete device grants through their existing poll endpoint; reserve status-based completion for terminal login.
- Refresh the Settings snapshot when a pending device/token action finishes.
- Exercise the real provider and Settings component in five interaction regressions, using Happy DOM as a test dependency.

## Verification

- Before the fix: four of the five regression cases fail, covering stale connectivity, an existing App, pending status refresh, and the final connected identity.
- `bun run --cwd apps/dashboard test --maxWorkers=2`: 114 files, 1186 tests passed.
- `bun run --cwd apps/dashboard lint`: passed.
- Root non-CLI workspace tests: all six tasks passed.
- Ran `bun run format` in an isolated checkout and retained only formatting within this change.
- Rendered the actual click/poll states in Chromium at 1280×800 and 390×844; code, verification link and connected identity remain visible without horizontal overflow. UI detector reported no findings.

## Checklist

- [x] One issue per PR, with a scoped diff
- [x] Regression tests fail before the fix and pass afterward
- [x] Relevant tests, typecheck and formatting completed locally
- [x] Reviewed the implementation and test behavior


Hydralerne:
Incorporated this PR into [`feat/bugfix-integration-2026-09-15`](https://github.com/oblien/openship/tree/feat/bugfix-integration-2026-09-15) via [cfdea39b](https://github.com/oblien/openship/commit/cfdea39b), preserving the original commits and adapting the fix to the current architecture.

Pending GitHub device codes stay visible across stale status refreshes and account changes.

Validation:

- Five DOM interaction tests passed
- Two device-flow regressions failed against unchanged main

PR #863, opened 2026-09-08
fix(cli): require folder upload opt-in and isolate archive staging

### LINKED ISSUE #853: [Bug] openship deploy: folder upload tarball is written inside its own archive root when cwd is $TMPDIR — orphaned tarballs nest and fill the disk
### Summary

`openship deploy`'s folder-upload path writes its staging tarball into `os.tmpdir()` but sets the archive root to `cwd`, and excludes only `node_modules` / `.git` / `.DS_Store` — never the tarball itself. If `cwd` is `$TMPDIR` (or any ancestor of it), `tar` archives its own growing output. Because the `rmSync` cleanup sits in a `try/finally` that begins *after* the `tar` call, the failed tarball is never removed — so each subsequent run swallows every earlier orphan and the file sizes grow geometrically.

On our self-hosted box this filled a dedicated 25 GB `/tmp` filesystem to 0 bytes free, which fails every `/tmp` write host-wide (not just Openship's).

### Code

`src/lib/folder-deploy.ts` → `deployFolder()` (read from the installed `openship` CLI, `dist/index.js`):

```js
const tarball = join(tmpdir(), `openship-upload-${session.sessionId}.tar.gz`);
execFileSync(
  "tar",
  ["-czf", tarball,
   "--exclude=./node_modules", "--exclude=./.git", "--exclude=./.DS_Store",
   "-C", cwd, "."],
  { stdio: "ignore" }
);
step("Uploading source");
try {
  const body = readFileSync(tarball);
  // ... upload ...
} finally {
  try { rmSync(tarball, { force: true }); } catch {}
}
```

Two independent defects:

1. **`tarball` can live inside the archive root.** `-C cwd .` with the output at `join(tmpdir(), …)` is only safe while `cwd` is not `$TMPDIR` or an ancestor of it. Nothing excludes `basename(tarball)`.
2. **Cleanup is unreachable on the failure that this causes.** `execFileSync` throws when `tar` exits non-zero, which is *before* the `try`, so the orphan is left on disk. `deployFolder` then throws, so no project or deployment record is created — the leak leaves no trace in the Openship UI or API, which makes it hard to attribute.

### Reproduction

```console
$ mkdir -p /tmp/repro/src && head -c 3000000 /dev/urandom > /tmp/repro/src/payload.bin
$ cd /tmp/repro
$ tar -czf /tmp/repro/out.tar.gz --exclude=./node_modules --exclude=./.git -C /tmp/repro .
tar: .: file changed as we read it
$ echo $?
1
```

That is the exact argv the CLI builds when `cwd` and `tmpdir()` coincide. Running `openship deploy` from `/tmp` reproduces it end to end.

### Observed impact

Five orphaned `/tmp/openship-upload-*.tar.gz` files, 18.8 GB total, on a 25 GB filesystem. Nesting is directly verifiable — the 4.33 GB orphan contains the 2.16 GB orphan written 90 seconds earlier as a member:

```console
$ tar tvzf openship-upload-<b>.tar.gz | sort -k3 -nr | head -1
-rw-r--r-- root/root  2164768151 2026-09-06 15:50 ./openship-upload-<a>.tar.gz
```

Sizes across the set: 2.16 GB → 4.33 GB → 6.52 GB. None of the archives contained any project source — only the contents of `/tmp`.

### Suggested fix

- Stage into a per-run directory outside `cwd`: `const dir = mkdtempSync(join(tmpdir(), "openship-upload-")); const tarball = join(dir, "source.tar.gz");` and `rmSync(dir, { recursive: true, force: true })` in the `finally`. (The server-side path in `src/…/server` already uses `mkdtemp` this way.)
- Move the `execFileSync("tar", …)` call *inside* the `try` so the `finally` also covers a `tar` failure.
- Belt and braces: add `--exclude=./<basename(tarball)>`, and refuse to run when `resolve(cwd)` is `tmpdir()` or a prefix of it, with a clear error rather than a bare `tar` exit code.

### Environment

- `openship` CLI installed at `/usr/lib/node_modules/openship` (self-hosted, Ubuntu, GNU tar)

AbdullahM07:
## Summary

Require an explicit folder upload choice and stage its archive outside the source tree. Clean up the staging directory even when `tar` fails before upload.

## Motivation

Running `openship deploy --project ... --service-ids ...` from a non-Git scratch directory silently uploaded that directory. When it contained `TMPDIR`, the archive could include earlier uploads and its own output. Packaging happened before the cleanup block, leaving failed archives behind to grow on subsequent attempts.

## Related issue

Fixes #853

## Changes

- Add `--folder` for explicit upload, including from a Git checkout; preserve `--name` as an upload opt-in outside Git.
- Refuse implicit uploads outside Git. An explicit `--branch` selects the existing Git deployment endpoint; reject Git-only options combined with `--folder`.
- Resolve symlinks before rejecting the temporary directory or an ancestor as the source, before provisioning an upload session.
- Use a private per-run staging directory and cover packaging and upload with the same `finally` cleanup.
- Update CLI documentation and add real-tar regressions plus command-level coverage for both deployment paths.

## Verification

- Before the fix: four archive regressions fail, including a partial tarball surviving a packaging error.
- `bun run --cwd apps/cli test --maxWorkers=2 --testTimeout=15000`: 40 files, 506 tests passed. The final two explicit-upload cases also pass in `bun run --cwd apps/cli test deploy-folder-opt-in --maxWorkers=2` (6 tests).
- `bun run --cwd apps/cli lint`: passed.
- Root non-CLI workspace tests: all six tasks passed.
- Ran `bun run format` in an isolated checkout and retained only scoped formatting.

The CLI suite uses a 15-second timeout because its existing port-release test waits six seconds when local port 3001 is occupied; the running local server was left in place.

## Checklist

- [x] One issue per PR, with a scoped diff
- [x] Regression tests fail before the fix and pass afterward
- [x] Relevant tests, typecheck and formatting completed locally
- [x] Reviewed the implementation and test behavior


Hydralerne:
Incorporated this PR into [`feat/bugfix-integration-2026-09-15`](https://github.com/oblien/openship/tree/feat/bugfix-integration-2026-09-15) via [e1082f44](https://github.com/oblien/openship/commit/e1082f44), preserving the original commits and adapting the fix to the current architecture.

Integrate PR #863 in the shared platform/SDK packager: reject staging within source and require an explicit folder-upload choice.

Validation:

- 41 platform, SDK and CLI tests pass; platform and CLI type checks pass.
- All three source/staging overlap regressions fail against main (unsafe archives were created), and pass with the guard. Cases include exact paths, ancestors, symlink aliases/parents, common prefixes, retained build outputs, and session/upload failure cleanup.

Issue #844, opened 2026-09-06
[Bug] service.environment silently overrides project env on every deploy path (and stores secrets in plaintext)


greerso:
Self-hosted Openship 0.7.2. Two related findings about how service-level environment is stored and applied. The first cost us a botched credential rotation that reported success; the second is a hardening point.

## 1. `service.environment` silently overrides the project env store, on every deploy path

A key present in the `environment` jsonb on a `service` row is applied at container creation and **wins over the project-level env store**. `openship project env set` on such a key:

- stores the new value correctly,
- reports success,
- is carried into the deployment snapshot,

…and the container still comes up with the **old** service-level value. No warning anywhere.

Keys *absent* from `service.environment` follow the project store normally, for both upserts and deletions — which is exactly what makes this hard to diagnose. Half your changes land and half do not, with no signal distinguishing them.

**Every deploy path we tried is affected**: `deploy --refresh`, `deployment redeploy`, and a full git rebuild producing a brand-new image. All three reported success and changed nothing.

### Why it is worse than a normal precedence rule

Precedence between layers is fine and expected. The problem is that the losing layer is the one the CLI writes to and reports on, so the tool tells you the change landed. We hit this rotating an OAuth signing key: the CLI said the key was updated, the deploy went `ready`, and the application kept signing with the old key. The only way to find out was `docker exec <container> printenv <KEY>`.

### Reproduce

```bash
# 1. Pin a key at the service layer
openship api "/projects/<projId>/services/<svcId>" -X PATCH -d '{"environment":{"FOO":"from-service"}}'

# 2. Change it at the project layer — reports success
openship project env set <projId> --set FOO=from-project

# 3. Deploy by any path, then look at what the container actually got
docker exec <container> printenv FOO      # -> from-service
```

### Suggested fixes, roughly in order of value

1. **Warn on the write.** When `project env set` targets a key that is pinned in some `service.environment` in that project, say so on stdout and name the service. This alone would have saved the incident.
2. **Surface it in the deployment output**, listing keys whose effective value came from the service layer rather than the project store.
3. **Expose the effective environment** — an `openship project env get --effective` (or `--resolved`) that shows, per key, the value that will actually reach the container and which layer it came from.
4. Document the precedence in the env docs. It is currently discoverable only by querying the database.

The supported workaround, once you know, is to write the service layer directly (merge semantics; omit a key to keep it, `null` to remove) and then update the project layer so the two agree:

```bash
[...]

PR #864, opened 2026-09-08
fix: protect build arguments across API and native SDK

### LINKED ISSUE #854: [Bug] buildArgs returned unmasked in the deployments API while the same key is masked in environment
## Summary

Openship masks secret values in `environment` but returns `buildArgs` **in the clear** from the deployments API. Because deployment records are retained per build, every secret a project has ever passed as a build arg remains readable to any holder of an Openship API token, indefinitely. Rotating the secret does not remove the old value from history — it adds a second readable value alongside it.

The masking logic already exists and works correctly. It simply is not applied to `buildArgs`.

## Reproduce

On a project whose service passes a secret as a build arg (e.g. a Next.js app that needs an auth secret present during static analysis):

```
GET /projects/<project_id>/deployments

  meta.composeServices[].environment.<SECRET_NAME>  ->  "••••••••"    masked
  meta.composeServices[].buildArgs.<SECRET_NAME>    ->  <plaintext>   not masked
```

Every secret-shaped key under `environment` in the same response is masked correctly. Only `buildArgs` is unmasked.

The same asymmetry appears on the service endpoints — `GET`/`PATCH /projects/<id>/services/<svc>` returns `buildArgs` verbatim, so a routine inspection of a service definition prints secrets to whatever is reading.

## Why retention makes it worse

Observed on one project with 20 retained deployment records: 17 carried the build-arg secret, none masked. After rotating that secret, the history contained **two distinct recoverable values** — the previous secret in the older records and the new one in the records created by the rotation. Both readable in a single API response.

So the exposure is not "the current secret is readable". It is "every secret this project has ever built with is readable, and rotating adds to the set rather than clearing it".

## Why "just don't pass secrets as build args" isn't a complete answer

It is the right advice and we are doing it on our side, but it does not resolve the platform behaviour:

- The build arg is often load-bearing. Next.js evaluates module-level initialisation during `collecting page data`, so a framework that constructs an auth client at import time needs the value present at build time or the build fails. Several projects added the build arg specifically to fix that failure.
- Existing deployment records already contain historical values. Even a project that stops passing build args today keeps its old records readable.

A dummy literal in the Dockerfile works where the value only needs to be non-empty at build time, but that is a per-app workaround for a platform disclosure.

## The tension a fix needs to resolve

Unmasked `buildArgs` is currently also the **only read-back channel** for verifying that an env write actually landed. `environment` is masked on every read, and there is no endpoint that returns an effective value, so operators verifying a rotation end up relying on the unmasked `buildArgs` field, or on `docker exec printenv` in the running container after a recreate.

Masking `buildArgs` without replacing that channel would make correct rotations harder to verify, which tends to push people toward riskier verification habits.

## Suggested direction

1. Apply the existing `environment` masking to `buildArgs` in all API responses (deployments, services).
2. Scrub or mask build-arg values in **retained** deployment records, not just new ones — otherwise the fix does nothing for the accumulated history, which is the larger part of the exposure.
3. Provide a non-disclosing verification channel so rotations stay verifiable: a per-deployment digest (e.g. a short hash) of the effective value would let an operator confirm a write landed without the value being returned.

Item 3 is what makes 1 and 2 adoptable rather than something operators work around.

## Notes

- Reported from a self-hosted deployment. Happy to provide further detail privately if you would prefer this handled through a security contact rather than a public issue — say the word and I will close this and re-send.
- Related to the `environment` / `service.environment` precedence behaviour in #844, in that both concern how env values are stored and surfaced, though the fixes are independent.

AbdullahM07:
Service and deployment reads exposed current and historical build-argument secrets. This change masks non-empty values at the shared platform boundary and recovers echoed masks on writes, preserving working builds and rollback snapshots. Fixes #854 when the integration branch reaches main.

- Covers service CRUD, Compose sync and drift, source scans, deployment history and build status through both HTTP and the native SDK.
- Preserves empty strings, inherited `null` arguments and whole-map replacement. Masked arguments retain their stored interpolation provenance; new literal values cannot inherit a stale template marker.
- Adds project/service/key-scoped HMAC fingerprints for comparing stored literal values without disclosing them. Inherited or interpolated arguments are excluded; fingerprints do not attest running-container values.
- Keeps explicitly authorized `includeEnv: true` source editing scans functional. Internal merge baselines stay private, and retained snapshots remain usable for rollback.

Abdullah Mohamed's original commit is retained. The PR is updated for the current shared platform architecture and targets `feat/bugfix-integration-2026-09-15`, ahead of the draft integration PR #891.

Validation with repository-pinned Bun 1.3.10:

- 292 targeted API tests passed, including six tests using real PGlite, PAT authentication, service writes and native/HTTP deployment reads.
- 29 SDK tests passed, including real native-worker source preparation and staged-upload scans.
- API typecheck and documentation checks passed (including SDK examples).
- Historical-disclosure and masked-sync regressions both fail on the pre-fix branch and pass with this change.

The separate encryption-at-rest and environment-precedence concerns in #844 remain under review.


Hydralerne:
Merged into `feat/bugfix-integration-2026-09-15` via 8c20a9b3, with Abdullah Mohamed's original bcdc8401 commit preserved. Main is unchanged; the combined review is draft PR #891.

The adaptation moves protection into the shared platform, exposes fingerprints in the SDK contract, preserves masked build-argument interpolation on Compose sync, and keeps authorized source editing scans usable. Real PGlite/native/HTTP tests cover writes, old and new deployment snapshots, build status, drift, read-only tokens and tenant isolation.

Validation: 292 API tests and 29 SDK tests passed; API typecheck and documentation checks passed. Both historical-disclosure and masked-sync regressions fail on the pre-fix code and pass after the change. Every GitHub CI job passed: https://github.com/oblien/openship/actions/runs/35015115546

Issue #854 is fixed on the integration branch, awaiting merge to main. The separate #844 work remains open.

PR #871, opened 2026-09-09
feat(compose): preserve container hardening controls

### LINKED ISSUE #749: feat(compose): preserve container hardening controls
## Problem

OpenShip now correctly reports critical Compose fields that it cannot preserve, following #533. Five common runtime-hardening controls are still reported as unsupported and cannot be represented by a native OpenShip service:

- `read_only: true`
- `cap_drop`
- `security_opt` (especially `no-new-privileges:true`)
- `tmpfs`
- `user`

For a security-sensitive container, importing through OpenShip therefore requires either weakening the container or keeping the workload outside native OpenShip management. The current fail-closed reporting is the right behavior; this proposal is to make these fields genuinely supported rather than suppress their warnings.

## Proposed scope

Preserve these controls through the same full lifecycle as the existing `ComposeAdvanced` fields:

1. raw Compose parsing and the CLI `docker compose config` mapper;
2. strict API validation and the `advanced` JSONB service model;
3. Compose sync, deployment snapshots, drift comparison, and rollback;
4. Docker runtime creation and inspection/readback;
5. explicit unsupported-runtime reporting for adapters that cannot honor them.

For Docker, the final container shape should map to the native Engine fields:

- `user` → top-level `Config.User`
- `read_only` → `HostConfig.ReadonlyRootfs`
- `cap_drop` → `HostConfig.CapDrop`
- `security_opt` → `HostConfig.SecurityOpt`
- `tmpfs` → `HostConfig.Tmpfs`

The implementation should preserve meaningful false/empty/clear semantics and validate malformed values before deployment. It should not silently downgrade a requested control. Long-form `volumes: [{ type: tmpfs, ... }]` can remain a separately blocked shape if supporting the top-level Compose `tmpfs` field keeps this contribution reviewable.

Because `advanced` is JSONB, this appears not to require a database migration.

## Acceptance criteria

- Raw-file import and CLI sync produce the same normalized `advanced` representation for all five fields.
- Service create/update accepts, validates, persists, reads back, and merge-patches the fields without erasing unrelated advanced settings.
- Deployment snapshots, drift detection, and rollback retain the exact hardening request.
- Docker create applies all five fields, and an inspect-based test proves the resulting container shape.
- Adapters that cannot implement a field report it explicitly through their unsupported-Compose capability rather than silently dropping it.
- Tests cover each field independently, the combined hardened-service shape, defaults/clears, invalid input, import/sync, and Docker mapping.

## Contribution question

Would you accept this as one end-to-end contribution, or would you prefer the model/parser work and Docker runtime work split into separate PRs? I can implement it once the intended scope is confirmed.

scubashack808:
## Summary

Adds `read_only`, `cap_drop`, `security_opt`, `tmpfs` and `user` to the `advanced` seam, so a compose
file that hardens a container gets a hardened container instead of a warning. One `@repo/core`
authority module defines what the five accept and how they normalize, imported by the raw Compose
parser, the CLI sync mapper and the Docker runtime, following the pattern `compose-namespace.ts` sets
for exactly this problem.

## Motivation

Write any of these five in a compose file today and Openship imports the stack, reports honestly that
the field is not modeled, and runs the container without it. The reporting is correct; the outcome is
that a security-sensitive workload either runs weaker than its file specifies or stays off Openship
entirely. They are the set [OWASP's Docker Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Docker_Security_Cheat_Sheet.html)
recommends combining.

`cap_add`, `privileged`, `devices` and `sysctls` are deliberately not here. The line is which way the
omission fails: dropping those leaves a container MORE restricted than its file asked for, so silence
is safe, while dropping any of these five leaves it weaker while the deploy reports success. Honoring
the first group decides Openship's posture rather than preserving the operator's, and that seemed
like yours to make separately. That test does reach wider than the five, `pids_limit` being the
clearest neighbour it also catches, and this PR ships only the published OWASP set rather than a line
I drew myself.

On process, so nothing is overstated: you welcomed a PR here, I posted the shape I intended on
2026-08-31, and I built on that shape without waiting for an answer, so every boundary here is
movable and I would rather move one now than after merge. Two corrections to my own acceptance
criteria, both in that comment: these are round-trip only and not
authored through the API, following how `AdvancedSchema` already treats `networkMode`, `pidMode` and
`entrypoint` because the compose file owns them; and the tests are ones that can fail rather than a
coverage matrix.

## Related issue

Closes #749

## Changes

- **`packages/core`** New `compose-hardening.ts`, the single authority: what the five accept and how
  they normalize (`parseComposeHardening`), the Engine create payload (`dockerHardening`), that shape
  read back off a live container (`inspectHardening`), plus `pickHardening` and
  `COMPOSE_HARDENING_KEYS`. `ComposeAdvanced` gains the five flat, so they store, diff and round-trip
  exactly like `networkMode`. JSONB, so no migration, and drift, snapshots and rollback needed no
  change because they carry `advanced` wholesale; the round-trip tests pin that rather than assume it.
- **`apps/api`** `parseAdvanced` reads the five through the authority and they leave
  `UNSUPPORTED_SERVICE_KEYS` in the same commit; the authority says per issue whether it blocks.
  `service.schema.ts` accepts them for round-trip, bounds only.
- **`apps/cli`** `service sync` maps them off `docker compose config` through the same authority, and
  a value that cannot be honored exits non-zero rather than uploading a list that omits it.
- **`packages/db`** `COMPOSE_OWNED_ADVANCED_KEYS` gains the five, spread from
  `COMPOSE_HARDENING_KEYS` rather than respelled. This is the only way a control is turned off.
- **`packages/adapters`** Docker create maps `user` to `Config.User` and the other four onto
  `HostConfig.ReadonlyRootfs` / `CapDrop` / `SecurityOpt` / `Tmpfs`, spread rather than assigned so an
  unrequested control leaves no key behind. `inspectContainer` reads them back, and the cloud runtime
  declares all five in `unsupportedComposeKeys`.
- Two edges: the long-form `volumes: [{type: tmpfs}]` refusal stays and its message now names the key
  that is honored; and a container adopted with no compose file describing it keeps the confinement
  it is running with, as adoption already keeps its cpu and memory caps.

**What this can now refuse that imported yesterday.** A behavior change worth stating plainly: before
this, no compose key could refuse an import for a hardening reason, and now five shapes do. A
`read_only` that is not one of compose's own boolean spellings, an unparseable `user`, a relative
`tmpfs` target, two `tmpfs` entries on one target, and a `security_opt` entry with no separator and no
bare form. A blocked import is annoying; a silently unconfined container is the failure this PR exists
to prevent, so the refusals go the safe way, and if you would rather any of them warn instead that is
one line per case. Two shapes were checked against the tools rather than assumed, because either would
otherwise have refused something legal: `read_only` is read after interpolation and through compose's
own boolean casting (`toBoolean` in compose-go's `loader/interpolate.go`, confirmed by measuring
Compose 2.40.3), so `read_only: ${RO}` and `read_only: "true"` import identically through both doors;
and `security_opt` is split the way the daemon splits it, at the first `=` when there is one and only
then at `:` (moby `daemon/daemon_unix.go`, `parseSecurityOpt`), so `label=user:USER` and
`label:user:USER` both parse.

**An entry asking for `unconfined` is reported and not applied.** `seccomp=unconfined`,
`apparmor=unconfined` and their neighbours ask for LESS confinement than your default, which is the
one direction this PR does not take the file's word for, and it is the call Openship already makes on
`cap_add`: the container keeps Docker's default profile, the report is not blocking, and the rest of
the list still lands. The API's import report carries that line; `service sync` has no channel but
its errors, so on that door the entry is dropped without a message, which is how it already treats
`cap_add`. Say the word if you would rather store them verbatim and let the daemon decide,
because it is a few lines in one place and I would rather move it before merge than after.

**`read_only: false` and `cap_drop: []` are never stored, and that is forced rather than chosen.**
`docker compose config` erases both from its normalized output, so the CLI door cannot see them at
all, and storing them on the YAML door would make one file produce a different `advanced` through each
door and report drift no operator could resolve; clearing happens through the key going absent, which
is what `COMPOSE_OWNED_ADVANCED_KEYS` is for. Capability spelling is kept as written for the same
reason: the Engine API stores what it is given and `net_raw`, `NET_RAW` and `CAP_NET_RAW` all clear
the same bit, so normalizing would only make the inspect readback disagree with the file.

## Verification

Bun 1.3.10, Node 22.22.0, on a quiet Linux host, from the committed lockfile. Every number is from a
run. Each suite was also run with this branch's test files against unmodified `main`, to confirm the
tests can actually fail:

| suite | on `main` @ `30ae48a4` | on this branch |
| --- | --- | --- |
| `packages/core/src/compose-hardening.test.ts` | fails to load, the module does not exist | **28 passed** |
| `packages/adapters/src/runtime/docker-hardening.test.ts` | **2 failed**, 1 passed | **3 passed** |
| `apps/api/test/lib/compose-parser.test.ts` | **6 failed**, 121 passed | **127 passed** |
| `apps/api/test/modules/deployments/hardening-survives-deploy.test.ts` | **1 failed**, 4 passed | **5 passed** |
| `apps/api/src/modules/migration/docker-reconcile.test.ts` | **3 failed**, 36 passed | **39 passed** |
| `apps/cli/test/unit/service-sync-mapping.test.ts` | **4 failed**, 13 passed | **17 passed** |
| `apps/api/test/e2e/compose-hardening.e2e.test.ts` (real daemon) | **3 failed** | **3 passed** |

```
FAIL docker-hardening.test.ts > splits the five across Config.User and HostConfig
  AssertionError: expected undefined to be '1000:1000'
```

One service deployed through the product's own `deployServiceWorkload` with all five set, then
inspected. Real busybox container, Docker Engine 29.8.0, captured at commit `40acf919`:

```
=== docker inspect (the fields the compose file asked for) ===
{
  "Config.User": "1000:1000",
  "HostConfig.ReadonlyRootfs": true,
  "HostConfig.CapDrop": [
    "ALL"
  ],
  "HostConfig.SecurityOpt": [
    "no-new-privileges:true"
  ],
  "HostConfig.Tmpfs": {
    "/run": "size=64m,mode=1777"
  }
}
=== container stdout (what it can actually do, from the inside) ===
tmpfs-write=ok
rootfs=readonly
uid=1000:1000
NoNewPrivs:1
CapEff:0000000000000000
=== inspectContainer readback (the shape adoption stores) ===
{
  "user": "1000:1000",
  "readOnly": true,
  "capDrop": [
    "ALL"
  ],
  "securityOpt": [
    "no-new-privileges:true"
  ],
  "tmpfs": [
    "/run:size=64m,mode=1777"
  ]
}
```

Shape and effect are asserted separately on purpose: a container can carry `CapDrop: ["ALL"]` in an
inspect and still have been started before the field reached the daemon, so the test also asks the
container from the inside what it can do. Carrying the same file back to unmodified `main` and
running it there is what makes the point: the identical probe reports `rootfs=WRITABLE`, `uid=0:0`,
`NoNewPrivs:0` and `CapEff:00000000a80425fb`, the full default capability set, which is the silent
downgrade this PR closes.

```bash
bun run test                                                      # 900 files, 12,236 tests, 3 skipped, 8m31s
bun run --cwd {apps/api,apps/cli,apps/dashboard} lint              # tsc --noEmit, all clean
bun run --cwd {packages/core,packages/adapters,packages/db} lint   # all clean
RUN_DOCKER_E2E=1 bun run --cwd apps/api test:e2e                   # real daemon, never on a PR
```

**Testing.** The full suite is green on a quiet Linux host, installed from the
committed lockfile: `bun run test` exits 0 with 900 test files and 12,236 tests passing, 3 skipped,
in 8m31s. On unmodified `main` at `30ae48a4` the same command passes 897 files and 12,184 tests with
the same 3 skipped, so this branch adds 3 files and 52 tests and regresses nothing. The six workspace
typechecks above are clean.

**On `bun format`.** Running it bare rewrites most of the tree, because the committed state is already
prettier-dirty in files this PR happens to touch, so per CONTRIBUTING I formatted only the lines this
branch adds. `prettier --check` on the files this PR changes flags two, and every hunk it would
rewrite in them sits on a line this branch never touched; happy to send a formatting-only PR instead.

## Deliberately out of scope

The privilege-granting keys above, `pids_limit`, authoring any of the five through the dashboard or
the API, long-form `volumes: [{ type: tmpfs }]` (a mount whose kind would change is a different
question from the top-level `tmpfs:` key), and the bare runtime, which declares no
`unsupportedComposeKeys` today and which I left alone. The adoption wiring in `docker-reconcile.ts`
and `migrate.service.ts` is the one piece slightly past the issue's five points; it exists so the
inspect readback has a real consumer rather than dead weight, and its fallback is wholesale rather
than per key, matching how `resources` already behaves two lines above. A per-key merge would be
better for both fields and is a follow-up rather than something to apply to one of them alone. I can
split it out on request.

## Checklist

`bun format` is left unticked because running it bare reformats pre-existing drift across the repo,
which CONTRIBUTING tells me not to do; the other three commands are green and prettier passes on every
line this PR adds.

- [x] One change per PR: one bug, or one agreed feature, with nothing unrelated bundled in
- [x] The diff is scoped: no reformatting or lint fixes on lines I wasn't otherwise changing

Offered for your review at whatever pace suits you. Happy to move any of the boundaries above.


Hydralerne:
Closing this PR as outside the current integration scope, which the maintainer has limited to bugs and general improvements. This is new feature work. Feature request #749 remains open for a separate feature effort.

Deferred: new supported Compose hardening fields; PR #871 is not included. Existing unsupported-field reporting remains.

PR #874, opened 2026-09-09
fixed fetch git submodules during build checkouts

### LINKED ISSUE #835: [Bug] Git-based builds never fetch submodules (no --recurse-submodules / submodule update), so repos with submodule dependencies fail at build
## Summary

Projects whose repository uses git submodules cannot be built by Openship: the source checkout produced for the build contains the submodule directories as empty folders, so any Dockerfile step that references them (`COPY lib/foo ./lib/foo`, `cd lib/foo && npm ci`) fails. There is no project setting or flag to enable submodule fetching.

## Environment

- Self-hosted Openship server v0.7.0 (CLI 0.6.6), Linux host, GitHub-linked projects with `autoDeploy: true`.
- Observed on six projects in one org; the only project that builds is the one with no submodules.

## Reproduction

1. Create a repo with a git submodule, e.g. `lib/auth` → another repo in the same org, and a Dockerfile containing `COPY lib/auth ./lib/auth` followed by `RUN cd lib/auth && npm ci`.
2. Link it to an Openship project and deploy.
3. Build log: `[install] ↷ Repository Dockerfile owns dependency installation`, build context is tiny (the submodule dir is empty), then `npm ci` inside `lib/auth` fails (it walks up and finds the root `package-lock.json` because the submodule dir has no `package.json`).

Equivalent local reproduction of what the build sees: `git clone --depth 1 --branch main <repo> dir` leaves `dir/lib/auth` empty.

## Where

Inspected in the running `openship-api` container for v0.7.0: every clone site issues a plain `git clone --depth N --branch <branch> <url> <dir>` with no `--recurse-submodules` and no follow-up `git submodule update --init --recursive`:

- `packages/adapters/src/runtime/docker.ts` (two clone sites, around lines 1611 and 1642)
- `packages/adapters/src/runtime/build-pipeline.ts` (around lines 312–342)
- `packages/adapters/src/runtime/cloud.ts` (around line 1505)

`grep -r submodule` across `/app/apps` and `/app/packages` (excluding node_modules) returns nothing. Current `main` appears to prefer a GitHub archive tarball download before falling back to `git clone` (docker.ts ~1708–1746); archive tarballs also exclude submodule contents, so the tarball path has the same gap.

## Expected

Either fetch submodules by default (`--recurse-submodules --shallow-submodules`, or `git submodule update --init --recursive` after clone, using the same credentials as the parent clone so relative `../repo.git` submodule URLs resolve), or expose a per-project "fetch submodules" option. For the tarball path, fall back to git clone when the repo has a `.gitmodules` file.

## Workaround in use

Building the same repos by hand with `git submodule update --init --recursive` before `docker compose build` works, which confirms the checkout is the only difference.

DrDead0:
## Summary

This PR enables automatic git submodule fetching across all clone sites, resolving build failures for repositories containing submodules by ensuring the build contexts populate those submodule directories correctly.

## Motivation

Projects whose repositories use git submodules could not be built by Openship because the source checkout produced empty folders for the submodule directories. Any Dockerfile step that referenced them would fail (e.g., `npm ci` inside a submodule walking up to the root package-lock.json). This change allows submodule fetching by default matching the depth of the parent clones to natively support submodule-dependent projects without requiring user configuration.

## Related issue

Closes #835

## Changes

**packages/adapters:**
- **`src/runtime/build-pipeline.ts`:** Appended `--recurse-submodules --shallow-submodules` to the shallow clone strings. Added `git submodule update --init --recursive` securely for pinned commit checkouts.
- **`src/runtime/cloud.ts`:** Appended `--recurse-submodules --shallow-submodules` to the Cloud workspace checkout, and passed identical `credential.helper` properties to ensure relative submodule URLs resolve seamlessly. 
- **`src/runtime/docker.ts`:** Added `--recurse-submodules --shallow-submodules` everywhere to match local behavior.
- **`src/runtime/docker.ts`:** Added a `test -f .gitmodules` guard in the GitHub Tarball downloader fallback logic. Since GitHub archive tarballs exclude submodules, this accurately triggers a fallback to the native `git clone` when submodules are required.
- **`src/runtime/docker.ts` / `cloud.ts`:** Updated the build image `.git` directory cleaner strings to use `find ... -name .git -prune -exec rm -rf {} +` instead of static `rm -rf`, safely removing any nested submodule `.git` references out of the build context.

## Verification

Locally executed tests to confirm `packages/adapters` passes existing suites and the implementation does not break parent checks:

```bash
bun run turbo run test --filter=@repo/adapters


Hydralerne:
Incorporated this PR into [`feat/bugfix-integration-2026-09-15`](https://github.com/oblien/openship/tree/feat/bugfix-integration-2026-09-15) via [4470767d](https://github.com/oblien/openship/commit/4470767d), preserving the original commits and adapting the fix to the current architecture.

Git-based builds now materialize recursive submodules at the requested parent commit across server Docker, orchestrator Docker, shared build pipelines, and cloud contexts/source inspection. GitHub archives containing .gitmodules fall back to Git.

Validation:

- All 3,681 adapter tests passed; adapter TypeScript check passed.
- 85 focused tests passed, including 11 tests executing real Git commands against recursive/relative submodule fixtures, branch heads, pinned rollbacks, and unavailable dependencies.
- Seven real Git fixture tests fail against the original PR, covering missing architecture paths and rollback ordering.
- Real Git configuration matching verifies that the relay header is excluded from other repositories, lookalike paths, hosts, and HTTP downgrades.

Issue #565, opened 2026-08-12
ClamAV cannot start in the Docker deployment and fails silently — mail passes unscanned but still gets an X-Virus-Scanned header


yashthakur1:
In the Docker Compose deployment, clamd can never start. Two separate permission problems are each independently fatal, and neither is surfaced to the operator — every health indicator reports the scanner as running while nothing is being scanned.

## 1. Signature directory is not writable by clamd's uid

The bind-mounted signature directory is root-owned on the host, but clamd runs as uid 102 inside the container, so `freshclam` can never write:

```
ERROR: Can't create temporary directory /var/lib/clamav/tmp.<random>
Hint: The database directory must be writable for UID 102 or GID 106
ERROR: Update failed.
```

## 2. clamd cannot create its socket directory

With signatures fixed, clamd still cannot start, because it tries to create its socket directory under root-owned `/var/run`:

```
ERROR: LOCAL: Could not create socket directory: /var/run/clamav: Permission denied
ERROR: LOCAL: Socket file /var/run/clamav/clamd.ctl could not be bound: No such file or directory
```

This loops indefinitely. Meanwhile `supervisorctl status` reports `clamav-daemon RUNNING` and `ps` shows the process alive, so both of the obvious checks say the service is healthy.

## Impact

This is the part that matters most. Mail continues to flow. Amavis logs:

```
(!!)AV: ALL VIRUS SCANNERS FAILED
... Passed UNCHECKED ...
```

but the delivered message still carries an `X-Virus-Scanned: Debian amavis` header. An operator inspecting the deployment has every reason to believe virus scanning is active — supervisor says RUNNING, the process exists, and delivered mail is stamped as scanned — while in fact nothing is being scanned.

Also worth noting: `/var/run` is tmpfs, so even a manual fix inside the container is lost on every container restart.

## Expected behaviour

- clamd should start successfully in the Docker deployment without manual intervention.
- If clamd has no working socket, that should be reported as a failure, and mail should not be stamped `X-Virus-Scanned`.

## Reproduction

1. Install with `openship up --compose` (openship CLI 0.6.5, `openship-mail:0.6.5`), single box.
2. Deploy the mail engine from the Emails UI.
3. `docker logs openship-mail` — freshclam update failures, then the socket directory failure looping.
4. `supervisorctl status` inside the container — `clamav-daemon RUNNING`.
[...]

Issue #876, opened 2026-09-10
[Bug]: Email service is waiting for emails to go out forever


bigbasti:
### Existing issue

- [x] I have searched the existing issues and found no duplicate.

### Problem description

I created a fresh server end installed email.

Everything works, but when i send an email to postmaster@my_domain.com the email never reaches the predefined postmaster inbox, instead an additional email is queued and never leaves:

<img width="693" height="394" alt="Image" src="https://github.com/user-attachments/assets/da156821-f155-4a9d-8e4e-ed4f0deda4ac" />

Every time i sent an email to postmaster, the number increases.

Is there a way to see and cancel those queued emails?

### Steps to reproduce

see above

### Environment

- OpenShip version/commit: 0.7.2
- Deployment method: docker
- Operating system: ubuntu 24
- Browser (if applicable): edge

### Logs and screenshots

```shell

```

Hydralerne:
The bugfix integration branch now includes #885, preserving Francisco Trillo’s contribution. I reproduced its specific failure: leftover Amavis PID files after a container restart can make Amavis refuse to start, leaving mail deferred. The entrypoint now cleans those runtime files before starting the daemons. Two real-Docker tests and the updated PR CI passed; main is unchanged.

I am keeping this issue open because the original report describes a fresh installation, and the screenshot’s temporary failure alone does not establish the same cause. Please share the sanitized deferred reason from the mail queue and the Amavis/Postfix startup log around one failed attempt, along with the Openship/mail image version. Remove credentials, message content and personal addresses. That will identify whether this fix applies or a separate first-install failure remains.

PR #885 is merged into `feat/bugfix-integration-2026-09-15` and is waiting for that branch to merge into main.

Hydralerne:
Status correction after the latest review: **#885 is already in main through #891**. It fixes the reproduced Amavis restart failure caused by stale runtime PID/lock/socket files. #892 also includes #477's repair for Amavis transport TLS settings, including persisted mail configuration.

I checked this report's screenshot again. It records only “temporary failure” on a fresh installation, which does not establish either of those causes. This issue remains open because its fresh-install delivery failure is not yet verified as resolved; it is not waiting solely for #885 to merge.

[...]

PR #885, opened 2026-09-13
fix(mail): drop stale Amavis pid files on container restart

Frtrillo:
## Summary

- On `docker restart`, `openship-mail` keeps leftover Amavis pid/lock/socket files in `/var/run/amavis`.
- Net::Server then abort-loops (`Pid_file already exists for running process`) when that PID now belongs to another daemon (usually Dovecot).
- Originating mail defers on `127.0.0.1:10026` even though Health still shows Amavis as running.

## Changes

- **`apps/email/docker/entrypoint.sh`:** Unlink stale Amavis runtime files and recreate `/var/run/amavis` before supervisord starts. This entrypoint is the first process in a fresh pid namespace, so those files cannot refer to a living Amavis.
- **`apps/email/docker/supervisord.conf`:** Note why the entrypoint does this, next to the Amavis program.

This matches the existing ClamAV `/run` handling (issue #565): Debian's tmpfiles.d rule does not run under supervisord, and `docker restart` is not the same as `docker recreate`.

## Test plan

- [x] Scoped to the mail engine image entrypoint; no API/dashboard behavior change
- [ ] Recreate vs restart: `docker restart openship-mail` should leave Amavis listening on `127.0.0.1:10024` and `127.0.0.1:10026`
- [ ] `supervisorctl status amavis` should be `RUNNING`, not crash-looping on the leftover pid
- [ ] Deferred originating mail should flush (`postqueue -f`) once Amavis is up
- [ ] Health → Outbound delivery should not report `connect to 127.0.0.1:10026: Connection refused` after restart

Verified on a live `ghcr.io/oblien/openship-mail:0.7.2` box: leftover pid `67` was Dovecot after restart; removing the pid file and restarting Amavis unstuck 6 deferred messages through SES.

Made with [Cursor](https://cursor.com)


Hydralerne:
Thanks for the fix. I reproduced the stale-PID startup failure in the published 0.7.2 mail image and adapted this PR onto `feat/bugfix-integration-2026-09-15`, preserving your original commit `b7d53e97` and authorship. This PR targets the integration branch; main remains unchanged.

The maintainer addition moves the startup cleanup into a small helper invoked by the entrypoint. Failure to prepare the runtime directory now stops startup instead of continuing into an unusable mail daemon.

Validation:
- Two real-Docker tests passed against Debian 12 Amavis. A leftover PID naming an unrelated live process first makes Amavis fail; cleanup preserves that process and restores SMTP responses on both 10024 and 10026.
- A real container restart preserves the stale runtime files, which cleanup removes while repairing permissions and preserving mail data. Missing directories and repeated preparation are covered.
- API type checking and Bash syntax checks passed.

The test sends EHLO/QUIT over loopback only. It does not claim to reproduce every possible deferred-mail cause in #876. I will merge this PR into the integration branch after its updated CI passes.

Issue #608, opened 2026-08-17
[Bug]: Shows no system info for macbook


chbndrhnns:
0.6.5

I connected an old macbook pro 2016 via SSH but I am not getting any system info:

<img width="971" height="203" alt="Image" src="https://github.com/user-attachments/assets/6ca54fa8-0251-45b1-bfd9-51740ebd20ad" />

Hydralerne:
Fixed in #892, **waiting for merge into main**, by adapting and merging @chbndrhnns’s PR #645 (5bde484890beb362925c870dec603efe9fde4e79).

Reproduced the invalid JSON on an actual Mac with main’s Linux-only command. The shared platform monitor now uses standard macOS probes and retains the same API/native response format. Memory uses the host page size; CPU uses an interval sample; failed probes return an error instead of false zero readings.

Seven command execution tests pass, including actual macOS sh/zsh samples and Linux/Intel/ARM fixtures. All workspace build/typecheck tasks pass. Closing as fixed on the integration branch, pending #892.

PR #888, opened 2026-09-14
fix: support macOS servers in the CPU/memory/disk/uptime monitor

zerox9dev:
## Summary
- The CPU/memory/disk/uptime monitor for self-hosted "servers" reads straight from `/proc`, which macOS doesn'''t have — a Mac added as an Openship deploy target always showed dashes in System Health.
- Adds a Darwin branch (`top`/`vm_stat`/`sysctl`/`df`) and dispatches on `uname -s`, so one SSH round-trip still gets the stats regardless of platform. No new dependency — every command used ships with macOS.
- The Darwin branch avoids `<<<` on purpose: zsh (macOS'''s default shell, and what `ssh host cmd` actually runs) implements here-strings via a real temp file, which made this exact probe fail with ENOSPC on a nearly-full disk — precisely the box an operator most wants a disk-usage reading from. Found and fixed while testing against a real Mac mini with 99% disk usage.
- Caught (and fixed) a join bug of my own along the way: concatenating the two branches into one `if/then/else/fi` command needs an explicit `;` before `else`/`fi`, or the trailing `printf` swallows them as arguments instead of shell keywords and the whole probe never runs on either OS. The added `bash -n` syntax-check test catches this class of bug directly.

## Test plan
- [x] `bunx vitest run apps/api/src/modules/system/server-stats-command.test.ts` — 4 new tests: both branches parse as valid bash, the combined command still dispatches correctly, the Darwin branch has no `<<<`, and the Linux branch actually runs on the (Linux) CI runner and emits the documented JSON schema.
- [x] Manually verified the Linux branch on an Ubuntu box and the Darwin branch over a real SSH exec against a macOS 15.4 (Apple Silicon) Mac mini — both return valid, sane values.
- [x] `tsc --noEmit` — no new errors introduced.


Hydralerne:
Reviewed alongside #645 against current main. Both PRs address the same macOS failure and modify the API controller that has since moved to the shared platform layer. I adapted and merged the earlier, smaller #645 into #892, retaining @chbndrhnns’s original commit and GitHub merge credit.

The integrated command also addresses the shell concern raised here: neither OS path uses here-strings, and actual macOS execution through both sh and zsh passes. It keeps used + available memory equal to total, uses an interval CPU sample, and reports incomplete probes as errors. This PR’s separate controller implementation would duplicate that fix; its active/wired/compressed “used” value and free-only “available” value would also leave the memory fields inconsistent.

Thanks for investigating the macOS probe and shell behavior. Closing this overlapping PR as superseded by the adapted #645 in #892; #608 is tracked as fixed there, awaiting main merge.

